PR #11, opened 2021-01-27
T31260 content page

manuq:
https://phabricator.endlessm.com/T31260

### DIFF
--- a/kolibri_explore_plugin/apps/pbskids/metadata.json
+++ b/kolibri_explore_plugin/apps/pbskids/metadata.json
@@ -1,0 +1,3 @@
+{
+  "contentBackgroundColor": "#ffe000"
+}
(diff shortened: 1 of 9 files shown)


danigm (approved):
LGTM

danigm (approved):
LGTM

manuq:
> LGTM

After a while working on this I was expecting more from the review :-P Thanks anyways for looking :+1:

danigm:
> > LGTM
> 
> After a while working on this I was expecting more from the review :-P Thanks anyways for looking +1

We've been trying different solutions here, and at the end, I think it's okay to go with this implementation, this solution looks good to me and it works correctly.

I know that the LGTM comment is not too much feedback, I could leave a comments about some details, for example the movement of the background url string build in the handler instead of the component methods, but at the end I think it's a good solution and I didn't want to start a discussion just for this kind of tech decisions.

PR #14, opened 2021-02-01
Pre commit: use Python 3.7

manuq:
https://phabricator.endlessm.com/T31260

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -14,7 +14,6 @@ repos:
   rev: v1.3.3
   hooks:
   - id: reorder-python-imports
-    language_version: python2.7
 - repo: local
   hooks:
   - id: lint-frontend


manuq:
@danigm the "Lint failed" above means we should also change the CI script?

.pre-commit-config.yaml
```diff
@@ -14,7 +14,7 @@ repos:
   rev: v1.3.3
   hooks:
   - id: reorder-python-imports
-    language_version: python2.7
+    language_version: python3.7
```
danigm, line 17: I think that we can even remove this line and it will use the system version.

manuq: But I think different Python versions will probably give different import orders, and we are trying to ensure the exact same output. That's the whole point of using pipenv, no?

manuq: I see we are not specifying the python version in the other hooks, so I'll try removing as you ask.

danigm: The main problem with this is that this won't work in systems with others python versions, ubuntu uses 3.8. With 2.7 it worked because the python2.7 is the last version of python2 supported and that won't change, if you have python2 you'll have python2.7 on your system. But I think it's better to remove to make it work on local system and have a fixed version in the CI so we'll test against the same version on the server.

PR #17, opened 2021-02-03
Set yarn timeout to 10 minutes

manuq:
The github workflow check usually fails to download the
material-design-icons tarball with a timeout. Let's see
if 10 minutes is enough.

https://phabricator.endlessm.com/T31344


dnohales (approved):
Makes sense.

PR #18, opened 2021-02-05
Improve custom app zip urls

danigm:
https://phabricator.endlessm.com/T31363

### DIFF
--- a/kolibri_explore_plugin/views.py
+++ b/kolibri_explore_plugin/views.py
@@ -50,11 +50,8 @@ class AppView(AppBase):
     @cache_forever
     @xframe_options_exempt
     @add_security_headers
-    def get(self, request, app, path):
+    def get(self, request, app, path=""):
         filename = self._get_file(app, "custom-channel-ui.zip")
-
-        if path.startswith("/"):
-            path = path[1:]
 
         with zipfile.ZipFile(filename) as zf:
             response = get_embedded_file(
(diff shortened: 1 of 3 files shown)


kolibri_explore_plugin/views.py
```diff
@@ -50,12 +50,9 @@ class AppView(AppBase):
     @cache_forever
     @xframe_options_exempt
     @add_security_headers
-    def get(self, request, app, path):
+    def get(self, request, app, path="index.html"):
```
manuq, line 53: The path was always "/", I'm not sure why change it to "index.html".

manuq (approved):
I left a question, other than that looks good!

kolibri_explore_plugin/views.py
danigm: Indeed this is exactly what it is done in the kolibri.core function that we're using: https://github.com/endlessm/kolibri/blob/release-v0.14.x/kolibri/core/content/views.py#L244-L245

We are removing the `/` from the path right now: https://github.com/endlessm/kolibri-explore-plugin/blob/master/kolibri_explore_plugin/views.py#L56-L57, this PR removes that code because we removed the leading `/` from the `path`, because it's outside of the url regular expression.

If we send the empty string the file returned will be `index.html`. I've just set that to be more explicit, but in any case, we can delegate that transformation to the `get_embedded_file` function and just pass an empty string.

manuq: Ah I see! Yeah, I think is better to just pass an empty path to `get_embedded_file()` and don't duplicate it's implementation here.

PR #19, opened 2021-02-11
Avoid cache in AppView

danigm:
The cache is done by the browser and even with a fresh built endless
key, the cache in the browser is used to render.

This patch removes the cache_forever and adds a never_cache to avoid
browser caching.

In the future we can think about using the version number in urls to be
able to use the cache_forever, but for now this will fix the known
issues.

https://phabricator.endlessm.com/T31425

### DIFF
--- a/kolibri_explore_plugin/urls.py
+++ b/kolibri_explore_plugin/urls.py
@@ -1,5 +1,6 @@
 from django.conf.urls import url
 
+from . import __version__ as VERSION
 from .views import AppFileView
 from .views import AppMetadataView
 from .views import AppView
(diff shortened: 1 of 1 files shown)


tox.ini
```diff
@@ -65,7 +65,8 @@ commands =
     yarn
     make dist
     # Test whl installation
-    pip install {toxinidir}/dist/kolibri_explore_plugin-0.0.1-py2.py3-none-any.whl
+
+    pip install {toxinidir}/dist/kolibri_explore_plugin-0.0.3-py2.py3-none-any.whl
```
manuq, line 69: We should stop doing hotfixes and fix this once and for all, by removing the hardcoded value. Maybe it's just a matter of:

```
    pip install {toxinidir}/dist/kolibri_explore_plugin-*.whl
```

We are tracking this issue in a separate ticket: https://phabricator.endlessm.com/T31414

kolibri_explore_plugin/views.py
```diff
@@ -47,7 +47,7 @@ def _get_file(self, app, path):
 
 
 class AppView(AppBase):
-    @cache_forever
+    @never_cache
```
manuq, line 50: As discussed in the ticket this hotfix is not a good idea.

manuq (changes requested):

manuq (approved):

PR #20, opened 2021-02-15
CI tool: Remove hardcoded version

manuq:
From the github-actions configuration.

https://phabricator.endlessm.com/T31414

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -65,7 +65,9 @@ commands =
     yarn
     make dist
     # Test whl installation
-    pip install {toxinidir}/dist/kolibri_explore_plugin-0.0.1-py2.py3-none-any.whl
+    # FIXME: Disabled until tox.ini supports expanding globs.
+    # See: https://github.com/tox-dev/tox/issues/1571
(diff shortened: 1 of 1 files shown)


tox.ini
```diff
@@ -65,7 +65,7 @@ commands =
     yarn
     make dist
     # Test whl installation
-    pip install {toxinidir}/dist/kolibri_explore_plugin-0.0.1-py2.py3-none-any.whl
+    pip install {toxinidir}/dist/kolibri_explore_plugin-*.whl
```
danigm, line 68: This is not working, I think it's because the tox tooling that sees that like a filename and doesn't apply the glob. Maybe we should add the version here like a variable

manuq: Yeah. I don't know about tox but I will check how to do that.

danigm: Maybe this is interesting: https://github.com/tox-dev/tox/issues/1571

danigm: If it's too complex we can just remove this line from tests, it's something not required to test. With the `make dist` is enough.

manuq: Indeed. I updated the PR disabling the command and referencing the ticket in tox.

PR #23, opened 2021-02-22
T31413 bump version

manuq:
https://phabricator.endlessm.com/T31413

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,0 +1,6 @@
+[bumpversion]
+current_version = 0.0.4
+commit = True
+tag = True
+
+[bumpversion:file:kolibri_explore_plugin/__init__.py]
--- a/Makefile
+++ b/Makefile
@@ -3,8 +3,9 @@
 help:
 	@echo "clean-build - remove build artifacts"
 	@echo "clean-pyc - remove Python file artifacts"
-	@echo "release - package and upload a release"
+	@echo "bumpversion - bump the package version"
 	@echo "dist - package"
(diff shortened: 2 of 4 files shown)


danigm (commented):
We're using the `requirements.txt` dependencies file here, so if we're going to track the Pipfile with dependencies there, we need to sync the requirements with the `requirements.txt` file, or replace completely.

Right now the `requirements.txt` is used for tests, but we can generate that file automatically for tests, before running tox: https://pipenv.pypa.io/en/latest/advanced/#generating-a-requirements-txt

Maybe that's the way to go, just remove the `requirements.txt` and update the python workflow to install and run pipenv to generate the requirements.txt file before calling the tox command.

manuq:
> We're using the `requirements.txt` dependencies file here, so if we're going to track the Pipfile with dependencies there, we need to sync the requirements with the `requirements.txt` file, or replace completely.
> 
> Right now the `requirements.txt` is used for tests, but we can generate that file automatically for tests, before running tox: https://pipenv.pypa.io/en/latest/advanced/#generating-a-requirements-txt
> 
> Maybe that's the way to go, just remove the `requirements.txt` and update the python workflow to install and run pipenv to generate the requirements.txt file before calling the tox command.
> 
> Or maybe we can use this for tox: https://tox-pipenv.readthedocs.io/en/latest/

Ah sorry, I though that we were using Pipenv in this repository. But now I see that I started using it locally myself. So I will remove the Pipenv files from this PR.

danigm (approved):

PR #29, opened 2021-03-03
Demo: Add header component

danigm:
This new header component has the endless logo and two links, one for
the home and another fake link with the text "Learn".

The search input has been moved to this new component.

The base component has been updated to remove the SideNav component so
there's no way to see the kolibri sidenav inside the plugin.

https://phabricator.endlessm.com/T31548


manuq:
@danigm could you fix the conflict and then merge directly?

PR #33, opened 2021-03-03
T31545 tags component

manuq:
https://phabricator.endlessm.com/T31545

### DIFF
--- a/kolibri_explore_plugin/assets/src/components/TagRow.vue
+++ b/kolibri_explore_plugin/assets/src/components/TagRow.vue
@@ -97,19 +97,24 @@
       window.addEventListener('resize', this.onResize);
       const style = getComputedStyle(document.body);
       this.breakpoints['xs'] = parseInt(
-        style.getPropertyValue('--breakpoint-xs').replace('px', '')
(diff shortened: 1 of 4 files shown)


kolibri_explore_plugin/assets/src/components/TagRow.vue
```diff
@@ -0,0 +1,203 @@
...
+    },
+    watch: {
+      containerWidth: function() {
+        this.debouncedUpdate();
+      },
+    },
+    created() {
+      window.addEventListener('resize', this.onResize);
+      const style = getComputedStyle(document.body);
+      this.breakpoints['xs'] = parseInt(
```
danigm, line 99: Here we need the `radix` parameter in all the `parseInt`: https://eslint.org/docs/rules/radix

danigm (changes requested):
Looks good to me. Just one comment about the `parseInt` function. With that I think we  can merge this and continue working.

kolibri_explore_plugin/assets/src/components/TagRow.vue
manuq: Added!

danigm (approved):

PR #37, opened 2021-03-04
Demo: Animate tagrow with transform

danigm:
This patch implements the animation using css transformation instead of vue javascript code and margin. I don't know if this is better than the other option, but I had something similar in the past and I just want to share this solution.

This is not perfect right now. There's a small problem with the offset cards that aren't there until the end of the animation so this is not really smooth, but could be a quick solution for the demo if the other animation takes more time.

@manuq, take a look and decide if you like this or if you want to go ahead with the other implementation. Don't hesitate to ignore this if you've a clear idea of the animation, this is just a suggestion PR.

https://phabricator.endlessm.com/T31545


manuq (approved):
This is great!

PR #51, opened 2021-03-31
Add tags to content nodes

manuq:
https://phabricator.endlessm.com/T31705


danigm (approved):
Looks good to me. I've just left a couple of comments about code styling for lists, but at the end it's the same, so feel free to apply the suggestions or just merge as it is right now.

manuq:
> Looks good to me. I've just left a couple of comments about code styling for lists, but at the end it's the same, so feel free to apply the suggestions or just merge as it is right now.

Thank you very much. Indeed list comprehesions are better than map.

PR #52, opened 2021-04-01
docker: Add populate script to master image

danigm:
This PR adds the populate script to the docker image and tries to populate the initial content in the `entrypoint` if it was not populated before. After the population it creates the file `$KOLIBRI_HOME/populated`. If this file exists, the content is not populated, this way we only populate the content once.

https://phabricator.endlessm.com/T31632

### DIFF
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -1,6 +1,15 @@
 #!/bin/bash
 
+# plugin enable
+kolibri plugin enable kolibri.plugins.app
+kolibri plugin disable kolibri.plugins.learn
+kolibri plugin enable kolibri_explore_plugin
+
+# Ensure that db is created and updated with the current kolibri version
 kolibri manage migrate
+# Collect statics to ensure they are updated
+kolibri manage collectstatic --clear --noinput
+
 kolibri services --background
 
 uwsgi --ini uwsgi.ini
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -17,14 +17,10 @@ ENV KOLIBRI_CHERRYPY_START=false
 RUN pip install uwsgi
 RUN pip install psycopg2-binary
 RUN pip install kolibri
-RUN pip install kolibri-explore-plugin
-
-# plugin enable
-RUN kolibri plugin enable kolibri.plugins.app
-RUN kolibri plugin disable kolibri.plugins.learn
-RUN kolibri plugin enable kolibri_explore_plugin
 
 COPY uwsgi.ini /docker/uwsgi.ini
 COPY entrypoint.sh /docker/entrypoint.sh
+
+RUN pip install kolibri-explore-plugin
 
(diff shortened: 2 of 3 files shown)


docker/entrypoint.sh
```diff
@@ -1,6 +1,26 @@
...
+kolibri plugin enable kolibri.plugins.app
+kolibri plugin disable kolibri.plugins.learn
+
+# Disable and enable the kolibri_explore_plugin to ensure that the static files
+# are correctly copied to the static folder
+kolibri plugin disable kolibri_explore_plugin
+kolibri plugin enable kolibri_explore_plugin
+
+# Initial populate
+if [ $POPULATE_CONTENT = 'true' -a ! -e $KOLIBRI_HOME/populated ]
```
manuq, line 18: Nit: why not just `if [ ! -e $KOLIBRI_HOME/populated ]`. In other words, why is `POPULATE_CONTENT` useful at all if it's always true?

docker/entrypoint.sh
```diff
@@ -1,6 +1,26 @@
 #!/bin/bash
 
-kolibri manage migrate
-kolibri services --background
+if [ $RUN_BG_SERVICE = 'true' ]
```
manuq, line 3: Nit: same as the other env var. Is there any case where entrypoint.sh is called without RUN_BG_SERVICE being true?

docker/entrypoint.sh
```diff
@@ -1,6 +1,26 @@
...
+kolibri plugin disable kolibri_explore_plugin
+kolibri plugin enable kolibri_explore_plugin
+
+# Initial populate
+if [ $POPULATE_CONTENT = 'true' -a ! -e $KOLIBRI_HOME/populated ]
+then
+    kolibri manage migrate
+    # populate initial content
+    /docker/populate/populate.sh
+    echo "done" > $KOLIBRI_HOME/populated
```
manuq, line 23: :+1:

docker/populate/kolibri-pick-content-from-channel
```diff
@@ -0,0 +1,185 @@
+#!/usr/bin/python3
```
manuq, line 1: Can't we fetch this file from the original repository? Instead of copying it and the 99% chance that it will get outdated as soon as the first change.

docker/populate/populate.sh
```diff
@@ -0,0 +1,49 @@
+echo "Populating initial content in ${KOLIBRI_HOME}"
```
manuq, line 1: Same thing as above. If we keep duplicating files we are going to shoot ourselves in the foot at one point.

manuq (approved):
Looks good. Approving for practical reasons, but honestly I'm aganst duplicating files in multiple repositories.

danigm:
I've removed all the populate script and just let the unification of `Dockerfiles` and `entrypoint.sh`. Because the populate is something to do just once, I'll try to do that with a branch build.

manuq (approved):
Perfect.

PR #53, opened 2021-04-05
ContentPage: Display tags

manuq:
Just display them for now. This will change once we have a lightbox
with the content. Then the tags should be displayed by the custom
presentation.

https://phabricator.endlessm.com/T31711

### DIFF
--- a/kolibri_explore_plugin/assets/src/views/ContentPage.vue
+++ b/kolibri_explore_plugin/assets/src/views/ContentPage.vue
@@ -41,6 +41,9 @@
           <p v-if="content.author">
             {{ $tr('author', { author: content.author }) }}
           </p>
+          <p v-if="content.tags">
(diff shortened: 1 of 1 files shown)


danigm (approved):
LGTM

kolibri_explore_plugin/assets/src/views/ContentPage.vue
```diff
@@ -41,6 +41,9 @@
           <p v-if="content.author">
             {{ $tr('author', { author: content.author }) }}
           </p>
+          <p v-if="content.tags">
```
danigm, line 44: if `content.tags` is an empty list this will be shown anyway because `Boolean([]) === true`,  but in any case, I think it's okay for this initial version.

manuq: True, maybe we should improve the comparison.

PR #55, opened 2021-04-07
Sync tags with upstream

manuq:
It is more performant to use the queryset parameter in the consolidate
method. See https://github.com/learningequality/kolibri/pull/7941

https://phabricator.endlessm.com/T31711

### DIFF
--- a/kolibri_explore_plugin/viewsets.py
+++ b/kolibri_explore_plugin/viewsets.py
@@ -1,5 +1,5 @@
 from kolibri.core.content.api import ContentNodeViewset
-from kolibri.core.content.models import ContentNode
+from kolibri.core.content.models import ContentTag
 
 
 class CustomContentNodeViewset(ContentNodeViewset):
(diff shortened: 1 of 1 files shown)


kolibri_explore_plugin/viewsets.py
```diff
@@ -1,15 +1,24 @@
...
+        tags = {}
+        for t in ContentTag.objects.filter(tagged_content__in=queryset).values(
+            "tag_name",
+            "tagged_content",
+        ):
+            if t["tagged_content"] not in tags:
+                tags[t["tagged_content"]] = [t["tag_name"]]
+            else:
+                tags[t["tagged_content"]].append(t["tag_name"])
+
```
danigm, line 19: This could be simplified with [`values_list`](https://docs.djangoproject.com/en/3.1/ref/models/querysets/#values-list) instead of `values`, with something like:

```python
for name, content in ContentTag.objects.filter(tagged_content__in=queryset).values_list(
    "tag_name",
    "tagged_content",
):
    if content not in tags:
        tags[content] = [name]
    else:
        tags[content].append(name)
```

In any case, this is only syntax sugar, and this is now upstream, so there's no need to change.

danigm (approved):

kolibri_explore_plugin/viewsets.py
manuq: This is just to sync with the code already merged upstream. Would you like to make that suggestion as a PR for kolibri upstream?

PR #57, opened 2021-04-23
T31806 navigation

manuq:
https://phabricator.endlessm.com/T31806


danigm (approved):
LGTM. Adds a new button instead of using the current `x`, but I assume that's by desing. At some point the `x` button could be converted to an Endless logo or other kind of button that points to the channel list.

manuq:
> LGTM. Adds a new button instead of using the current `x`, but I assume that's by desing. At some point the `x` button could be converted to an Endless logo or other kind of button that points to the channel list.

Yes, it should behave like a lightbox, like the upstream solution. We are not there yet. This X button mimics closing that lightbox.

PR #58, opened 2021-04-28
CI: Build whl artifact

danigm:
https://phabricator.endlessm.com/T31830


manuq (commented):
This is for having master in the online instance, right? I wonder if you could use `pip install -e` like we do for development to create a symlink. And thus not having to fake the version number or install the wheel file each time.

danigm:
Let's close this for now, we can open a new PR if we decide to build the package in CI again.

PR #59, opened 2021-04-28
Base: Change X icon with the Endless logo

manuq:
Use a flat KButton from the Kolibri UI components. The icon is pasted
directly into the template because there is no webpack config to
import SVGs as vue components in this app. Also, we have to add an
exception for the line max length.

https://phabricator.endlessm.com/T31864


danigm (approved):
LGTM

PR #60, opened 2021-04-29
T31806 lightbox close

manuq:
https://phabricator.endlessm.com/T31806

### DIFF
--- a/kolibri_explore_plugin/assets/src/modules/topicsTree/handlers.js
+++ b/kolibri_explore_plugin/assets/src/modules/topicsTree/handlers.js
@@ -152,3 +152,37 @@ export function showTopicsContent(store, id) {
     }
   );
 }
+
+export function showTopicsContentInLightbox(store, id) {
+  const promises = [
+    ContentNodeResource.fetchModel({ id }),
+    ContentNodeResource.fetchNextContent(id),
+    store.dispatch('setChannelInfo'),
+  ];
+  ConditionalPromise.all(promises).only(
+    samePageCheckGenerator(store),
+    ([content, nextContent]) => {
+      const currentChannel = store.getters.getChannelObject(content.channel_id);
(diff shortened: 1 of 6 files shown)


danigm (approved):
LGTM

dylanmccall (commented):
I'm trying to get to the bottom of an issue here where, if I choose a _video_ from the channel root (as opposed to a topic, and then a video in that topic), the lightbox opens but behind that we're redirected to Kolibri's Device dashboard. Is anyone else seeing that?

The end result looks like this:

Screenshot_2021-05-04 Healthy Living

The URL stays the same, like http://0.0.0.0:5780/en/learn/#/topics/85461059d7d158f7ac03d050ddbf734e.

To reproduce, you could try navigating to that URL (replacing `:5780` with wherever your Kolibri is), and then click "Yoga for a Happier Back", "Trauma-Conscious Meditation for Tension Release and Self-Regulation", or some other item which leads directly to content rather than to a topic.

danigm:
> To reproduce, you could try navigating to that URL (replacing `:5780` with wherever your Kolibri is), and then click "Yoga for a Happier Back", "Trauma-Conscious Meditation for Tension Release and Self-Regulation", or some other item which leads directly to content rather than to a topic.

It's working correctly for me. I had a similar issue testing with and old version of custom channel application, but I've tested with the current master and the lightbox is working as expected for me.

erikos:
> > To reproduce, you could try navigating to that URL (replacing `:5780` with wherever your Kolibri is), and then click "Yoga for a Happier Back", "Trauma-Conscious Meditation for Tension Release and Self-Regulation", or some other item which leads directly to content rather than to a topic.
> 
> It's working correctly for me. I had a similar issue testing with and old version of custom channel application, but I've tested with the current master and the lightbox is working as expected for me.

Thanks for checking - let's get it in!

PR #62, opened 2021-05-05
ContentLightbox: Remove download and share buttons

dylanmccall:
https://phabricator.endlessm.com/T31881


danigm (changes requested):
Looks good, but the lint is failing.

You should install the pre-commit hook to pass the linter before every commit: https://github.com/endlessm/kolibri-explore-plugin#configuring-the-precommit-hook

manuq (approved):
Thanks for fixing the lint issues. Looks great now.

PR #63, opened 2021-05-07
ContentLightbox: Add Close button

dylanmccall:
This change moves the body part of ContentLightbox into a ContentItem component, as in upstream Kolibri, so we can think about the lightbox independently. I'm tweaking the modal to put a close button alongside the title using the existing UiToolbar component. This isn't _perfect_ (we need to add some extra CSS for UiToolbar to behave itself and it doesn't expect to have an arbitrarily long title). But it feels like a step in the right direction.

https://phabricator.endlessm.com/T31882

### DIFF
--- a/kolibri_explore_plugin/assets/src/views/CustomChannelsPage.vue
+++ b/kolibri_explore_plugin/assets/src/views/CustomChannelsPage.vue
@@ -57,9 +57,13 @@
     bottom: 0;
     left: 0;
 
+    /* Above the sidenav */
+    z-index: 16;
+
     /* With a semi transparent background */
     background: rgba(0, 0, 0, 0.5);
   }
+
   .lightbox {
     /* Center in overlay */
     position: absolute;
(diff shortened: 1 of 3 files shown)


dylanmccall:
Here's a screenshot. Note that the lightbox overlay overlaps the explore-buttons container so it is not interactive. Still need to get this thing sizing itself correctly, and all sorts of visual enhancements, but I'm considering those out of scope for this branch.

Screenshot_2021-05-06 Healthy Living

kolibri_explore_plugin/assets/src/views/CustomChannelsPage.vue
```diff
@@ -57,9 +57,13 @@
     bottom: 0;
     left: 0;
 
+    /* Above the sidenav */
+    z-index: 16;
```
manuq, line 61: :+1:

manuq (approved):
Great, it works as expected. Next time please add a paragraph in the commit message explaining the approach and all things done in more detail, example:

- The lightbox is now on top of the sidenav.
- Added ContentItem component because...
- Removed PageHeader because...
- ...

And also if the commit does too many things, consider splitting into multiple commits.

dylanmccall:
Okay, did some rebasing to clear up these commits. Thanks for the review :)

PR #67, opened 2021-05-21
T31931 monorepo

manuq:
https://phabricator.endlessm.com/T31931


danigm (approved):
Looks good to me. This is a great example widget. I've just do the test and it worked correctly for me with `yarn dev`.

PR #68, opened 2021-05-21
T31931 monorepo

manuq:
https://phabricator.endlessm.com/T31989


danigm (approved):
LGTM, we'll need to take care to put the corresponding styles in the correct module, to be able to override what we need and to reuse as much as possible. This new structure will let us to do that.

PR #71, opened 2021-05-25
Template: Add placeholder while loading channel

danigm:
This patch adds some placeholder carousel and first row of content while
loading the content so we give some feedback to the user about the
loading state.

https://phabricator.endlessm.com/T31986


manuq (approved):
Nice, let's see what Joana thinks about this. Is certainly an improvement. Another option is to use a spinner. https://bootstrap-vue.org/docs/components/spinner . Let's merge since you already did the work but maybe next time ask Joana first.

PR #72, opened 2021-05-26
T31941 channel cards

danigm:
Initial implementation of the channel list with custom component `ChannelCard` in the `eos-components` package.

https://phabricator.endlessm.com/T31941
https://phabricator.endlessm.com/T31946


kolibri_explore_plugin/assets/package.json
```diff
@@ -7,7 +7,8 @@
     "bootstrap": "^4.5.3",
     "bootstrap-vue": "^2.21.2",
     "eos-components": "^0.1.0",
-    "kolibri-design-system": "https://github.com/learningequality/kolibri-design-system"
+    "kolibri-design-system": "https://github.com/learningequality/kolibri-design-system",
+    "underscore": "^1.13.1"
```
manuq, line 11: :+1:

manuq: In template_ui we have "underscore": "^1.11.0", so we need to bump that one to keep our dependencies the same. I'll do as follow-up. Also lerna has commands to manage the same dependency in multiple packages.

kolibri_explore_plugin/assets/src/views/ChannelsPage.vue
```diff
@@ -44,53 +32,38 @@
...
     mixins: [commonCoreStrings],
-    data() {
-      return {
-        searchQuery: '',
-        searchFocus: false,
-      };
+    props: {
+      columns: {
+        type: Number,
+        default: 3,
```
manuq, line 48: Yep I though it was 4, but looking at Figma Joana is using 3 columns. +1

kolibri_explore_plugin/assets/src/views/ChannelsPage.vue
```diff
@@ -1,41 +1,29 @@
...
-        @focus="searchFocus = true"
-        @blur="searchFocus = false"
-      >
-
-      <KIconButton
-        icon="search"
-        :appearance="searchAppareance()"
-        @click="filter"
+  <b-container class="channels mb-3 mt-3">
+    <b-card-group
```
manuq, line 4: b-card-group are not responsive. I guess you are trying to do the same as in the template for CardRow, so it's fine for now.

packages/eos-components/src/components/ChannelCard.vue
```diff
@@ -0,0 +1,82 @@
...
+        };
+      },
+    },
+  }
+</script>
+
+<style lang="scss" scoped>
+  @import '../styles';
+  @import 'bootstrap/scss/bootstrap';
+  @import 'bootstrap-vue/src/index';
```
manuq, line 50: I will open a PR with a new structure for the SCSS. For now yes, you have to do this. I'm afraid we'll need to rethink the overrides.

manuq (approved):

PR #75, opened 2021-05-26
Add override for TEDEd Think Like A Coder channel

erikos:


manuq (approved):

manuq:
I'm not sure why, but our checks fail with "customApps.js did not conform to prettier standards". Maybe the new app name is too long?
```
Kolibri Linter: 
Kolibri Linter: Linting errors for kolibri_explore_plugin/assets/src/customApps.js
Kolibri Linter: kolibri_explore_plugin/assets/src/customApps.js did not conform to prettier standards
```

This can be reproduced locally by running `pre-commit run -a`. If you setup the pre-commit hook as in the README, it won't let you commit this at all.

PR #78, opened 2021-05-27
T31895 typographic

manuq:
https://phabricator.endlessm.com/T31895

### DIFF
--- a/packages/eos-components/src/components/TopicCard.vue
+++ b/packages/eos-components/src/components/TopicCard.vue
@@ -1,0 +1,85 @@
+<template>
+  <b-card
+    noBody
+    class="my-2 rounded-lg"
+    :class="{
+      shadow: !isHovered,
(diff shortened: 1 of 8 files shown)


packages/eos-components/src/components/TopicCard.vue
```diff
@@ -0,0 +1,85 @@
...
+@import '../styles.scss';
+
+.card {
+  background-color: $primary;
+  transition: all ease .4s;
+}
+
+.card-body {
+  // This 8 is an estimation:
+  min-height: card-body-height(8);
```
danigm, line 67: The card image height is calculated with a percentage and a padding top: https://github.com/endlessm/kolibri-explore-plugin/blob/master/packages/template-ui/src/components/RegularCard.vue#L79

danigm (approved):
LGTM

PR #79, opened 2021-05-28
T31985 ie11 fixes

danigm:
Fix of various IE11 problems with the template:

 * Arrow buttons
 * Card body flex-box positioning
 * Fix the search input field

https://phabricator.endlessm.com/T31985

### DIFF
--- a/packages/template-ui/vue.config.js
+++ b/packages/template-ui/vue.config.js
@@ -1,4 +1,7 @@
 module.exports = {
   publicPath: './',
   transpileDependencies: ['vue-clamp', 'resize-detector'],
+  css: {
+    extract: false,
+  },
 };
--- a/packages/template-ui/vue.config.js
+++ b/packages/template-ui/vue.config.js
@@ -2,6 +2,7 @@ module.exports = {
   publicPath: './',
   transpileDependencies: ['vue-clamp', 'resize-detector'],
   css: {
+    // FIXME this is because of IE11 compatibility:
     extract: false,
   },
 };
--- a/packages/eos-components/src/components/TopicCard.vue
+++ b/packages/eos-components/src/components/TopicCard.vue
@@ -64,7 +64,7 @@ export default {
 
 .card-body {
   // This 8 is an estimation:
-  min-height: card-body-height(8);
+  height: card-body-height(8);
   display: flex;
   flex-direction: column;
   justify-content: space-between;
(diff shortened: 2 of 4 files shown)


packages/template-ui/vue.config.js
```diff
@@ -1,4 +1,7 @@
 module.exports = {
   publicPath: './',
   transpileDependencies: ['vue-clamp', 'resize-detector'],
+  css: {
+    extract: false,
```
manuq, line 5: This is unfortunate, as removing the standalone CSS file makes the JS huge. But on the other hand is great that you found the source of the error.
[This issue](https://github.com/vuejs/vue-cli/issues/4646#issuecomment-538605177) marks mini-css-extract-plugin "^0.6.0" as rotten. We have 2 versions, 0.4.5 and 0.9.0. The older because of kolibri-tools:

```
$ yarn why mini-css-extract-plugin
yarn why v1.22.10
[1/4] Why do we have the module "mini-css-extract-plugin"...?
[2/4] Initialising dependency graph...
[3/4] Finding dependency...
[4/4] Calculating file sizes...
=> Found "mini-css-extract-plugin@0.4.5"
info Reasons this module exists
   - "_project_#discovery-page#kolibri-tools" depends on it
   - Hoisted from "_project_#discovery-page#kolibri-tools#mini-css-extract-plugin"
info Disk size without dependencies: "68KB"
info Disk size with unique dependencies: "396KB"
info Disk size with transitive dependencies: "3.63MB"
info Number of shared dependencies: 12
=> Found "@vue/cli-service#mini-css-extract-plugin@0.9.0"
info This module exists because "_project_#pbskids-ui#@vue#cli-service" depends on it.
info Disk size without dependencies: "100KB"
info Disk size with unique dependencies: "468KB"
info Disk size with transitive dependencies: "3.8MB"
info Number of shared dependencies: 19
```

We should at least add a comment here explaining that this is because of IE11.

manuq, line 5:
```suggestion
    // FIXME this is because of IE11 compatibility:
    extract: false,
```

manuq (approved):
Great work, thanks for the detailed descriptions.

packages/template-ui/vue.config.js
danigm: Good idea, with the comment we can remove this once we can get rid of IE. I don't know exactly why this is failing, because the rules are in the css file, but who knows.

PR #81, opened 2021-05-28
template: Route to a specific content or topic

manuq:
The route to a topic was already implemented, but it was broken
because the routes changed. Also, add route to content. If both are
given, content takes precedence.

https://phabricator.endlessm.com/T31936

### DIFF
--- a/packages/template-ui/src/App.vue
+++ b/packages/template-ui/src/App.vue
@@ -58,9 +58,16 @@ export default {
       this.$store.commit('setHomeNavigation');
       const uri = window.location.search.substring(1);
       const params = new URLSearchParams(uri);
+      // Check if we need to navigate to a specific content or topic. Content takes precedence.
+      const contentId = params.get('contentId');
+      if (contentId) {
+        this.$router.push(`/c/${contentId}`);
+        return;
+      }
       const topicId = params.get('topicId');
       if (topicId) {
-        this.$router.push(`/${topicId}`);
+        this.$router.push(`/t/${topicId}`);
+        return;
       }
     },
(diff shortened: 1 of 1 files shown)


packages/template-ui/src/App.vue
```diff
@@ -58,9 +58,14 @@ export default {
...
       const params = new URLSearchParams(uri);
+      // Check if we need to navigate to a specific content or topic. Content takes precedence.
+      const contentId = params.get('contentId');
+      if (contentId) {
+        this.$router.push(`/c/${contentId}`);
+      }
       const topicId = params.get('topicId');
       if (topicId) {
-        this.$router.push(`/${topicId}`);
+        this.$router.push(`/t/${topicId}`);
```
dylanmccall, line 68: It looks like this is going to behave differently from the comment (and some poking with Firefox's dev tools seems to confirm it). If there's both a `contentId` and a `topicId`, we're going to call `this.$router.push` _twice_. Perhaps this would be better as an `if / else if`?

dylanmccall (changes requested):

packages/template-ui/src/App.vue
danigm: I don't know if the case of passing the `topicId` and the `contentId` is something that could happen, but I don't think so.

manuq: It should never happen, but it's a very good point. I will make it an if/else if.

manuq: Done, I prefer early return than nested conditionals.

PR #82, opened 2021-05-28
Discovery page: Don't use branding colors in lightbox

manuq:
Done hardcoding color codes for now, maintaining the defaults from
vuex.  Because this will be migrated soon to Bootstrap theme anyways.

https://phabricator.endlessm.com/T32023


dylanmccall:
Makes sense to me. Ideally we'll want to swap this component with a [bootstrap modal](https://bootstrap-vue.org/docs/components/modal) anyway :)

PR #84, opened 2021-05-31
template: Add navbar to template

danigm:
https://phabricator.endlessm.com/T31933

### DIFF
--- a/packages/eos-components/src/components/Header.vue
+++ b/packages/eos-components/src/components/Header.vue
@@ -1,5 +1,5 @@
 <template>
-  <b-navbar class="header" sticky>
+  <b-navbar class="header" fixed="top">
     <b-container>
       <img class="logo" :src="logo" @click="$emit('click-logo')">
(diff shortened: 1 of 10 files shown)


packages/eos-components/src/components/Header.vue
```diff
@@ -30,16 +30,22 @@
...
+    height: $navbar-height;
   }
 
   .header > .container {
     padding: $card-spacer-x;
   }
 
   $logo-size: 50px;
+  // 2.27 is the AR of the endless logo
+  $logo-height: $logo-size / 2.27;
```
manuq, line 42: Nit: if we now have $logo-height then the $logo-size is actually a $logo-width. Good variable names are also documentation, so you could do:
```
  $logo-width: 50px;
  $logo-aspect-ratio: 2.27;
  $logo-height: $logo-width / logo-aspect-ratio;
```

manuq (approved):
This is fantastic.

PR #85, opened 2021-06-01
T31894 assessments

manuq:
https://phabricator.endlessm.com/T31894


dylanmccall (approved):
I understand this is basically the same as the upstream plugin. So, seems logical to me :)

PR #87, opened 2021-06-01
Use yarn scripts instead of Makefile

manuq:
Add the following new yarn scripts:

- yarn build-dist
- yarn bump-version
- yarn release

Move most of the cleanup commands in `make dist` to a
`scripts/clean.sh` script, which is called by `yarn build-dist`.

Also, remove the Makefile. And update the README accordingly.

https://phabricator.endlessm.com/T32041


danigm (approved):
Great improvement, it's simpler without the Makefile.

PR #88, opened 2021-06-02
T31949 search polish

manuq:
https://phabricator.endlessm.com/T31949

### DIFF
--- a/packages/template-ui/src/constants.js
+++ b/packages/template-ui/src/constants.js
@@ -12,5 +12,7 @@ export const StructuredTags = {
   TYPE: 'type',
 };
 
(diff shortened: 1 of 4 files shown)


danigm (approved):
Looks Good To Me. I'll need to move some of this code to the `eos-components` to be able to render cards correctly on the discovery page, but I'll try to maintain this structure so this code doesn't change a lot.

manuq:
> Looks Good To Me. I'll need to move some of this code to the `eos-components` to be able to render cards correctly on the discovery page, but I'll try to maintain this structure so this code doesn't change a lot.

Maybe we should parse the nodes and add the structured tags at the same time we fetch tags from the backend?

PR #89, opened 2021-06-02
Discovery: Add SearchPage

danigm:
This new page uses the same SearchBar component that's used on custom
channel apps.

The search is done using the kolibri backend, so there's no filtering
operations done on the frontend.

This first implementation just shows the channels that match the query
but not the content cards, that's something that will be done in
following commits.

The search page shows also a list of recommended channels. Right now
these channels are choosen randomly.

https://phabricator.endlessm.com/T31938


kolibri_explore_plugin/assets/src/views/SearchPage.vue
```diff
@@ -0,0 +1,167 @@
...
+      }),
+      recommended() {
+        if (!this.channels || !this.channels.length) {
+          return null;
+        }
+
+        const channels = [];
+        const allChannels = [...this.channels];
+        while (channels.length < 3 && allChannels.length > 0) {
+          const index = _.random(0, allChannels.length - 1);
```
manuq, line 115: Please add a FIXME comment explaining that these random channels are a placeholder.

kolibri_explore_plugin/assets/src/views/SearchPage.vue
```diff
@@ -0,0 +1,167 @@
...
+  .header {
+    background: white;
+  }
+
+  .search-page {
+    padding-top: $navbar-height;
+  }
+
+  .recommended {
+    background-color: $gray-300;
```
manuq, line 223: Oh I used gray-200 for the theme body. I need to check Figma again. In any case, this should be `$body-bg`.

packages/eos-components/src/components/SearchBar.vue
```diff
@@ -4,7 +4,8 @@
       <b-input-group>
         <template #prepend>
           <b-input-group-text>
-            <b-icon-search />
+            <b-spinner v-if="loading" small label="Spinning" />
+            <b-icon-search v-else />
```
manuq, line 8: :+1: great!

packages/eos-components/src/components/SearchBar.vue
```diff
@@ -16,6 +17,7 @@
           ref="searchInput"
           class="form-control"
           placeholder="Search"
+          :disabled="loading"
```
manuq, line 20: :+1: !

packages/eos-components/src/components/SearchBar.vue
```diff
@@ -34,9 +36,17 @@
...
+        default: 500,
+      },
+      loading: {
+        type: Boolean,
+        default: false,
+      },
     },
     created() {
-      this.updateValue = _.debounce(this.inputUpdated, 500);
+      this.updateValue = _.debounce(this.inputUpdated, this.debounce);
```
manuq, line 49: :+1: !

packages/eos-components/src/components/Card.vue
```diff
@@ -33,6 +32,19 @@ export default {
           return 'RegularCard';
       }
     },
+    subtitle() {
+      let fallback = '';
+      if (this.$store) {
+        const { state } = this.$store;
```
manuq, line 38: Oh, this is a great way to use the vuex store from the reusable components. I've been investigating another approach: https://dev.to/blankazucenalg/como-utilizar-un-proyecto-de-vue-como-biblioteca-de-componentes-4ko

packages/eos-components/src/components/CardBody.vue
```diff
@@ -32,8 +32,7 @@
 
 <script>
 import VClamp from 'vue-clamp';
-import { mapGetters } from 'vuex';
-import { StructuredTags } from '@/constants';
+import { StructuredTags, getFirstStructuredTag } from './utils';
```
manuq, line 35: StructuredTags is a constant so it should go to the new `eos-components/src/constants.js`.

packages/eos-components/src/components/CollapsibleCardGrid.vue
```diff
@@ -1,12 +1,20 @@
 <template>
   <span>
-    <GridPage :nodes="nodes.slice(0, itemsPerPage)" />
+    <GridPage
+      :nodes="nodes.slice(0, itemsPerPage)"
+      :cardColumns="cardColumns"
+      :mediaQuality="mediaQuality"
```
manuq, line 6: :+1:

packages/eos-components/src/components/PaginatedCardGrid.vue
```diff
@@ -42,11 +39,14 @@ export default {
     },
   },
   watch: {
-    filterQuery() {
-      // Change to first page if the filter changes. This is needed because the
-      // number of cards to show will change and the previous pagination can be
-      // wrong.
-      this.currentPage = 1;
+    '$store.state.filters.query': {
+      deep: true,
```
manuq, line 43: Great to know about the use of deep here, and watching the store directly.

packages/eos-components/src/components/utils.js
```diff
@@ -0,0 +1,103 @@
...
+
+/** Structured tags **/
+
+export const StructuredTags = {
+  TOPIC: 'topic',
+  SUBJECT: 'subject',
+  LEVEL: 'level',
+  GRADE: 'grade',
+  TYPE: 'type',
+};
```
manuq: This should go to constants.

packages/eos-components/src/components/utils.js
```diff
@@ -0,0 +1,103 @@
...
+
+export const StructuredTagsRegExp = new RegExp('(.*)=(.*)');
```
manuq, line 14: This regex is a constant as well, so it should go to constants.

packages/eos-components/src/main.js
```diff
@@ -1,13 +1,45 @@
...
+  HighResCard,
+  LowResCard,
+  PaginatedCardGrid,
+  RegularCard,
   SearchBar,
+  SlidableCardGrid,
+  TopicCard,
+  cardMixin,
+  utils,
+  constants,
```
manuq, line 44: It's amazing the amount of components you managed to migrate. Thanks!

kolibri_explore_plugin/assets/src/apiResources.js
```diff
@@ -37,3 +37,9 @@ export const ContentNodeResource = new Resource({
     return this.fetchDetailCollection('next_steps', Store.getters.currentUserId, getParams);
   },
 });
+
+export const ContentNodeSearchResource = new Resource({
+  name: 'customcontentnodesearch',
+  // useContentCacheKey: true,
```
manuq, line 43: Nit: commented line.
```suggestion
```

kolibri_explore_plugin/assets/src/app.js
```diff
@@ -12,6 +12,19 @@ Vue.use(IconsPlugin);
 
 Vue.config.productionTip = false;
 
+/** Dynamic load of all eos-components **/
+const requireContext = require.context('eos-components/src/components', false, /.vue$/);
```
manuq, line 16: Why do we need dynamic reload in the Discovery page? We don't have overrides here. If it's just for having the components as globals and not importing them all the time, I think it's better to do a similar thing as Bootstrap-vue does, which is implement a vuejs plugin and then do:

```
Vue.use(EosComponents);
```

The `require.context` is a trick we do only for overrides, and has downsides. For instance, webpack tree shaking won't cut the unused components, making the build bigger than needed.

kolibri_explore_plugin/assets/src/views/AssessmentWrapper/index.vue
```diff
@@ -224,13 +225,11 @@ oriented data synchronization.
               } else if (this.correct === 0 && this.hintWasTaken) {
                 // not correct and hint
                 return 'hint';
-              } else {
-                // not correct and no hint
-                return 'wrong';
```
manuq, line 229: Can you explain in the commit message what are the changes in AssessmentWrapper about? Is this related to adding content cards to the search result?

kolibri_explore_plugin/assets/src/views/SearchPage.vue
```diff
@@ -41,9 +41,18 @@
...
-              </h4>
+              <b-row>
+                <b-col>
+                  <h4 class="text-muted">
+                    {{ channelTitle(channelId) }}
+                  </h4>
+                </b-col>
+                <b-col class="text-primary text-right">
+                  <b-link @click="goToChannel(channelId)">
+                    go to channel <b-icon-arrow-right />
```
manuq, line 52: :+1

kolibri_explore_plugin/assets/src/modules/topicsTree/index.js
```diff
@@ -10,6 +10,8 @@ function defaultState() {
   return {
     channel: {},
     content: {},
+    contentId: null,
+    topicId: null,
```
manuq, line 14: Can you explain in the commit message why are these new states needed? I think I get it, but it can be confusing as at first glance they may be derived from the state above. Ie. contentId can be content.id. Or if it's an ID for a different content, then we need a more descriptive name for the variable.

manuq (approved):
Looks great! And is working flawlessly, so feel free to merge. Thanks for the super detailed messages in the first commits. I would welcome the same attention to detail in the last commits.

kolibri_explore_plugin/assets/src/views/AssessmentWrapper/index.vue
danigm: That should be an automatic linter change, I didn't remember updating that, so it should be a change that's done during my commits after running the `yarn run lint-frontend:format` and I added it. It's not related to this task and we can remove these changes from the PR.

danigm: I've removed this change.

kolibri_explore_plugin/assets/src/modules/topicsTree/index.js
danigm: I've updated this code to use a unique variable and I think now is clear. The idea was to be able to link to content and topic easily without messing around with the existing app, because for the custom app presentation, the content points to the channel node, so it's not possible to reuse that.

kolibri_explore_plugin/assets/src/app.js
```diff
@@ -12,18 +13,7 @@ Vue.use(IconsPlugin);
...
-requireContext.keys().forEach(fileName => {
-  // Remove './' prefix and '.vue' suffix from file name:
-  const componentName = fileName.slice(2, -4);
-  if (componentName in Vue.options.components) {
-    return;
-  }
-  const component = requireContext(fileName).default;
-  Vue.component(componentName, component);
-});
+Vue.use(eosComponents);
```
manuq, line 16: Excellent!

kolibri_explore_plugin/assets/src/views/SearchPage.vue
```diff
@@ -142,10 +142,16 @@
...
-        // Add thumbnails to nodes
+        // Add thumbnails and links to nodes
         nodes.forEach(node => {
           const thumbnailUrl = getContentNodeThumbnail(node);
           node.thumbnail = thumbnailUrl;
+          const base = `/topics/${node.channel_id}`;
+          if (node.kind === 'topic') {
+            node.nodeUrl = `${base}/t/${node.id}`;
+          } else {
+            node.nodeUrl = `${base}/c/${node.id}`;
```
manuq, line 153: :+1:

manuq (approved):

PR #99, opened 2021-06-08
components: Prevent bubbling click of PlayButtons

manuq:
Prevent bubbling the click event of PlayButtons using the @click.stop
modifier.

Also, use a click event handler instead of the `:to` in the
ContentLink, which is actually rendered as a
`<router-link>`. Otherwise the event is not stopped.

https://phabricator.endlessm.com/T32063

### DIFF
--- a/packages/eos-components/src/components/PlayButton.vue
+++ b/packages/eos-components/src/components/PlayButton.vue
@@ -3,7 +3,7 @@
     pill
     variant="light"
     :class="`card-media-type m-2 ${node.kind}`"
-    @click="$emit('click')"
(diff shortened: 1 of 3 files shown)


packages/eos-components/src/components/PlayButton.vue
```diff
@@ -3,7 +3,7 @@
     pill
     variant="light"
     :class="`card-media-type m-2 ${node.kind}`"
-    @click="$emit('click')"
+    @click.stop="$emit('click')"
```
danigm, line 6: 👍

danigm (approved):
Looks good to me.

PR #100, opened 2021-06-09
T32072 channel icons

danigm:
https://phabricator.endlessm.com/T32072

### DIFF
--- a/packages/eos-components/src/components/ChannelLogo.vue
+++ b/packages/eos-components/src/components/ChannelLogo.vue
@@ -1,0 +1,86 @@
+<template>
+  <div
+    class="thumbnail"
+    :class="size"
+    :style="thumbnailStyles"
+  >
+  </div>
+</template>
+
+<script>
+  export default {
+    name: 'ChannelLogo',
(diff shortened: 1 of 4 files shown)


packages/eos-components/src/components/ChannelLogo.vue
```diff
@@ -0,0 +1,115 @@
...
+        const ctx = canvas.getContext('2d');
+        const img = new Image();
+        img.onload = () => {
+          ctx.drawImage(img, 0, 0);
+          const { data } = ctx.getImageData(0, 0, 1, 1);
+          const [,,, alpha] = data;
+          this.hasTransparentBg = (alpha === 0);
+        };
+        img.src = this.channel.thumbnail;
+      },
```
manuq, line 64: The logic looks great. But I thought we talked about not doing this for performance reasons. Also I'm not sure if this 80% resize will work for all the icons. In any case since you already implemented it we can merge it and see in the online instance. I'm really against fixing assets with code.

manuq (approved):
I left suggestions.

PR #101, opened 2021-06-09
T32068 footer design

danigm:
https://phabricator.endlessm.com/T32068


kolibri_explore_plugin/assets/src/views/ChannelsPage.vue
```diff
@@ -49,35 +49,31 @@
...
-              target="_blank"
-              href="https://www.endlessos.org/key"
-            >
-              Learn more <b-icon-chevron-right />
-            </b-button>
-          </b-col>
-        </b-row>
-      </b-container>
-    </div>
+    <Footer>
```
manuq, line 52: Great to have the footer as a shared component!

kolibri_explore_plugin/assets/src/customApps.js
```diff
@@ -28,20 +28,39 @@ export const ThumbApps = [
   'healthy-living',
 ];
 
+export function getAppNameByID(id) {
+  return CustomChannelApps[id] || 'default';
+}
+
+function normalizeChannelName(channel) {
```
manuq, line 35: Let's reuse the code we already have this time, because the logic here is slightly different than in the template. So I would move the template `getSlug()` to the components `utils`, and import it here.

kolibri_explore_plugin/assets/src/views/CustomChannelPresentationApp.vue
```diff
@@ -31,11 +31,12 @@
       rooturl() {
         const app = getAppNameByID(this.channel.id);
         const url = urls['kolibri:kolibri_explore_plugin:app_custom_presentation']({ app: app });
+        const endlessApp = isEndlessApp(this.channel) ? 'endless=true' : '';
         if (this.customAppContent.id) {
           const { kind, id } = this.customAppContent;
-          return `${url}?${kind}Id=${id}`;
+          return `${url}?${kind}Id=${id}&${endlessApp}`;
```
manuq, line 37: This is sending a new parameter to the custom channel presentation, which wasn't discussed in Learning Equality upstream API (the customApps.js file). The other option is to do it as a template override setting, in the `options.json` of each override. For instance in packages/template-ui/channel-overrides/healthy-living/options.json:

```
{
  "displayLogoInHeader": false,
  "isEndlessApp": true
}
```

manuq (changes requested):
Looks good! I guess we can go with the `customApps.js` for now if you want. But let's try to reuse the normalize/slug function.

manuq (approved):

PR #103, opened 2021-06-09
T32065 fonts

manuq:
https://phabricator.endlessm.com/T32065


danigm (approved):
This is really nice wit the fontsource dependency, great work.

PR #108, opened 2021-06-11
Template: Do not include topics in filtering result

danigm:
https://phabricator.endlessm.com/T31988


manuq (approved):
Great, this is working as expected and looks good!

PR #109, opened 2021-06-11
T32024 lightbox bootstrap

manuq:
https://phabricator.endlessm.com/T32024


kolibri_explore_plugin/assets/src/views/AssessmentWrapper/ExerciseAttempts/AnswerIcon.vue
```diff
@@ -4,33 +4,34 @@
     <div ref="icon" class="icon">
       <span
         v-if="answer === 'right'"
+        v-b-tooltip:lightbox.hover.right
```
dylanmccall, line 7: This is making a very strong assumption about being inside a container with the ID "lightbox", which seems like it could be trouble down the line. Can we pass this through as a prop (or provide / inject)?

dylanmccall (commented):
Yay, lots of code being removed! :)

Is there anything stopping us from actually using [bootstrap-vue's Modal component](https://bootstrap-vue.org/docs/components/modal) here? From a cursory glance I imagine that would be preferable to building our own at this point.

kolibri_explore_plugin/assets/src/views/ContentLightbox.vue
```diff
@@ -94,6 +87,8 @@
     width: 85%;
     max-height: calc(100vh - 2rem);
     margin: 0 auto;
+    color: $white;
```
danigm, line 90: There's a problem with this style, it's applied to the lightbox content, for example in some `practice` the question is not visible because it's white text over white background.

imagen

danigm (changes requested):
Looks really good. I've just found a little problem with exercise rendering inside the lightbox, but the rest looks good.

kolibri_explore_plugin/assets/src/views/AssessmentWrapper/ExerciseAttempts/AnswerIcon.vue
manuq: Yes actually if we use a bootstrap-vue modal this won't be necessary at all.

kolibri_explore_plugin/assets/src/views/ContentLightbox.vue
manuq: Good catch! I've been trying only one exercise and I thought that area was just empty. I will check with other examples.

manuq:
Ready for review again. For the bootstrap-vue modal I will try it in a separate PR on top of this one. I like the idea but I'm a bit skeptical because we want custom things like getting rid of the letterboxing.

danigm (approved):

PR #113, opened 2021-06-14
T31944 cards row

manuq:
https://phabricator.endlessm.com/T31944


danigm (approved):
LGTM, All looks more consistent with these new styles.

PR #115, opened 2021-06-15
T32016 fix build

manuq:
https://phabricator.endlessm.com/T32016

### DIFF
--- a/packages/eos-components/src/styles.scss
+++ b/packages/eos-components/src/styles.scss
@@ -56,7 +56,7 @@ $carousel-indicator-hit-area-height: 10px !default;
 $carousel-indicator-spacer:          3px !default;
 $carousel-indicator-transition:      all .6s ease !default;
 
-@import 'channel-overrides';
+@import "../../template-ui/src/overrides/styles.scss";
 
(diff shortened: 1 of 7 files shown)


packages/eos-components/src/styles.scss
```diff
@@ -56,7 +56,7 @@ $carousel-indicator-hit-area-height: 10px !default;
 $carousel-indicator-spacer:          3px !default;
 $carousel-indicator-transition:      all .6s ease !default;
 
-@import 'channel-overrides';
+@import "../../template-ui/src/overrides/styles.scss";
```
danigm, line 59: this will work, but we'll have the styles override from the latest channel override, right now I think it's sikana. Maybe we should override that file after the last custom channel is built.

danigm (changes requested):

packages/eos-components/src/styles.scss
manuq: Good catch! I just appended a commit with the fix.

danigm (approved):
Looks good, let's merge this.

PR #117, opened 2021-06-15
Discovery: Add development version info tag

danigm:
https://phabricator.endlessm.com/T32125

### DIFF
--- a/kolibri_explore_plugin/assets/src/views/DevTag.vue
+++ b/kolibri_explore_plugin/assets/src/views/DevTag.vue
@@ -1,0 +1,73 @@
+<template>
+
+  <div class="dev fixed-top mx-1 text-right">
+    <span v-b-modal.buildinfo class="badge badge-warning">{{ title }}</span>
+
+    <b-modal v-if="info" id="buildinfo" title="Build Info" okOnly>
+      <b-list-group>
+        <b-list-group-item>Commit: {{ info.commit }}</b-list-group-item>
+        <b-list-group-item>Date: {{ info.date }}</b-list-group-item>
+        <b-list-group-item>Last Release: {{ info.last_release }}</b-list-group-item>
+      </b-list-group>
+      <h3>Log</h3>
+      <b-list-group>
+        <b-list-group-item v-for="commit in info.log" :key="commit.commit">
+          {{ commit.subject }} <br> <i>{{ commit.author }}</i>
+        </b-list-group-item>
+      </b-list-group>
+    </b-modal>
+  </div>
+
+</template>
+
+
+<script>
+
+  import urls from 'kolibri.urls';
+
+  export default {
+    name: 'DevTag',
+    data() {
+      return {
+        info: null,
+      };
+    },
+    computed: {
+      title() {
+        if (this.info) {
+          return this.info.commit;
+        }
+
+        return 'loading...';
+      },
+    },
+    mounted() {
+      this.getBuildInfo();
+    },
+    methods: {
+      getBuildInfo() {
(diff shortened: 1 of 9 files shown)


kolibri_explore_plugin/assets/src/views/DevTag.vue
```diff
@@ -0,0 +1,75 @@
...
+
+  export default {
+    name: 'DevTag',
+    data() {
+      return {
+        info: null,
+      };
+    },
+    computed: {
+      dev() {
```
manuq, line 35: Please use better names. For booleans is always better to use a prefix like "isFoo", "hasBar", "canBaz". And avoid abbreviations. Maybe "isDevelopmentBuild"?

scripts/build_info.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
+
+HEAD=$(git rev-parse --verify HEAD --short)
+DATE=$(date +%Y-%m-%d)
+TIME=$(date +%H:%M:%S)
+COMMITS=$(git log --no-merges --pretty=format:'%h|%s|%aN' -n 10)
+DEV_INFO=${DEV_INFO:-false}
```
manuq, line 7: I guess this is the environment variable that does the magic? Is not explained in the commit message.

scripts/build_info.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
+
+HEAD=$(git rev-parse --verify HEAD --short)
+DATE=$(date +%Y-%m-%d)
+TIME=$(date +%H:%M:%S)
```
manuq, line 5: This is date and time from the system. I would use the information from the latest commit.

scripts/build_info.sh
```diff
@@ -0,0 +1,31 @@
...
+COMMITS=$(git log --no-merges --pretty=format:'%h|%s|%aN' -n 10)
```
manuq, line 6: Showing the last 10 is rather arbitrary. What about showing the commits since the last release?

manuq (changes requested):
The commit message is empty! There are several things you can comment in the commit message to make the review easy and for future maintenance. For example:

- Explain in a bit more detail what are the changes about. There seem to be multiple changes, from creating a JSON file at build time to adding a new component to the frontend.
- What means that a build is "dev".
- What is the approach you are taking, how are you storing the information.
- What files are you adding and why.

Also it would be nice to update the README with instructions for making a "development" build. Looking at the code it seems to be an special environment variable.

Also it would be better to split in smaller commits. At least frontend change to build changes.

Here is the approach I would take:
- Always include the commit info in the build. Something like this output https://www.npmjs.com/package/git-describe#example-output .
- Instead of having the "dev" boolean in the JSON, use an environment variable where the Kolibri server runs to display the component with the information.

But I may be wrong since I had to guess what the code is doing :)

kolibri_explore_plugin/assets/src/views/DevTag.vue
danigm: I've used `showBuildInfo` that's more descriptive.

scripts/build_info.sh
danigm: I've changed the environment variable needed to `SHOW_BUILD_INFO` and add it to the commit message to it's more clear.

scripts/build_info.sh
danigm: Good point, I was thinking about it like the "build time", but it's maybe more interesting the "commit time".

manuq (approved):
Great update! I have 2 last suggestions, feel free to take them. One is inline and the other here: I would add a "How to display the build information" section in the README.

PR #119, opened 2021-06-16
T32136 carousel

manuq:
https://phabricator.endlessm.com/T32136


danigm (approved):
LGTM. 👍 The utils is better outside the components

PR #120, opened 2021-06-16
Discovery page: Add more negative margin to the iframe

manuq:
One pixel is enough to remove the double scrollbar. Ideally we should
find why this negative margin is needed, but we are not there yet.

https://phabricator.endlessm.com/T32139


danigm (approved):
LGTM

PR #121, opened 2021-06-17
T32082 filter dropdown

danigm:
https://phabricator.endlessm.com/T32082

### DIFF
--- a/packages/template-ui/src/App.vue
+++ b/packages/template-ui/src/App.vue
@@ -1,7 +1,7 @@
 <template>
   <div id="app" class="d-flex flex-column h-100">
     <!-- Wrapper needed to fix flexbox footer positioning on IE11 -->
-    <div class="flex-shrink-0">
+    <div class="flex-fill flex-shrink-0">
       <ChannelNavBar />
       <router-view>
         <ChannelHeader />
@@ -77,18 +77,18 @@ export default {
 <style lang="scss">
 @import '@/styles.scss';
 
-html,
-body {
+html {
   height: 100%;
 }
-
 /* Always show the vertical scrollbar */
 body {
   overflow-y: scroll;
+  min-height: 100vh;
 }
 
 #app {
   -webkit-font-smoothing: antialiased;
   -moz-osx-font-smoothing: grayscale;
+  min-height: 100vh;
 }
 </style>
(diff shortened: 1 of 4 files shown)


packages/template-ui/src/components/FilterDropDown.vue
```diff
@@ -0,0 +1,88 @@
...
+</script>
+
+<style lang="scss" scoped>
+  @import '@/styles.scss';
+
+  .filter-dropdown {
+    max-width: 720px;
+    max-height: 480px;
+    min-width: 380px;
+    min-height: 185px;
```
manuq, line 84: The dropdown size is all hardcoded! Can we agree in not having hardcoded values anymore? This is also not responsive. I think it should fit the 100% width in mobile. And for other breakpoints it may be a % of the breakpoint width.

About the height, why limit it?

manuq (changes requested):
The dropdown size is all hardcoded in pixels so I stopped my review in the first commit, I'm not reviewing the rest for now. It works very well though!

packages/template-ui/src/App.vue
```diff
@@ -77,18 +77,16 @@ export default {
...
 
-html,
-body {
+html {
   height: 100%;
 }
-
-/* Always show the vertical scrollbar */
 body {
-  overflow-y: scroll;
```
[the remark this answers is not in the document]
danigm: my bad, I removed this as part of my testing to try to fix the dropdown positioning and never bring back.

packages/template-ui/src/components/FilterDropDown.vue
danigm: These are the values that are in the spec. I don't know how to handle these specs, if I should try to follow as much as possible, or if we can just adapt the spacing to existing components.

I've just updated to follow your suggestion, just remove the min-height and just limit the width using responsive breakpoints.

I've left the `max-height` limit because in other case we won't see the vertical scrollbar when there are a lot of tags, just the dropdown get bigger and can get the full height.

manuq: Yes, I understand you. We as engineers are responsible for the technical part, so we should advice Joana if something isn't specified following good practices. Please always try to reach Joana with questions or suggestions for better implementations. I do, and she hears. She is always open to suggestions.

Also there is a bit of interpretation any time you read a spec. For instance:
- We use REM units, not pixels (Bootstrap does). But you can assume that 1rem == 16px, so you can do the conversion yourself.
- For the height maybe she meant to fix N number of rows, then start scrolling? (ask her) If so we can do the calc in the SCSS, which is better for maintenance. If the button size changes in the future the hardcoded height will break.

In the worst case she may say "I really want this box to be this min and max size, regardless of the responsiveness", in which case we should at least move  the values to the styles file, so they are documented and overrides can override them.

PR #123, opened 2021-06-18
Discovery: Add ContentModal component

danigm:
This new component is the content lightbox and it's added to the base
ExploreIndex component, so the play button in cards will work
everywhere.

This patch also includes some improvements to the lightbox experience:

 * Convert ContentLink to span, using the role attribute, to avoid the
   page to scroll to top on every click
 * Body height css adjustment to make it take the whole page and not the
   "100%" that produced some issues with the scroll.
 * Remove the right padding from the body when the lightbox is open in
   discovery page.
 * Make the lightbox scrollable so when the content is larger the
   scroll is inside the ligthbox and not another external scroll.

https://phabricator.endlessm.com/T32137


kolibri_explore_plugin/assets/src/views/ExploreIndex.vue
```diff
@@ -17,22 +18,23 @@
...
   import SearchPage from './SearchPage';
   import DevTag from './DevTag';
+  import ContentModal from './ContentModal';
 
   const pageNameToComponentMap = {
     [PageNames.TOPICS_ROOT]: ChannelsPage,
-    [PageNames.TOPICS_CHANNEL]: CustomChannelsPage,
-    [PageNames.TOPICS_TOPIC]: CustomChannelsPage,
+    [PageNames.TOPICS_CHANNEL]: CustomChannelPresentationApp,
+    [PageNames.TOPICS_TOPIC]: CustomChannelPresentationApp,
```
manuq, line 30: :+1:

kolibri_explore_plugin/assets/src/views/ContentModal.vue
```diff
@@ -0,0 +1,110 @@
+<template>
+
+  <b-modal
+    id="content-modal"
```
manuq, line 4: Great, is better to have the modal as a separate component.

kolibri_explore_plugin/assets/src/views/CustomChannelPresentationApp.vue
```diff
@@ -1,27 +1,28 @@
 <template>
 
-  <iframe
-    ref="iframe"
-    class="custom-presentation-iframe"
-    frameBorder="0"
-    :src="rooturl"
-  >
-  </iframe>
+  <div>
```
manuq, line 3: I don't see why this new div wapping the iframe is needed. Maybe for the scroll?

kolibri_explore_plugin/assets/src/views/ExploreIndex.vue
```diff
@@ -89,9 +91,19 @@
...
   body,
   body > div {
-    height: 100%;
+    min-height: 100vh;
+  }
+
+  // Remove the padding added by the modal
+  // https://stackoverflow.com/questions/32862394/bootstrap-modals-keep-adding-padding-right-to-body-after-closed
+  body.modal-open {
+    padding-right: 0 !important;
```
manuq, line 106: Great, thanks for the reference!

kolibri_explore_plugin/assets/src/views/SearchPage.vue
```diff
@@ -1,6 +1,6 @@
 <template>
 
-  <div class="d-flex flex-column h-100 search-page">
```
manuq, line 3: There is a helper in bootstrap for min-height 100vh: `min-vh-100`.
https://getbootstrap.com/docs/5.0/utilities/sizing/#relative-to-the-viewport

packages/eos-components/src/components/ContentLink.vue
```diff
@@ -1,11 +1,12 @@
 <template>
-  <b-link
+  <span
     v-b-hover="handleHover"
+    role="button"
     class="text-decoration-none text-reset"
-    @click="onClick"
+    @click.stop="onClick"
```
manuq, line 6: :+1:

packages/eos-components/src/components/ContentLink.vue
```diff
@@ -1,11 +1,12 @@
 <template>
-  <b-link
+  <span
```
manuq, line 2: :+1:

manuq (approved):
Looks great, I left a nitpick about the bootstrap helper class for sizing. Feel free to merge.

kolibri_explore_plugin/assets/src/views/CustomChannelPresentationApp.vue
danigm: It's not needed at all, it's there because at first I added the new `ContentModal` component inside this, instead of in the `ExploreIndex` one and I forget to remove the div wrapper.

kolibri_explore_plugin/assets/src/views/SearchPage.vue
danigm: Thanks! I'll use that instead of the manual css.

PR #125, opened 2021-06-21
scripts: Check the TAG release not the TAG commit

danigm:
The yarn bump-version command creates a tag with -a so a new object is
created, so the command:

git ls-remote https://github.com/endlessm/kolibri-explore-plugin.git $TAG

Returns the hash of the new tag object and not the head that we want. To
do it correctly we should query the $TAG^{} that resolves to the real
commit pointed by the tag, not to the tag object.

https://phabricator.endlessm.com/T32041

### DIFF
--- a/scripts/check_can_relese.sh
+++ b/scripts/check_can_relese.sh
@@ -10,7 +10,7 @@ then
     exit 1
 fi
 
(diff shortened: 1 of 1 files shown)


manuq (approved):
Great, this is the part I didn't test. Thanks!

PR #126, opened 2021-06-22
Discovery: Make card grid responsive

danigm:
This patch makes the discovery page card grid and search grids
responsive, showing three cards for large screens, two for small and
one for really small screens.

A new mixin was introduced to calculate the column width and also to
change the width data in the component on resize, so it can be used to
change inside components properties.

The bootstrap sass breakpoints are also exposed in the styles.scss so it
can be used inside the javascript to determine at witch point to change
the number of elements to show.

https://phabricator.endlessm.com/T32160

### DIFF
--- a/packages/eos-components/src/components/SlidableCardGrid.vue
+++ b/packages/eos-components/src/components/SlidableCardGrid.vue
@@ -60,9 +60,11 @@
 <script>
 import _ from 'underscore';
 import { MediaQuality } from '../constants';
+import responsiveMixin from './mixins/responsiveMixin';
 
 export default {
   name: 'SlidableCardGrid',
+  mixins: [responsiveMixin],
   props: {
     nodes: Array,
     mediaQuality: String,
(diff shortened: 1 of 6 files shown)


packages/eos-components/src/components/mixins/columnsMixin.js
```diff
@@ -0,0 +1,34 @@
...
+      if (this.width <= Number(md)) {
+        return 2;
+      }
+
+      return 3;
+    },
+  },
+  created() {
+    this.width = window.innerWidth;
+    window.addEventListener("resize", this.onResize);
```
manuq, line 24: This is the same way we implemented it for the chromeos demo. Since then I learnt about [matchMedia()](https://developer.mozilla.org/en-US/docs/Web/API/Window/matchMedia) which is more performant because it triggers the callback when the breakpoint changes. See also [this bootstrap-vue discussion](https://github.com/bootstrap-vue/bootstrap-vue/issues/5284#issuecomment-622392181).

packages/eos-components/src/components/SlidableCardGrid.vue
```diff
@@ -60,9 +60,12 @@
 <script>
 import _ from 'underscore';
 import { MediaQuality } from '../constants';
+import { sm, md } from '../styles.scss';
```
manuq, line 63: Excellent to have these exposed!

manuq (approved):
Looks good. I left a suggestion to improve the resize check with matchMedia().

packages/eos-components/src/components/mixins/columnsMixin.js
danigm: Oh, thanks for the links, I didn't know about the `matchMedia` and that's exactly what we need here.

PR #127, opened 2021-06-22
T31752 pbs

manuq:


packages/eos-components/src/styles.scss
```diff
@@ -61,6 +91,30 @@ $carousel-indicator-transition:      all .6s ease !default;
...
 
+@import 'mixins';
+
+// Change button dark outline variant:
+.btn-outline-dark {
+  $value: theme-color("dark");
+  @include eos-button-outline-variant($value, $color-hover: $value, $border-color: $gray-500, $active-background: transparent, $active-border: $primary);
+}
+// Change buttons variant:
+@each $color, $value in $theme-colors {
```
danigm, line 102: This is great! using the bootstrap scss for custom buttons is the way to go.

packages/template-ui/src/views/Test.vue
```diff
@@ -0,0 +1,85 @@
+<template>
```
danigm, line 1: This is a nice way to show the custom channel theme. Maybe we can add a link to this in the `DevTag` for each custom channel.

kolibri_explore_plugin/assets/src/customApps.js
```diff
@@ -3,7 +3,7 @@ import urls from 'kolibri.urls';
 // List of channel ids and the app name with the corresponding
 // custom-channel-presentation zip bundle
 export const CustomChannelApps = {
-  '1306b5d7d1ce4b98b0039324f368ce6a': 'pbskids',
+  '1306b5d7d1ce4b98b0039324f368ce6a': 'pbs-kids',
```
danigm, line 6: It's not part of this PR, but maybe we can think about to use the "normalized" channel name instead of the channel-id. I've used that solution for the big thumbnails and I think it's something easier to maintain and update.

danigm (approved):

packages/template-ui/src/views/Test.vue
manuq: That's a great idea! It will allow Joana to see it.

kolibri_explore_plugin/assets/src/customApps.js
manuq: Hmm. The problem with normalized names is that they are not unique. But yeah it wouldn't matter if 2 channels have the same name.

PR #141, opened 2021-06-28
template: Make navbar breadcrumb responsive

manuq:
- Show only the last parent for XS and SM breakpoints.
- Hide channel logo for XS breakpoint.

Also:

- Move the "notAtHome" conditional to the navbar.
- Remove commented style.

https://phabricator.endlessm.com/T32096

### DIFF
--- a/packages/template-ui/src/components/Breadcrumb.vue
+++ b/packages/template-ui/src/components/Breadcrumb.vue
@@ -1,28 +1,43 @@
 <template>
   <b-navbar-nav
-    v-if="(node.ancestors && node.ancestors.length) || notAtHome"
     class="mt-3"
     aria-label="breadcrumb"
   >
-    <ChannelLogo class="mt-1" :channel="channel" size="sm" />
+    <ChannelLogo class="d-none d-sm-block mt-1" :channel="channel" size="sm" />
     <ol
       v-if="node.ancestors && node.ancestors.length"
       class="bg-transparent breadcrumb px-2"
     >
-      <li
-        v-for="a in node.ancestors"
-        :key="a.id"
-        class="breadcrumb-item"
-      >
-        <b-link
-          :to="getTopicUrl(a)"
+      <template v-if="xs || sm">
+        <li v-if="node.ancestors.length > 1" class="active breadcrumb-item">
+          ...
+        </li>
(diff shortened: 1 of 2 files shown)


packages/template-ui/src/components/Breadcrumb.vue
```diff
@@ -1,28 +1,43 @@
 <template>
   <b-navbar-nav
-    v-if="(node.ancestors && node.ancestors.length) || notAtHome"
     class="mt-3"
     aria-label="breadcrumb"
   >
-    <ChannelLogo class="mt-1" :channel="channel" size="sm" />
+    <ChannelLogo v-if="!xs" class="mt-1" :channel="channel" size="sm" />
```
danigm, line 6: This is something that can be also done with the bootstrap [display classes](https://getbootstrap.com/docs/4.6/utilities/display/#hiding-elements).

The use of `responsiveMixin` looks simpler, but I'm not sure if we should try to use the bootstrap standard classes when it's possible.

packages/template-ui/src/components/Breadcrumb.vue
```diff
@@ -1,28 +1,43 @@
...
     >
-      <li
-        v-for="a in node.ancestors"
-        :key="a.id"
-        class="breadcrumb-item"
-      >
-        <b-link
-          :to="getTopicUrl(a)"
+      <template v-if="xs || sm">
+        <li class="active breadcrumb-item">
```
danigm, line 12: This "three dots" element is shown always, even if there's only one item in the breadcrumb, in that case we shouldn't show it.

```suggestion
        <li v-if="node.ancestors.length > 1" class="active breadcrumb-item">
```

danigm (changes requested):
We don't need to show the "⋯" when there's only one element. About the use of `responsiveMixin` or bootstrap display classes, I don't have a strong opinion, but I feel that it'll be better to use the bootstrap display classes in templates when it's possible.

packages/template-ui/src/components/Breadcrumb.vue
manuq: Oh yes! Indeed we should use the display class whenever possible. Thanks!

PR #142, opened 2021-06-29
template: Fix lint error on Section

danigm:
This change was introduced as part of the PR #130:
https://github.com/endlessm/kolibri-explore-plugin/pull/130/files#diff-188f7b897a1e20b9d5f7891fe048bc8b7868578210d30dbc5d7600575edf7199R26-R31

And the lint CI didn't complain there, but I'm getting some errors when
trying to run the development instance about "no-else-return":
https://eslint.org/docs/rules/no-else-return

This patch just fixes that issue and also adds the trailing `;` to the
two "return" lines.


manuq:
Merging. Also please make sure you are using the same dev environment as ensured by the project setup!

PR #144, opened 2021-06-29
template: Add options to remove elements from home view

manuq:
The elements that can be removed by the overrides are:

- The row with main sections and search button.
- The carousel.
- The filter row.

Also, make the PBS-Kids override use all of them.

Also, improve the margins and replace divs with templates so the
margins count.


danigm (approved):
LGTM

PR #145, opened 2021-06-29
T32075 bundles

manuq:
https://phabricator.endlessm.com/T32075

### DIFF
--- a/packages/eos-components/src/components/Card.vue
+++ b/packages/eos-components/src/components/Card.vue
@@ -3,6 +3,7 @@
     :is="cardVariant"
     :node="node"
     :subtitle="subtitle"
+    :isBundle="isBundle"
     :url="getNodeUrl(node)"
   />
 </template>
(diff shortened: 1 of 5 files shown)


danigm (approved):
LGTM. We'll have different results rendering bundled cards on the discovery page, but that's something we can't solve easily, because the information about the bundle type is in the custom template. Even the links to specific cards on the search won't "work correctly".

I think that at some point we should duplicate the "bundle" information to know that at discovery page scope.

manuq:
> LGTM. We'll have different results rendering bundled cards on the discovery page, but that's something we can't solve easily, because the information about the bundle type is in the custom template. Even the links to specific cards on the search won't "work correctly".
> 
> I think that at some point we should duplicate the "bundle" information to know that at discovery page scope.

And there is another issue: clicking on a leaf content in the discovery page will navigate to the content detail in the custom presentation `/c/ID`, not to the bundle page `/t/parentID`.

So yes we should discuss this technically. It will not be performant at all to add the "is bundle" information in the discovery page for each node. If we really want bundles, the bundle metadata should be in the backend.

For now I will summarize these issues in the ticket.

PR #146, opened 2021-06-29
components: Fix console error in Header

manuq:
For adding the shadow on scroll I first tried using an external
library that adds a v-scroll directive. I couldn't make it work so I
implemented the functionality myself. But I also left the directive
which is raising an error because it doesn't exist.


danigm (approved):
LGTM

PR #147, opened 2021-06-30
Discovery: Add loading overlay to custom channel view

danigm:
This patch adds an overlay to show a loading screen during the load of
the custom channel presentation iframe.

https://phabricator.endlessm.com/T32182

### DIFF
--- a/kolibri_explore_plugin/assets/src/views/CustomChannelPresentationApp.vue
+++ b/kolibri_explore_plugin/assets/src/views/CustomChannelPresentationApp.vue
@@ -5,6 +5,7 @@
     class="custom-presentation-iframe"
     frameBorder="0"
     :src="rooturl"
+    @load="$emit('load')"
   >
   </iframe>
 
@@ -43,6 +44,7 @@
     },
     mounted() {
       window.addEventListener('message', this.onMessage);
+      this.$emit('loading');
     },
     beforeDestroy() {
       window.removeEventListener('message', this.onMessage);
--- a/kolibri_explore_plugin/assets/src/views/ExploreIndex.vue
+++ b/kolibri_explore_plugin/assets/src/views/ExploreIndex.vue
@@ -3,7 +3,17 @@
   <div>
     <ContentModal />
     <DevTag v-if="showBuildInfo" />
-    <component :is="currentPage" v-if="currentPage" />
+    <component
+      :is="currentPage"
+      v-if="currentPage"
+      @loading="onLoading"
+      @load="onLoad"
+    />
+    <b-overlay :show="isLoading" noWrap>
+      <template #overlay>
+        <img :src="loadingImg" class="loading">
+      </template>
+    </b-overlay>
     <router-view />
   </div>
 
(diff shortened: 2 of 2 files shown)


kolibri_explore_plugin/assets/src/views/CustomChannelPresentationApp.vue
```diff
@@ -144,4 +162,20 @@
     margin-bottom: -6px;
   }
 
+  .loading {
+    animation: spin 2s infinite linear;
+  }
+
+  // Based on this css
+  // https://github.com/loadingio/loading.css/blob/master/dist/loading.css#L1643
```
manuq, line 170: Nice library of loading animations. Let's merge this one and propose the other options to Joana.

kolibri_explore_plugin/assets/src/views/CustomChannelPresentationApp.vue
```diff
@@ -1,12 +1,18 @@
 <template>
 
-  <iframe
-    ref="iframe"
-    class="custom-presentation-iframe"
-    frameBorder="0"
-    :src="rooturl"
-  >
-  </iframe>
+  <b-overlay :show="isLoading">
```
manuq, line 3: I wonder if wrapping the iframe will be problematic. The default is to wrap the overlay. But there is also a no-wrap option. Have you played with it? https://bootstrap-vue.org/docs/components/overlay#non-wrapping-mode

manuq (approved):
Looks great. I only left a suggestion to consider the no-wrap option of the b-overlay.

kolibri_explore_plugin/assets/src/views/CustomChannelPresentationApp.vue
danigm: I've tested with `noWrap` and it works correctly, maybe we can use that one. I've updated the PR to use the "non-wrapping-mode".

kolibri_explore_plugin/assets/src/views/CustomChannelPresentationApp.vue
```diff
@@ -1,12 +1,21 @@
 <template>
 
-  <iframe
-    ref="iframe"
-    class="custom-presentation-iframe"
-    frameBorder="0"
-    :src="rooturl"
-  >
-  </iframe>
+  <div>
```
manuq, line 3: Hmm but you are adding a div wrapper anyways.

manuq (changes requested):

kolibri_explore_plugin/assets/src/views/CustomChannelPresentationApp.vue
danigm: Yes, it's the only way here we've just the iframe so we should wrap or add the overlay in the parent component, that'll be a bit more complex because we should pass a callback property or use the store to show/hide.

In any case, I'll try to find a clean solution for that so there's no wrap on this component.

danigm: I've moved the overlay to the parent element and manage the state with signals.

kolibri_explore_plugin/assets/src/views/ExploreIndex.vue
```diff
@@ -3,7 +3,17 @@
...
-    <component :is="currentPage" v-if="currentPage" />
+    <component
+      :is="currentPage"
+      v-if="currentPage"
+      @loading="onLoading"
+      @load="onLoad"
+    />
+    <b-overlay :show="isLoading" noWrap>
+      <template #overlay>
+        <img :src="loadingImg" class="loading">
```
manuq, line 14: Nit: I would rename to `loadingImage`.

manuq (approved):
Excellent. Thanks for the extra effort.

PR #148, opened 2021-06-30
Components: Do not use css min property

danigm:
This property is not supported in old browsers. This patch replaces the
min property with breakpoint styles, so it's 50% for xs, sm and md and
400px for the rest.

https://phabricator.endlessm.com/T32184

### DIFF
--- a/packages/eos-components/src/components/CarouselCard.vue
+++ b/packages/eos-components/src/components/CarouselCard.vue
@@ -98,11 +98,15 @@ export default {
 }
 .card {
   border-radius: $border-radius-lg;
-  padding-left: min(400px, 50%);
   position: relative;
   background-color: $gray-700;
   color: $white;
   border: none;
+
+  padding-left: $carousel-image-width;
+  @include media-breakpoint-up(lg) {
+    padding-left: $carousel-image-lg-width;
+  }
 }
 .img {
   border-top-left-radius: $border-radius-lg;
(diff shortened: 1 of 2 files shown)


packages/eos-components/src/components/CarouselCard.vue
```diff
@@ -98,11 +98,15 @@ export default {
...
   border-radius: $border-radius-lg;
-  padding-left: min(400px, 50%);
+  padding-left: 50%;
   position: relative;
   background-color: $gray-700;
   color: $white;
   border: none;
+
+  @include media-breakpoint-up(lg) {
+    padding-left: 400px;
```
manuq, line 108: Can we also move this hardcoded 400px to the theme?

manuq (changes requested):
This is using a hardcoded value. It should be a SCSS variable in the theme. It dates back this change, but we should take the opportunity to remove the hardcoded values.

packages/eos-components/src/components/CarouselCard.vue
danigm: I've moved these values tot he theme styles.

packages/eos-components/src/styles.scss
```diff
@@ -117,6 +117,9 @@ $carousel-indicator-hit-area-height: 10px !default;
 $carousel-indicator-spacer:          3px !default;
 $carousel-indicator-transition:      all .6s ease !default;
 
+$carousel-image-width:    50%;
+$carousel-image-lg-width: 400px;
```
manuq, line 121: Excellent.

manuq (approved):

PR #151, opened 2021-06-30
Discovery page: Fix offset in html5_viewer icon

manuq:
The header bar with the fullscreen icon is not inside an iframe, so
our CSS affects it. This is a workaround to compensate the offset.

https://phabricator.endlessm.com/T32186

### DIFF
--- a/kolibri_explore_plugin/assets/src/views/ContentItem.vue
+++ b/kolibri_explore_plugin/assets/src/views/ContentItem.vue
@@ -156,3 +156,16 @@
   };
 
 </script>
+
+
+<style lang="scss" scoped>
+
+  @import '../styles';
+
+  // Fix icon offset in the Kolibri plugins:
+  .content-renderer::v-deep .button img,
+  .content-renderer::v-deep .button svg {
+    vertical-align: baseline;
+  }
+
+</style>


dylanmccall:
I would prefer if we can solve this closer to its source. I took a look at the offending icon and it looks like the problem is the following blob of CSS in kolibri_explore_plugin.app0.0.23-210630-0926.css:154:

```
img, svg {
    vertical-align: middle;
}
```

Looking at the surrounding context, I believe this comes from Bootstrap as part of its global CSS reset. It just happens that this is a surprising thing to reset it to and normally Kolibri assumes svg elements will be baseline-aligned. This is likely to cause issues with Kolibri's icon buttons anywhere they happen to appear. Instead, we can add the following to our global styles.scss:

```
.button img, .button svg {
    vertical-align: baseline;
}
```

manuq:
> I would prefer if we can solve this closer to its source, since I think the issue here is more broad. I took a look at the offending icon and it looks like the underlying problem is the following blob of CSS in kolibri_explore_plugin.app0.0.23-210630-0926.css:154:
> 
> ```
> img, svg {
>     vertical-align: middle;
> }
> ```
> 
> Looking at the surrounding context, I believe that comes from Bootstrap as part of its global CSS reset. It just happens that this is a surprising thing to reset it to and normally Kolibri assumes svg elements will be baseline-aligned. This is likely to cause issues with Kolibri's icon buttons anywhere they happen to appear. Instead, we can add the following to our global styles.scss:
> 
> ```
> .button img, .button svg {
>     vertical-align: baseline;
> }
> ```
> 
> Needing to use the generic `button` class name (as used by Kolibri's `KButton` component) is irritating. But fortunately - by some miracle! - Bootstrap's buttons are `btn` so there shouldn't be a collision.

Great finding! Still I think the reset is important for Bootstrap to work. We can re-reset the SVGs to baseliine only for the KContentRenderer.

manuq:
> I would prefer if we can solve this closer to its source, since I think the issue here is more broad. I took a look at the offending icon and it looks like the underlying problem is the following blob of CSS in kolibri_explore_plugin.app0.0.23-210630-0926.css:154:
> 
> ```
> img, svg {
>     vertical-align: middle;
> }
> ```
> 
> Looking at the surrounding context, I believe that comes from Bootstrap as part of its global CSS reset. It just happens that this is a surprising thing to reset it to and normally Kolibri assumes svg elements will be baseline-aligned. This is likely to cause issues with Kolibri's icon buttons anywhere they happen to appear. Instead, we can add the following to our global styles.scss:
> 
> ```
> .button img, .button svg {
>     vertical-align: baseline;
> }
> ```
> 
> Needing to use the generic `button` class name (as used by Kolibri's `KButton` component) is irritating. But fortunately - by some miracle! - Bootstrap's buttons are `btn` so there shouldn't be a collision.

I went to the source and the vertical-align middle is a workaround for IE, which as far as I know we are still trying to support:

```
svg {
  // Workaround for the SVG overflow bug in IE10/11 is still required.
  // See https://github.com/twbs/bootstrap/issues/26878
  overflow: hidden;
  vertical-align: middle;
}
```

PR #154, opened 2021-07-01
overrides: Split Healthy Living into Body and Mind

manuq:
Rename all the healthy-living to healthy-body and create a new
override for healthy-mind. Also move the assets accordingly and update
the customApps metadata.

https://phabricator.endlessm.com/T32192


danigm (approved):
LGTM. With this we've right now the same big thumbnail for both channels, but I think it's okay as a placeholder until we've the new asset.

manuq:
> LGTM. With this we've right now the same big thumbnail for both channels, but I think it's okay as a placeholder until we've the new asset.

Yes the thumbnails are going to be replaced soon. We are figuring out the right format first.

PR #155, opened 2021-07-01
template: Add grid for simple bundle page

manuq:
Add a nested grid to display 2 content images per row in medium size
screens and above. In small screens it all is piled vertically.

Make the content images "fluid" so they adjust to the container space.

Use the gutter width variable (defaults to 30px) for the bottom
margin, to have the same gutter between rows than between columns.

https://phabricator.endlessm.com/T32075


danigm (approved):
LGTM

PR #157, opened 2021-07-02
Components: Remove placeholder-loading dependency

danigm:
This patch removes the placeholder-loading dependency and uses the
bootstrap-vue skeleton component to do that.

https://phabricator.endlessm.com/T32183


manuq (approved):
Excellent! I also didn't know about the placeholder components in bootstrap-vue.

PR #162, opened 2021-07-05
Remove close button shadow and white border for videos

dylanmccall:
The existing border override needs `!important` to apply in all cases.

https://phabricator.endlessm.com/T32185

### DIFF
--- a/kolibri_explore_plugin/assets/src/views/ContentModal.vue
+++ b/kolibri_explore_plugin/assets/src/views/ContentModal.vue
@@ -99,7 +99,7 @@
 
   // This is overriding the Kolibri media player plugin:
   ::v-deep .content-renderer {
-    border: 0;
+    border: 0 !important;
   }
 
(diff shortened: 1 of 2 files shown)


kolibri_explore_plugin/assets/src/views/ContentModal.vue
```diff
@@ -99,7 +99,7 @@
 
   // This is overriding the Kolibri media player plugin:
   ::v-deep .content-renderer {
-    border: 0;
+    border: none !important;
```
manuq, line 102: Oh OK! I couldn't find a way to reproduce this. In any case, please use `0 !important!` or the linter will complain.

dylanmccall: Ah, that explains why it was `border: 0`, then :b Fixed!

manuq (approved):
Excellent!

PR #167, opened 2021-07-06
template: Improve backgrounds again

manuq:
Use the LG breakpoint (down md) and assume that background assets will
have 2.8444 aspect ratio. Same AR as the header at exactly the LG
breakpoint. So the header height is adjusted as well.

- Below LG: the background is displayed full height.
- LG and up: the background is displayed full width.

Also, remove the unused xxl breakpoint.


danigm (approved):
LGTM, it works correctly on all sizes and we've control over the background images, so we can fix the aspect ratio.

PR #168, opened 2021-07-06
template: Improve backgrounds again

manuq:
Use the XXL breakpoint (down xl) and assume that background assets
will have 2.82 aspect ratio. Same AR as the header at exactly the LG
breakpoint. So the header height is adjusted as well.

- Below XXL: the background is displayed full height.
- XXL and up: the background is displayed full width.


manuq:
Note: this is an alternative for #167

PR #169, opened 2021-07-06
ContentModal: Add special style rules for zim content

dylanmccall:
https://phabricator.endlessm.com/T32130

Screenshot 2021-07-07 at 18-46-56 Canal de Patatas


dylanmccall:
This is still a draft pending some changes to le-utils, making Zim content have its own node kind, but other than that detail (marked with a TODO), I think it matches the design.

manuq:
It's looking great! Nothing to point or suggest from my side.

danigm (approved):
Looks good to me, maybe we can start to think about moving these fixed styles to variables in the theme or something like that, but for now, for something specific like this, I think it's okay.

PR #171, opened 2021-07-06
components: Fix empty tags in card body

manuq:
By filtering the null values.

https://phabricator.endlessm.com/T32078


danigm (approved):
LGTM

PR #173, opened 2021-07-07
styles: Move global CSS rules to a separate index.scss

dylanmccall:
This change fixes a problem where global CSS rules were being duplicated
in the context of every Vue component. Instead, styles.scss should only
contain variables and mixins.

https://phabricator.endlessm.com/T32234


dylanmccall:
Some notes for review:

* The file kolibri_explore_plugin.app*.css should be _much_ smaller. (For me it went from ~12.5 MB to ~800 kB).
* This should cause absolutely no visual changes. If there are any, that's an error :)
* Renaming / deleting files is endlessly horrible, so let's be careful that I haven't deleted any CSS rules in the process of resolving merge conflicts.

manuq:
Oh wow great finding @dylanmccall ! I will take a detailed look tomorrow, it's end of the work day here. We need to make sure that this plays well with the SCSS overrides per channel. Thanks!

danigm (approved):
I've not detected any regression. This is a great performance improvement, and now that we'll use a modern Chrome with the electron app we can just remove the specific [fixes for IE11](https://github.com/endlessm/kolibri-explore-plugin/blob/fcce488171939210dcfe45f5beac7fd8158be3cf/packages/template-ui/vue.config.js#L6)

packages/eos-components/src/styles.scss
```diff
@@ -103,74 +105,72 @@ $carousel-indicator-transition:      all .6s ease !default;
...
+  background-color: $header-color;
+  background-attachment: fixed;
+  background-position: top center;
+  background-repeat: no-repeat;
 
-// Change header color:
-$header-color: scale-color($secondary, $lightness: 50%);
+  @include media-breakpoint-down(xxl) {
+    background-size: auto $header-height;
+  }
```
manuq: This rule was in the removed mixins.scss but there is no xxl breakpoint anymore. You probably removed the file and then rebased. https://github.com/endlessm/kolibri-explore-plugin/pull/173/files#diff-24cf3bf40e034d42265157f36698d67dea616cb72bd8aaf39433869c9c33b367L48-L50

packages/eos-components/src/styles.scss
```diff
@@ -103,74 +105,72 @@ $carousel-indicator-transition:      all .6s ease !default;
...
-.breadcrumb {
-  font-family: $headings-font-family;
-  font-weight: $headings-font-weight;
-}
-.badge {
-  font-family: $headings-font-family;
-  color: $text-muted;
+  @include media-breakpoint-up(xxl) {
+    background-size: 100% auto;
+  }
```
manuq: Same here, no xxl breakpoint anymore.

manuq:
I have appended a fixup commit for the 2 regressions: background size and breadcrumb colors. This is a performance boost in livecoding development as well! Great job @dylanmccall !

manuq (approved):

PR #174, opened 2021-07-07
T32077 cards

manuq:

### DIFF
--- a/packages/eos-components/src/components/mixins/cardMixin.js
+++ b/packages/eos-components/src/components/mixins/cardMixin.js
@@ -23,7 +23,7 @@ export default {
   },
   methods: {
     getThumbnail() {
-      if (!this.node.thumbnail && process.env.VUE_APP_USE_MOCK_DATA === 'true') {
+      if (!this.node.thumbnail || this.node.useDefaultThumbnail) {
         this.thumbnail = this.getAsset('defaultThumbnail');
         return;
       }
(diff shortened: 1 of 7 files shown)


packages/eos-components/src/components/mixins/cardMixin.js
```diff
@@ -23,7 +23,7 @@ export default {
   },
   methods: {
     getThumbnail() {
-      if (!this.node.thumbnail && process.env.VUE_APP_USE_MOCK_DATA === 'true') {
+      if (!this.node.thumbnail || this.node.useDefaultThumbnail) {
```
danigm, line 26: :+1:  Good idea to clean this kind of code that we're not using anymore.

packages/eos-components/src/utils.js
```diff
@@ -68,7 +68,11 @@ export function getNodeUrl(node, channelId) {
 
 export function getLeaves(node) {
   if (!node.children) {
-    return [node];
+    if (node.kind === 'topic') {
+      return [];
+    } else {
```
danigm, line 73: This else is not needed after a return: https://eslint.org/docs/rules/no-else-return

danigm (approved):
Looks good, just one eslint problem to fix.

packages/eos-components/src/utils.js
manuq: OK! I can change this, I prefer that too. But that eslint rule is not complaining in our checks.

PR #175, opened 2021-07-08
ContentModal: Use a dark styled top bar for content

dylanmccall:
https://phabricator.endlessm.com/T32187

Screenshot 2021-07-07 at 18-15-27 How to get started with Kolibri on Endless OS
Screenshot 2021-07-07 at 18-15-48 PhET Interactive Simulations (English)
Screenshot 2021-07-07 at 18-18-13 Canal de Patatas

### DIFF
--- a/kolibri_explore_plugin/assets/src/views/ContentItem.vue
+++ b/kolibri_explore_plugin/assets/src/views/ContentItem.vue
@@ -1,6 +1,6 @@
 <template>
 
-  <div>
+  <div :class="contentItemClass">
     <template v-if="sessionReady">
       <KContentRenderer
         v-if="!content.assessment"
@@ -73,6 +73,10 @@
         type: Object,
         required: true,
       },
+      dark: {
+        type: Boolean,
+        default: false,
+      },
     },
     data() {
       return {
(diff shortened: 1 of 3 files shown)


kolibri_explore_plugin/assets/src/views/ContentItem.vue
```diff
@@ -1,6 +1,6 @@
 <template>
 
-  <div>
+  <div :class="contentItemClass">
```
danigm, line 3: The classes can be done a bit simpler using the [vue syntax](https://vuejs.org/v2/guide/class-and-style.html#Object-Syntax), it could be something like:

```suggestion
  <div class="content-item" :class="{ 'content-item--dark': isDark }">
```

kolibri_explore_plugin/assets/src/views/ContentItem.vue
```diff
@@ -73,6 +73,10 @@
         type: Object,
         required: true,
       },
+      dark: {
```
danigm, line 76: We can use a more meaningful property name

```suggestion
      isDark: {
```

kolibri_explore_plugin/assets/src/views/ContentItem.vue
```diff
@@ -90,6 +94,13 @@
         extraFields: state => state.core.logging.summary.extra_fields,
         fullName: state => state.core.session.full_name,
       }),
+      contentItemClass() {
+        const classNames = ['content-item'];
+        if (this.dark) {
+          classNames.push('content-item--dark');
+        }
+        return classNames;
+      },
```
danigm, line 103: This can be removed with the previous suggestion.

```suggestion
```

danigm (changes requested):
Looks good, but I think that we can simplify a bit the content item classes

kolibri_explore_plugin/assets/src/views/ContentItem.vue
dylanmccall: Ooh, I didn't know we could give it a dict! That's handy :)

PR #177, opened 2021-07-08
content renderer: Remove fullscreen bar for T2 games

manuq:
Add a GameAppIDs array constant to store all the channels that need
the fullscreen bar removed. For now it's just the Terminal Two
channel.

Use the array to add a CSS class to the content renderer, and add a
display: hidden rule to it.

https://phabricator.endlessm.com/T32242


danigm (approved):
LGTM

PR #178, opened 2021-07-08
T32077 cards

manuq:


danigm (approved):
LGTM. There's some code very similar in the `RegularCard` and `TopicCard`, that maybe could be moved to a mixin or base component, but I think it's okay in this case, to avoid more complexity in the components.

manuq:
> LGTM. There's some code very similar in the `RegularCard` and `TopicCard`, that maybe could be moved to a mixin or base component, but I think it's okay in this case, to avoid more complexity in the components.

Yes, I will try to refactor, once I know that this is the last design iteration on the cards. Thanks for reviewing.

PR #181, opened 2021-07-09
T32244 discovery empty state

danigm:
https://phabricator.endlessm.com/T32244

### DIFF
--- a/kolibri_explore_plugin/assets/src/constants.js
+++ b/kolibri_explore_plugin/assets/src/constants.js
@@ -26,3 +26,5 @@ export const pageNameToModuleMap = {
 export const CUSTOM_PRESENTATION_TITLE = 'custom-channel-ui';
 
 export const CarouselItems = 5;
+
+export const searchTerms = ['STEM', 'Games', 'Fitness', 'Cooking', 'Arts'];
--- a/kolibri_explore_plugin/assets/src/views/SearchPage.vue
+++ b/kolibri_explore_plugin/assets/src/views/SearchPage.vue
@@ -14,11 +14,11 @@
       <SearchBar
         v-model="query"
         :debounce="800"
-        :loading="loading"
+        :loading="isLoading"
         @clear-input="clearInput"
       />
       <b-container class="pb-5 pt-3">
-        <div v-if="!hasResults && !loading" class="align-items-center d-flex">
+        <div v-if="!hasResults && !isLoading" class="align-items-center d-flex">
           <h5 class="mb-0 text-muted">
             Topic Ideas
           </h5>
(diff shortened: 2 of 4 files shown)


kolibri_explore_plugin/assets/src/constants.js
```diff
@@ -26,3 +26,5 @@ export const pageNameToModuleMap = {
 export const CUSTOM_PRESENTATION_TITLE = 'custom-channel-ui';
 
 export const CarouselItems = 5;
+
+export const searchTerms = ['STEM', 'Games', 'Fitness', 'Cooking', 'Arts'];
```
manuq, line 30: Yes, thanks! This should be a constant from day 1.

kolibri_explore_plugin/assets/src/views/SearchPage.vue
```diff
@@ -101,25 +101,13 @@
...
-        const channels = [];
-        const allChannels = [...this.channels];
-        while (channels.length < this.columns && allChannels.length > 0) {
-          const index = _.random(0, allChannels.length - 1);
-          const [channel] = allChannels.splice(index, 1);
-          channels.push(channel);
-        }
-        return [channels];
+      isLoading() {
+        return this.loading && !!this.query.trim();
```
manuq, line 110: :+1:

kolibri_explore_plugin/assets/src/views/SearchPage.vue
```diff
@@ -223,7 +217,7 @@
   @import '../styles';
 
   .main {
-    background-color: white;
+    background-color: $gray-300;
```
manuq, line 220: :+1:

kolibri_explore_plugin/assets/src/views/SearchPage.vue
```diff
@@ -59,18 +67,7 @@
...
-        <h4 class="text-muted">
-          Featured Channels
-        </h4>
-        <ChannelCardGroup
-          :rows="recommended"
-          :columns="columns"
-          @card-click="goToChannel"
-        />
-      </b-container>
-    </div>
```
manuq, line 73: I see recommended channels in the specs, even for the minimal version:
Captura de pantalla de 2021-07-12 10-55-21

So I would keep this.

manuq (changes requested):
The recommended channels should be kept according to specs. And there should be "some" space between buttons and the row of channels. This is a bit hidden in this comment: https://phabricator.endlessm.com/T32161#898155

The rest looks good.

kolibri_explore_plugin/assets/src/views/SearchPage.vue
danigm: This is for the "No results", that's a bit different from the "Empty state". In any case. I've recovered the recommended in the code and just show it when the result is empty, but not in the first view with an empty search query.

manuq (approved):

PR #183, opened 2021-07-12
Components: Fix thumbnail async load

danigm:
This patch fixes a regression introduced in the following commit:
https://github.com/endlessm/kolibri-explore-plugin/commit/beb2a48d375d0a5fc902e4068b8e35c82d655e4c

It shows the default thumbnail always in custom channels. This patch
fixes that in brings back the content thumbnail async load.


manuq:
Ouch! Sorry for this BIG regression and thanks for the quick fix!

PR #184, opened 2021-07-12
Discovery: Add loading state to search

danigm:
This patch adds the placeholder cards instead of the loading spinner in
the discovery search page.

It also includes some css changes to the searchbar:

 * Clean button restyled, bigger and with a circle
 * Larger text for the input

https://phabricator.endlessm.com/T32246


manuq (approved):
Excellent!

PR #187, opened 2021-07-13
Discovery: No results state in search

danigm:
* The buttons bar will be shown in the "empty" state and in the "no
   results" state, but in the second case without the label.
 * The no results message and suggested channels are only shown when a
   search is done and there's no results.

https://phabricator.endlessm.com/T32247


manuq (approved):
Excellent.

PR #188, opened 2021-07-13
components: Improve channel icon styles in cards

danigm:
Use h6 for the channel title so it has the same font and sizes than
other titles in cards.

https://phabricator.endlessm.com/T32067

### DIFF
--- a/packages/eos-components/src/components/CardBody.vue
+++ b/packages/eos-components/src/components/CardBody.vue
@@ -8,9 +8,13 @@
         {{ node.title }}
       </VClamp>
     </h5>
-    <div v-if="showChannelIcon" class="align-items-center d-flex mb-1">
-      <ChannelLogo class="mr-2" :channel="node.channel" size="sm" />
-      <span class="text-muted text-truncate">{{ node.channel.title }}</span>
+    <div v-if="showChannelIcon">
+      <h6 class="align-items-center d-flex mb-1 text-muted">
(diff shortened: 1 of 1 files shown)


packages/eos-components/src/components/CardBody.vue
```diff
@@ -8,9 +8,13 @@
         {{ node.title }}
       </VClamp>
     </h5>
-    <div v-if="showChannelIcon" class="align-items-center d-flex mb-1">
-      <ChannelLogo class="mr-2" :channel="node.channel" size="sm" />
-      <span class="text-muted text-truncate">{{ node.channel.title }}</span>
+    <div v-if="showChannelIcon">
+      <h6 class="align-items-center d-flex mb-1 text-muted">
```
manuq, line 12: Should be mb-3 (16px) as Joana said.

The h6 doesn't have the text size requested. The easiest is to add the "btn" class. But that also adds unwanted padding. So it has to be done adding a custom class and then:

```
.channel-title {
  font-family: $btn-font-family;
  font-size: $btn-font-size;
  font-weight: $btn-font-weight;
}
```

manuq (changes requested):

manuq:
I implemented this using another approach in https://github.com/endlessm/kolibri-explore-plugin/pull/188#pullrequestreview-705664377

danigm:
This change is superseded by #192

PR #189, opened 2021-07-13
template: Fix missing class for hasDarkHeader

manuq:
The title for small screens is a different element.

https://phabricator.endlessm.com/T32264


danigm (approved):
LGTM

PR #191, opened 2021-07-13
T32248 search results by type

danigm:
https://phabricator.endlessm.com/T32248

### DIFF
--- a/packages/eos-components/src/components/GridPage.vue
+++ b/packages/eos-components/src/components/GridPage.vue
@@ -1,5 +1,5 @@
 <template>
-  <b-row>
+  <transition-group name="fade" mode="in-out" tag="div" class="row">
     <b-col
       v-for="node in nodes"
       :key="node.id"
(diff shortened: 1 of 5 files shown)


packages/eos-components/src/components/GridPage.vue
```diff
@@ -28,4 +28,11 @@ export default {
 
   @import '../styles.scss';
 
+  .fade-enter-active, .fade-leave-active {
+    transition: all 0.3s;
```
manuq, line 32: If we start adding transitions I think it's better to use the theme. Bootstrap has:
```
  .fade-enter-active, .fade-leave-active {
    @include transition($transition-fade);
  }
```
Then Joana can decide if it looks good or not, and we just tweak the $transition-fade variable in the theme. What do you think?

manuq (approved):
Looks great! Nice work on the collapsible grid. I only left a suggestion about the transitions. Feel free to merge after that.

packages/eos-components/src/components/GridPage.vue
danigm: Yes, It's a good idea to move the transition to the theme. I have not played a lot with transitions, it could be better to implement an slide down animation or something like that, but the fade-in-out was the easier thing to implement, and I thought it could be sufficient.

PR #196, opened 2021-07-14
template: Add tooltips to breadcrumbs

manuq:
Because these texts can ellipsize.

https://phabricator.endlessm.com/T32233


danigm (approved):
Looks good!

PR #197, opened 2021-07-14
components: Fix margins of search buttons

manuq:
Also, refactor the 2 buttons in a single component.

https://phabricator.endlessm.com/T32266


dylanmccall:
Yay, more components! :star:

PR #198, opened 2021-07-14
T32278 flat

manuq:
https://phabricator.endlessm.com/T32278

### DIFF
--- a/packages/template-ui/src/views/Home.vue
+++ b/packages/template-ui/src/views/Home.vue
@@ -30,6 +30,7 @@
         <div v-else-if="contentNodes.length">
           <CardGrid
             :nodes="contentNodes"
+            :variant="hasFlatGrid ? 'collapsible' : 'slidable'"
             :mediaQuality="mediaQuality"
             :cardColumns="cardColumns"
           />
@@ -78,6 +79,7 @@ export default {
       'hasSectionsSearch',
       'hasCarousel',
       'hasFilters',
+      'hasFlatGrid',
       'displayHeroContent',
     ]),
     ...mapGetters({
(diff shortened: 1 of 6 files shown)


packages/template-ui/src/views/Home.vue
```diff
@@ -17,7 +17,16 @@
...
-      <div v-if="isFilterEmpty">
+
+      <div v-if="hasFlatGrid">
+        <CardGrid
+          variant="collapsible"
+          :nodes="contentNodes"
+          :mediaQuality="mediaQuality"
+          :cardColumns="cardColumns"
+        />
+      </div>
```
dylanmccall, line 28: Nitpick: could use an extra newline to line up with the `<div v-else>` later on :)

```suggestion

      </div>
```

packages/template-ui/src/views/Home.vue
```diff
@@ -17,7 +17,16 @@
       </b-container>
       <FilterContent v-if="hasFilters" />
 
-      <div v-if="isFilterEmpty">
+
+      <div v-if="hasFlatGrid">
```
dylanmccall, line 21: It looks like this is going to misbehave for a channel with filters. If some filters have been toggled, we'll still show the `<CardGrid>` of `contentNodes` instead of `<FilterResult>`. This might flow better if we put `<FilterResult>` in the first condition, as `v-if="isFilterSet` and then other states follow.

dylanmccall (changes requested):

packages/template-ui/src/views/Home.vue
manuq: Yeah I thought about that. This "flat" option is supposed to also enforce no filters according to the spec https://phabricator.endlessm.com/T32104 . But I thought it was better to avoid option dependencies as that's when things start to get complicated. The question is, should we consider all the option combinations if we are not going to use them? In this case I think you are right, the new option is supposed to only affect the tree of nodes and the home grid.

PR #201, opened 2021-07-16
template: Replace paginated grid with collapsible variant

manuq:
In all remaining cases.

https://phabricator.endlessm.com/T32151


erikos (approved):
Looks good!

manuq:
> Looks good!

I now updated it to fix the initial cards regression.

erikos:
Took another look, the logic seems fine on the show-less/show-more and the number of grid items.

erikos:
> > Looks good!
> 
> I now updated it to fix the initial cards regression.

Yes, already checked :)

PR #203, opened 2021-07-16
overrides: Terminal Two: Apply flat option

manuq:
https://phabricator.endlessm.com/T32289


erikos:
Thanks Manuel, looking good. Great to see the flexibility of the Template system in action!

PR #206, opened 2021-07-16
template: Ellipsize the header description

manuq:
And add a button to enlarge the header and show all the description
when it ellipsizes. Per design, we cut at 5 lines.

Also: when a channel has an image asset in the header, use the primary
color instead of the secondary color as fallback. This makes the
background look better in case Endless curated channels do ellipsize
their description (very unlikely). Add a hasHeaderImage computed
property to the mixin for this.

https://phabricator.endlessm.com/T32294

### DIFF
--- a/packages/template-ui/src/components/ChannelHeader.vue
+++ b/packages/template-ui/src/components/ChannelHeader.vue
@@ -1,8 +1,9 @@
 <template>
   <b-jumbotron
     fluid
-    :style="{ backgroundImage: headerImageURL }"
+    :style="{ backgroundImage: (isDescriptionExpanded ? 'none' : headerImageURL) }"
     class="mb-0"
+    :class="{ 'full-height': isDescriptionExpanded, 'has-image': hasHeaderImage }"
   >
     <template v-slot:default>
       <div class="align-items-start d-flex justify-content-between mt-3">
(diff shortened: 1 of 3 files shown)


packages/template-ui/src/components/ChannelHeader.vue
```diff
@@ -60,6 +91,12 @@ export default {
   padding-top: $navbar-height;
   height: $header-height;
```
dylanmccall, line 92: We should swap this `height` for `min-height` now that we're allowing the jumbotron to overflow. There's no guarantee that 5 lines of text will actually fit in $header-height, particularly as the title could wrap over two lines as well. For instance, the collapsed text will still overflow if the browser has a larger default font size.

dylanmccall (changes requested):

packages/template-ui/src/components/ChannelHeader.vue
```diff
@@ -60,6 +91,12 @@ export default {
...
   padding-bottom: $navbar-height;
+  &.full-height {
+    height: auto;
```
dylanmccall, line 95: Per Joana's spec (see https://phabricator.endlessm.com/T32103#899834), I think we may want to _remove_ the background image when expanded, so it's just the solid colour. I'll ask her on Slack just to be sure.

manuq: Yeah, anyways that's unlikely to happen. The easiest is to remove the header background image when expanded. If we also want to remove the navigation bar image we'll have to emit signals or add state, just for this corner case. I talked with Joana and she is fine removing the header background image only, which shows the solid color. I'm also not spending time on transitions or anything on purpose. Ideally it would be:
- Show more: transition to primary color (bg fade out), then expand.
- Show less: unexpand, then transition back (bg fade in).

This made me realize that using the `clamped` boolean was wrong. What we need is to check if the description is expanded or not in order to change the CSS. There is an `expanded` property in the slot. There is no event for `expanded` like `@clampchange=` in vue-clamp but fortunately the `expanded` property has the [sync modifier](https://vuejs.org/v2/guide/components-custom-events.html#sync-Modifier).

manuq:
Ready for review again!

danigm (approved):
Looks good  to me, we can merge.

PR #207, opened 2021-07-19
components: Move logo inside the margins for header

danigm:
This patch removes the negative margin for the Endless logo in the
header, so now the logo will be inside the container, like any other
component.

The negative margin is something that doesn't work correctly with
different screen sizes because the margin is not fixed and can be
smaller than the logo size.

With this change we loose the effect of the logo at the left of the
content with the scroll, but if makes it work correctly with diffrent
sizes.

This patch also hides the endless logo for "xs" sizes, showing it only
if there's no breadcrumb.

https://phabricator.endlessm.com/T32279

### DIFF
--- a/packages/eos-components/src/components/Header.vue
+++ b/packages/eos-components/src/components/Header.vue
@@ -1,7 +1,7 @@
 <template>
   <b-navbar class="header" :class="{ shadow: hasScrolled }" fixed="top">
     <b-container class="px-3">
-      <img class="logo" :src="logo" @click="$emit('click-logo')">
+      <img v-if="showLogo" class="logo" :src="logo" @click="$emit('click-logo')">
       <slot></slot>
 
       <b-navbar-nav class="ml-auto">
@@ -17,6 +17,12 @@
 
   export default {
     name: 'Header',
+    props: {
+      showLogo: {
+        type: Boolean,
+        default: true,
+      },
+    },
     data() {
       return {
         hasScrolled: false,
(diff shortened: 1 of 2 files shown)


packages/eos-components/src/components/Header.vue
```diff
@@ -1,7 +1,7 @@
 <template>
   <b-navbar class="header" :class="{ shadow: hasScrolled }" fixed="top">
     <b-container class="px-3">
-      <img class="logo" :src="logo" @click="$emit('click-logo')">
+      <img v-if="showLogo" class="logo" :src="logo" @click="$emit('click-logo')">
```
manuq, line 4: How could users navigate back to the landing page if they can't click on the logo? It's the only way to go back.

packages/template-ui/src/components/ChannelNavBar.vue
```diff
@@ -28,6 +30,9 @@ export default {
     notAtHome() {
       return this.$route.name !== 'Home';
     },
+    showLogo() {
+      return !(this.xs && this.node.ancestors.length);
```
manuq, line 34: Ah here is the answer to the question above. In XS it doesn't show the logo except for the Home.

manuq (commented):
This works, and also it departs from the design spec as explained in the commit message "we loose the effect of the logo at the left of the content". So I would like to propose an alternative solution to keep the design.

manuq (approved):

PR #212, opened 2021-07-20
ContentModal: Set dark color to content text

danigm:
https://phabricator.endlessm.com/T32300

### DIFF
--- a/kolibri_explore_plugin/assets/src/views/ContentModal.vue
+++ b/kolibri_explore_plugin/assets/src/views/ContentModal.vue
@@ -106,8 +106,18 @@
     border: 0 !important;
   }
 
+  ::v-deep .content-renderer .bg-light {
+    padding-right: $spacer !important;
+    padding-left: $spacer !important;
+    background-color: white !important;
+  }
+
   ::v-deep .modal-footer {
     display: none;
+  }
+
+  ::v-deep .modal-content .paragraph {
+    color: $gray-700 !important;
   }
 
   ::v-deep .modal-body {


kolibri_explore_plugin/assets/src/views/ContentModal.vue
```diff
@@ -116,6 +116,12 @@
     border-bottom-left-radius: $modal-content-border-radius;
   }
 
+  ::v-deep .modal-content .paragraph {
+    // The color used in the perseus plugin:
+    // https://github.com/learningequality/ka-lite/blob/571918ea668013dcf022286ea85eff1c5333fb8b/kalite/distributed/static/js/distributed/perseus/build/perseus-5.css#L60
+    color: #21242c;
```
manuq, line 122: Why use this color instead of $dark from our theme?

manuq (approved):
Looks good but I left a comment about using our theme.

kolibri_explore_plugin/assets/src/views/ContentModal.vue
danigm: I used this color to match the color of the questions, as I put in the comment, this is the color used by the plugin that renders the the exercise. It's possible to use the `$dark` value, but in that case the color of the test description will not match the answers color.

manuq: I see, please communicate with Joana in that case. Maybe she wants the 2 description and answer to use our theme color. See her spec https://phabricator.endlessm.com/T32300

danigm: I've used the `$gray-700` for all paragraphs in the modal, so it's the same for question and answers.

manuq (approved):

PR #214, opened 2021-07-21
T32306 async load

danigm:
The career-girls channel has a lot of content and the backend fails when we try to fetch all nodes with:

```
django.db.utils.OperationalError: Expression tree is too large (maximum depth 1000)
```
For channels like that, instead of querying all the content in one request it's possible to get the same data with one request per topic.

### DIFF
--- a/kolibri_explore_plugin/assets/src/apiResources.js
+++ b/kolibri_explore_plugin/assets/src/apiResources.js
@@ -1,3 +1,5 @@
+import _ from 'underscore';
+
 import { Resource } from 'kolibri.lib.apiResource';
 import Store from 'kolibri.coreVue.vuex.store';
 
@@ -35,6 +37,28 @@ export const ContentNodeResource = new Resource({
   },
   fetchNextSteps(getParams) {
     return this.fetchDetailCollection('next_steps', Store.getters.currentUserId, getParams);
+  },
+  fetchChannelAsync(channelId, parent = null) {
+    // Fetch all content for a channel recursively, making a request to the
+    // backend for each level in the nodes tree
+    const getParams = {
+      channel_id: channelId,
+      parent: parent || channelId,
+    };
+
+    return this.fetchCollection({ getParams }).then(n => {
+      return Promise.all(n).then(nodes => {
+        const promises = nodes
+          .filter(node => node.kind === 'topic')
+          .map(node => this.fetchChannelAsync(channelId, node.id));
+
+        return Promise.all(promises).then(array => {
+          return new Promise(resolve => {
+            resolve(_.union(nodes, _.flatten(array)));
+          });
+        });
+      });
+    });
   },
 });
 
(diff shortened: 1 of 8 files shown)


kolibri_explore_plugin/assets/src/apiResources.js
```diff
@@ -36,6 +38,28 @@ export const ContentNodeResource = new Resource({
...
   },
+  fetchChannelAsync(channelId, parent = null) {
+    // Fetch all content for a channel recursively, making a request to the
+    // backend for each level in the nodes tree
+    const getParams = {
+      channel_id: channelId,
+      parent: parent || channelId,
+    };
+
+    return this.fetchCollection({ getParams }).then(n => {
```
manuq, line 49: There must be a beter way to fetch all topic nodes of a parent node. Directly, without fetching all and then filtering. Maybe something like this that I just found in `kolibri/plugins/coach/assets/src/modules/lessonResources/handlers.js`:

```
    return ContentNodeSearchResource.fetchCollection({
      getParams: {
        search: params.searchTerm,
        ...pickBy({
          kind: query.kind,
          channel_id: query.channel,
        }),
      },
```

fetchDescendants

manuq (approved):
Looking good. I left some suggestions.

kolibri_explore_plugin/assets/src/apiResources.js
danigm: The `fetchDescendants` doesn't return all the node attributes, and if we filter by kind we won't get the content nodes that we're returning.

I played a bit with a custom filter in the `viewsets.py`, but at the end, if we get all the content in one query, there's an exception in the consolidate method, when it tries to generate the `ancestors`. We can even try to fix this on our override of the consolidate, but we'll loose the `ancestors` that's used in some places.

kolibri_explore_plugin/assets/src/apiResources.js
```diff
@@ -36,6 +38,28 @@ export const ContentNodeResource = new Resource({
...
+      return Promise.all(n).then(nodes => {
+        const promises = nodes
+          .filter(node => node.kind === 'topic')
```
manuq, line 52: Nit: use `ContentNodeKinds.TOPIC` not `'topic'`. As other Kolibri plugins do.

kolibri_explore_plugin/assets/src/apiResources.js
manuq: OK, sounds good then. Thanks for having taken an extra look.

manuq (approved):
Looks good to merge!

kolibri_explore_plugin/assets/src/apiResources.js
manuq: And sorry, the `fetchDescendants` was a leftover, it wasn't intended to be part of my comment. I copy/pasted while digging the source code.

PR #217, opened 2021-07-21
template: Fix detail page for flat channels

manuq:
The array of ancestors should be of node objects, not node IDs.

https://phabricator.endlessm.com/T32310


danigm (approved):
LGTM

PR #220, opened 2021-07-22
discovery page: Add synonyms to search terms

manuq:
To get better results from the backend, in particular for the main
buttons.

Also, update the list of search terms.

It's using a Map to ensure order.

https://phabricator.endlessm.com/T32320


danigm (approved):
Looks good to me!

PR #222, opened 2021-07-23
components: Use always 3 max lines for topic description

manuq:
In T32077 we made the false assumption that 4 lines of description
would fit when the title is not clamped, and 3 otherwise. But the
title can have 2 lines and still not clamp.

So we have no option but using 3 max lines always for the description.

https://phabricator.endlessm.com/T32326


manuq:
No review yet. Auto-merging.

PR #223, opened 2021-07-23
T32325 low q thumbs

manuq:
https://phabricator.endlessm.com/T32325

### DIFF
--- a/packages/eos-components/src/styles.scss
+++ b/packages/eos-components/src/styles.scss
@@ -78,10 +78,11 @@ $header-logo-width: 128;
 $header-height: 349px; // Forms an area with the LG breakpoint of aspect ratio: 2.8444
 $hover-lightness: 25%;
 $background-alpha: -8%;
-$card-image-ar: percentage(9 / 16);
+$card-image-ar: 9 / 16;
 
 :export {
   headerLogoWidth: $header-logo-width;
+  cardImageAspectRatio: $card-image-ar;
   xs: map-get($grid-breakpoints, "xs");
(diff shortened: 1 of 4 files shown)


manuq:
Note: This PR is on top of #222

packages/eos-components/src/styles.scss
```diff
@@ -78,10 +78,12 @@ $header-logo-width: 128;
 $header-height: 349px; // Forms an area with the LG breakpoint of aspect ratio: 2.8444
 $hover-lightness: -25%;
 $background-alpha: -8%;
-$card-image-ar: percentage(9 / 16);
+$card-image-ar-number: 9 / 16;
+$card-image-ar: percentage($card-image-ar-number);
 
 :export {
   headerLogoWidth: $header-logo-width;
+  cardImageARNumber: $card-image-ar-number;
```
dylanmccall, line 86: Nitpick: this would be clearer to me if it's something like…

```suggestion
  cardImageAR: $card-image-ar-number;
```

It's unfortunate Sass doesn't have a function that works as the opposite of `percentage` since I was going to suggest that :( But we could add a prefix to $card-image-ar-number to express that it shouldn't be used elsewhere.

dylanmccall (changes requested):
One tiny nitpick, but I think this looks fine. May as well wait for #222 before merging :)

packages/eos-components/src/styles.scss
manuq: Good point. What I will do is have just the number here, and use the `percentage` in the components that is needed.

PR #224, opened 2021-07-23
theme: Use lighter variant in media buttons

manuq:
Swith from darker variant to lighter.

https://phabricator.endlessm.com/T32321


dylanmccall (approved):
Does what it says :)

PR #225, opened 2021-07-23
theme: Tweak card body height

manuq:
To avoid tag badges to be cropped.

https://phabricator.endlessm.com/T32291

### DIFF
--- a/packages/eos-components/src/styles.scss
+++ b/packages/eos-components/src/styles.scss
@@ -95,7 +95,7 @@ $card-image-ar: percentage(9 / 16);
   $subtitle-line-height: $font-size-base * $line-height-base;
   $tags-line-height: $font-size-base * $line-height-base;
 
-  $margins: map-get($spacers, 1) * 4;
+  $margins: map-get($spacers, 1) * 6;
 
   @return ($type-line-height + $title-line-height +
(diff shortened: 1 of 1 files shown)


dylanmccall:
This works, but I don't really like how we're solving it. What's the reason that we're calculating a fixed card height like this in the first place? The card deck component is already making it so cards have a shared height, and the `.card-content` element is already a flex container. If we add a few more flex containers up the stack from there, we can set `flex-grow: 1` on `.card-content` and things should end up being aligned evenly (and in a less fragile way).

manuq:
> This works, but I don't really like how we're solving it. What's the reason that we're calculating a fixed card height like this in the first place? The card deck component is already making it so cards have a shared height, and the `.card-content` element is already a flex container. If we add a few more flex containers up the stack from there, we can set `flex-grow: 1` on `.card-content` and things should end up being aligned evenly (and in a less fragile way).

Yeah I don't like it either. The reason to calculate the height and set a fixed height is because we don't use the card deck component everywhere, we have cards grids without decks. And cards have different text content depending on where are they displayed and the metadata available. Maybe the final solution is to move everything to decks of cards, but that component is not responsive and it would be a bigger change at this point.

PR #226, opened 2021-07-23
components: Header: Fix logo alignment

manuq:
Move the padding from the navbar to the container. Also clean unused
CSS rules.

https://phabricator.endlessm.com/T32279


dylanmccall:
Looks like this works. _Much_ better!

PR #234, opened 2021-07-29
T32370 carousel

manuq:
https://phabricator.endlessm.com/T32370

### DIFF
--- a/kolibri_explore_plugin/viewsets.py
+++ b/kolibri_explore_plugin/viewsets.py
@@ -1,6 +1,8 @@
 from kolibri.core.content.api import ContentNodeSearchViewset
 from kolibri.core.content.api import ContentNodeViewset
 from kolibri.core.content.models import ContentTag
+from rest_framework.decorators import list_route
+from rest_framework.response import Response
 
 
 class CustomContentNodeViewset(ContentNodeViewset):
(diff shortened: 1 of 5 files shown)


erikos (commented):
The commit message is cropped here :)

manuq:
> The commit message is cropped here :)

Good catch! Fixed.

kolibri_explore_plugin/viewsets.py
```diff
@@ -28,6 +30,18 @@ def consolidate(self, items, queryset):
...
+    def random_from_channel(self, request, **kwargs):
+        channel_id = self.request.query_params.get("channel_id", None)
+        if channel_id is None:
+            return Response([])
+        count = int(self.request.query_params.get("count", 1))
+        queryset = self.filter_queryset(self.get_queryset())
+        queryset = queryset.filter(channel_id=channel_id)
+        queryset = queryset.order_by("?")[:count]
+        serialized = list(queryset.values("id"))
+        return Response(serialized)
```
dylanmccall, line 43: What's stopping us from returning the content directly instead of just the list of IDs? The extra round trip is pretty minor in this case but it's always nice when we can cut back on those. It looks like Kolibri has a mechanism where they'll run a `serialize` method against the queryset and it magically does the thing:

https://github.com/learningequality/kolibri/blob/release-v0.15.x/kolibri/core/content/api.py#L1067-L1082

https://github.com/learningequality/kolibri/blob/5c88490c692d2c1a21fa09da810a0e8fb4d6c9b5/kolibri/core/api.py#L244-L249

dylanmccall (approved):
One question / suggestion about network round trips, but all's good :)

kolibri_explore_plugin/viewsets.py
manuq: Yeah that was my initial attempt. But it seems that the consolidate method is changing the queryset. So doing something like:
```
queryset = queryset.order_by("?")[:count]
return Response(self.serialize(queryset))
```

I was getting an error from Django: "Cannot reorder a query once a slice has been taken.". In fact I was blocked by that for a while yesterday.

PR #238, opened 2021-07-30
Add more overrides for the endless curated channels

erikos:
Adding overrides for Financial Literacy, Reading and STEAM.

Adding the corresponding assets as well.

https://phabricator.endlessm.com/T32249


manuq (approved):
Fantastic!

PR #240, opened 2021-07-30
T32306 big channels

manuq:


danigm (approved):
Looks good to me. I've just added a new commit to use also the pagination in the discovery search, because we have the same problem there.

manuq:
> Looks good to me. I've just added a new commit to use also the pagination in the discovery search, because we have the same problem there.

Your commit introduced a lint error. But most importantly, the pagination in the search is being ignored. I tested by reducing the page items to 5 and logging the output. So I'm leaving your commit out and merging.

manuq:
@danigm I have left your commit in a draft PR to avoid losing it #241

PR #241, opened 2021-08-02
T32306 big channels search

manuq:


danigm:
We can close this, I've detected that this is not a real problem right now, there's a max-results parameter that defaults to 30 for search queries, so it's not possible to have this issue in the search.

PR #246, opened 2021-08-04
Components: Style Carousel card to avoid button wrap

danigm:
This patch set the style property nowrap to ensure that we always have
the icon and the text aligned in the PlayButton component.

It also adds some style classes to improve the responsiveness for the
CarouselCard, hidding the subtitle when there's no space.

https://phabricator.endlessm.com/T32389

### DIFF
--- a/packages/eos-components/src/components/PlayButton.vue
+++ b/packages/eos-components/src/components/PlayButton.vue
@@ -3,7 +3,7 @@
     pill
     size="sm"
     :variant="`${kind}-${variant}`"
-    class="border-0 pb-1 pl-1 play-button pt-1"
+    class="border-0 pb-1 pl-1 play-button pt-1 text-nowrap"
     @click.stop="$emit('click')"
   >
     <component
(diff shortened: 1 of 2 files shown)


packages/eos-components/src/components/PlayButton.vue
```diff
@@ -74,6 +74,10 @@ export default {
 <style lang="scss" scoped>
 @import '../styles.scss';
 
+.play-button {
+  white-space: nowrap;
```
manuq, line 78: In Bootstrap there is a `text-nowrap` class that does the same, so you don't have to add additional CSS.

manuq (changes requested):

packages/eos-components/src/components/PlayButton.vue
danigm: That's true! I've updated the PR to use that class instead of adding the css.

manuq (approved):

PR #247, opened 2021-08-04
Add more overrides for the Endless curated channels

erikos:
Some are flat since we only have one category/subject. For Sikana we will
only use the DIY content.

https://phabricator.endlessm.com/T32401


danigm (approved):
Looks good to me. I've not tested with all the channels updated, but I think it's safe to be merged.

danigm:
> Looks good to me. I've not tested with all the channels updated, but I think it's safe to be merged.

The lint is failing, we should check where's the problem before merging.

PR #251, opened 2021-08-04
T32306 loading

manuq:


danigm (approved):
Good work, the loading is a bit more smooth with these small changes.

PR #252, opened 2021-08-05
Components: Add description to CarouselCard

danigm:
Adds the description to the CarouselCard if exists and in that case the
title will have just one line.

If there's no description the behavior is the same as before.

https://phabricator.endlessm.com/T32345

### DIFF
--- a/packages/eos-components/src/components/CarouselCard.vue
+++ b/packages/eos-components/src/components/CarouselCard.vue
@@ -28,7 +28,7 @@
               </VClamp>
             </h4>
           </div>
-          <div class="align-items-center d-flex justify-content-between">
+          <div class="align-items-center d-flex justify-content-between mt-2">
             <div v-if="showChannelIcon" class="align-items-center d-flex">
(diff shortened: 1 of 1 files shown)


manuq (commented):
Code-wise looks good. I commented in the ticket.

packages/eos-components/src/components/CarouselCard.vue
```diff
@@ -138,9 +138,13 @@ export default {
 
 .card-content .head {
   min-height: 3 * ($h2-font-size * $headings-line-height);
+  max-width: calc(100% - 100px);
```
manuq, line 141: I guess we can use a flexbox to do this properly and avoid hardcoding the button size, which will change depending on the text label. Think if in the future we support translations.

manuq (changes requested):
I know Joana said "max width" but I think we should let the components allocate the desired size.

danigm:
Closing this, we don't need the description for now.

PR #254, opened 2021-08-05
Components: Add shadow and hover effect to Carousel

danigm:
https://phabricator.endlessm.com/T32346

### DIFF
--- a/packages/eos-components/src/components/Carousel.vue
+++ b/packages/eos-components/src/components/Carousel.vue
@@ -5,7 +5,7 @@
       fade
       indicators
       :interval="interval"
-      class="bg-secondary rounded-lg shadow"
+      class="bg-secondary rounded-lg with-shadow"
     >
       <b-carousel-slide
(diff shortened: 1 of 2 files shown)


packages/eos-components/src/components/Carousel.vue
```diff
@@ -71,4 +71,16 @@ export default {
   background-color: $gray-900;
 }
 
+.with-shadow {
+  transition: all ease .4s;
+  box-shadow: $box-shadow-sm !important;
+  &:hover {
+    box-shadow: $box-shadow !important;
+    ::v-deep h3, ::v-deep h4 {
```
manuq, line 79: Using `v-deep` for components we own is not ideal. That's more a workaround for tweaking unmanaged components. I guess you can pass a new boolean property `is-hovered` to the CarouselCard component.

manuq (changes requested):

manuq (approved):
Excellent!

PR #257, opened 2021-08-06
T32306 skip parsing

manuq:

### DIFF
--- a/packages/eos-components/src/utils.js
+++ b/packages/eos-components/src/utils.js
@@ -1,3 +1,5 @@
+import _ from 'underscore';
+
 import {
   StructuredTagsRegExp,
   StructuredTags,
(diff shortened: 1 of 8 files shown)


packages/eos-components/src/utils.js
```diff
@@ -7,11 +7,15 @@ import {
 
 /** Structured tags **/
 
-export function getFirstStructuredTag(node, matchKey) {
-  if (!(matchKey in node.structuredTags)) {
-    return null;
+export function getAllStructuredTags(node, matchKey) {
+  if (!('structuredTags' in node) || !(matchKey in node.structuredTags)) {
+    return [];
```
danigm, line 12: all of this could be done with the [`_.get`](https://underscorejs.org/#get) in just one line.

danigm (approved):
Looks good to me. Just one nitpick comment about the use of `_.get` to do the check get of structured tags.

packages/eos-components/src/utils.js
manuq: Good. I thought this was going to error when `node.structuredTags` was undefined, but it turns `_.get()` does consider `undefined`:

```
> _.get({foo: 'bar'}.structuredTags, 'baz', [])
[]
> _.get(undefined, 'baz', [])
[]
```

PR #258, opened 2021-08-09
overrides: Add Khan Academy override to remove filters

manuq:
For performance reasons.


danigm (approved):
Looks good to me.

PR #260, opened 2021-08-09
T32303 icons

manuq:
https://phabricator.endlessm.com/T32303


danigm (changes requested):
This solution only work for the home discovery page, in the search and in custom channel presentation the thumbnail used is the one that comes from the database.

manuq:
> This solution only work for the home discovery page, in the search and in custom channel presentation the thumbnail used is the one that comes from the database.

To me it is working in the search. But you are right, it's not working in the custom channel presentation:
Captura de pantalla de 2021-08-10 09-06-51

danigm:
> > This solution only work for the home discovery page, in the search and in custom channel presentation the thumbnail used is the one that comes from the database.
> 
> To me it is working in the search. But you are right, it's not working in the custom channel presentation:

I'm talking about the list of channels in the result of a search, I've added a new commit to solve that.

danigm (approved):
Looks good to me.

PR #263, opened 2021-08-10
template: Don't overuse the channel card in the footer

manuq:
In this case doing composition is a bit forced because it's not
actually a card, so it doesn't need ellipsizing at max lines. Also,
using composition we had to add too much CSS, it was more complicated.

Also, remove the infoCard variant of ChannelCard.

https://phabricator.endlessm.com/T32388


danigm (approved):
Looks good to me.

PR #270, opened 2021-08-12
Discovery: Wait for complete loading on search

danigm:
With the change to do a API query for each content kind we need to wait
for every request to finish before showing the result.

https://phabricator.endlessm.com/T32432


manuq (changes requested):
Code wise looks good, but I would like to talk about the approach in the ticket before merging.

PR #272, opened 2021-09-02
T32487 standalone channel

manuq:
https://phabricator.endlessm.com/T32487

### DIFF
--- a/kolibri_explore_plugin/assets/src/routes/index.js
+++ b/kolibri_explore_plugin/assets/src/routes/index.js
@@ -42,7 +42,7 @@ export default [
     name: PageNames.TOPICS_TEST,
     path: '/topics/:channel_id/test',
     handler: toRoute => {
-      store.commit('topicsTree/SET_CUSTOM_APP_CONTENT', { test: true });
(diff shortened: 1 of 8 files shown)


kolibri_explore_plugin/assets/src/routes/index.js
```diff
@@ -42,7 +42,7 @@ export default [
     name: PageNames.TOPICS_TEST,
     path: '/topics/:channel_id/test',
     handler: toRoute => {
-      store.commit('topicsTree/SET_CUSTOM_APP_CONTENT', { test: true });
+      store.commit('topicsTree/SET_CUSTOM_APP_PARAMETERS', { test: 'true' });
```
danigm, line 45: I think that this change to string is not needed, the encode function will translate this to string.

danigm (approved):
Looks good to me. Just an small comment about the use of string or boolean for the parameters.

kolibri_explore_plugin/assets/src/routes/index.js
manuq: You are right, the encode will do it.

PR #274, opened 2021-09-13
Get carousel random content from kolibri API

danigm:
This change uses the WIP API change done in this PR: https://github.com/endlessm/kolibri/pull/4

https://phabricator.endlessm.com/T32513

### DIFF
--- a/packages/template-ui/src/views/Home.vue
+++ b/packages/template-ui/src/views/Home.vue
@@ -127,9 +127,16 @@ export default {
           });
       }
       else {
-        // FIXME add API to query random content nodes
         console.debug(`Fetch ${this.carouselSlideNumber} random content nodes`);
-        this.loadingCarouselNodes = false;
+
+        window.kolibri.getContentByFilter({
+          random: true,
+          onlyContent: true,
+          pageSize: 5,
+        }).then((page) => {
+          this.carouselNodes = page.results;
+          this.loadingCarouselNodes = false;
+        });
       }
     },
     fetchContentNodes() {


packages/template-ui/src/views/Home.vue
```diff
@@ -129,7 +130,19 @@ export default {
       else {
         // FIXME add API to query random content nodes
         console.debug(`Fetch ${this.carouselSlideNumber} random content nodes`);
-        this.loadingCarouselNodes = false;
+
+        window.kolibri.getContentByFilter({
+          extra: {
```
manuq, line 135: Why this is called "extra" instead of adding to the current "options" attribute?

packages/template-ui/src/views/Home.vue
```diff
@@ -129,7 +130,19 @@ export default {
...
+            channel_id: this.channel.id,
```
manuq, line 136: All APIs should be constrained to the current channel, to avoid leaking outside the sandbox. So there shouldn't be a need to pass the channel ID.

packages/template-ui/src/views/Home.vue
```diff
@@ -129,7 +130,19 @@ export default {
...
+            random: Math.random().toString(),
```
manuq, line 137: I would prefer to ask "give me N random elements" than to passing a random number from the frontend to the backend. The backend should be a black box for the frontend.

packages/template-ui/src/views/Home.vue
```diff
@@ -129,7 +130,19 @@ export default {
...
+            kind_in: 'video,audio,document,exercise,html5,zim',
```
manuq, line 138: To avoid a explicit list, it is better to use a `only_content: true` rather than a list. If going wth a list, it should be an actual array of constant strings, not a string.

manuq (changes requested):
Here is a review from the frontend side. Also it needs some explanation in the commit message, but I think that's because this is work in progress.

packages/template-ui/src/views/Home.vue
danigm: This is something that's not implemented on the API, right now you only can pass the parent, but the parent could be any ID and if you don't pass the parent, there's no filter. If the main idea is to limit this API to the current channel, we can add the `channel_id` filter to all the API calls.

manuq: Yeah :+1: we should filter all API calls by the current channel.

manuq (approved):

PR #275, opened 2021-09-13
T32534 breadcrumbs

manuq:
https://phabricator.endlessm.com/T32534

### DIFF
--- a/packages/template-ui/src/components/ChannelHeader.vue
+++ b/packages/template-ui/src/components/ChannelHeader.vue
@@ -59,7 +59,7 @@
 </template>
 
 <script>
-import { mapState, mapGetters } from 'vuex';
+import { mapState } from 'vuex';
 import VClamp from 'vue-clamp';
 
 import headerMixin from '@/components/mixins/headerMixin';
@@ -70,6 +70,9 @@ export default {
     VClamp,
   },
   mixins: [headerMixin],
+  props: {
+    section: Object,
+  },
   data() {
     return {
       maxDescriptionLines: 6,  // Actually 5. One more for the show more/less button.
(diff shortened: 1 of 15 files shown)


packages/template-ui/src/views/Home.vue
```diff
@@ -2,7 +2,7 @@
   <div
     :style="{ backgroundImage: backgroundImageURL }"
   >
-    <ChannelNavBar />
+    <ChannelNavBar :atHome="true" />
```
danigm, line 5: Maybe the `atHome` could be replaced with a check for the node prop. It's not the same? I mean, in this case, `atHome` is used when there's no node to pass to the component, so maybe we can replace that property with a computed one.

packages/template-ui/src/components/ChannelHeader.vue
```diff
@@ -78,7 +81,18 @@ export default {
   },
   computed: {
     ...mapState(['displayLogoInHeader', 'hasDarkHeader']),
-    ...mapGetters(['headerTitle', 'headerDescription']),
+    sectionOrChannel() {
+      if (this.section === undefined || this.section.id === this.channel.id) {
```
danigm, line 85: Is this second expression needed? If the section and channel id are the same it should  be the same to return `this.channel` or `this.section`. In that case we can replace this whole method to `return this.section || this.channel;`

danigm (changes requested):
Looks good! I've just left a couple of questions

packages/template-ui/src/views/Home.vue
manuq: After going back to the code I realize why the `atHome` option is needed: the Search view doesn't have a node to pass to the header component, but the header should have the link to go back home. Only the Home view doesn't need the link. So I'm leaving this as-is.

packages/template-ui/src/components/ChannelHeader.vue
manuq: Yes, the second expression is needed. What this does is return the channel information in case the section (topic) is the root one, in which case the ID will match with the channel.

If we remove the second part, navigating to `/t/ROOT_ID` would have a different header information than navigating to `/`.

All I'm doing is maintaining the old behavior, so I prefer to keep it as-is.

PR #278, opened 2021-09-20
Bring back filters with new API call

danigm:
This PR is based on: https://github.com/endlessm/kolibri-explore-plugin/pull/286

This patch bring back filters (currently just the media type one) using
the kolibri.getContentByFilter new API.

https://phabricator.endlessm.com/T32509

This patch uses the new API option implemented here: https://github.com/endlessm/kolibri/pull/6


packages/template-ui/src/App.vue
```diff
@@ -30,17 +30,22 @@ export default {
...
 
+    const filterOptionsPromise = window.kolibri.getChannelFilterOptions()
+      .then((filterOptions) => {
+        this.$store.commit('filters/setFilterOptions', filterOptions);
+      });
+
     // Flat presentations don't need the main sections:
     if (this.hasFlatGrid) {
-      return channelPromise;
+      return Promise.all([channelPromise, filterOptionsPromise]);
```
manuq, line 40: Only fetch the filter options if the state `hasFilters` is true.

packages/template-ui/src/App.vue
```diff
@@ -30,17 +30,22 @@ export default {
...
     }
 
     const sectionsPromise = window.kolibri.getContentByFilter({ parent: 'self', onlyTopics: true })
       .then((page) => {
         this.$store.commit('setMainSections', { mainSections: page.results });
         this.handleRedirects();
       });
-    return Promise.all([channelPromise, sectionsPromise]);
+    return Promise.all([channelPromise, filterOptionsPromise, sectionsPromise]);
```
manuq, line 48: Same here, only if `hasFilters` is true.

packages/template-ui/src/store/index.js
```diff
@@ -47,8 +47,7 @@ const initialState = {
   hasDarkHeader: false,
   hasSectionsSearch: true,
   hasCarousel: true,
-  // FIXME reenable the filters and migrate them to the API:
-  hasFilters: false,
+  hasFilters: true,
```
manuq, line 50: :+1:

manuq (changes requested):
When testing this with the Dance channel the filter component is not shown. I'm getting: `TypeError: Cannot use 'in' operator to search for 'audio' in undefined`.

For this channel I expect to have a filter for the multiple authors.

Also, from the code I don't understand if you are trying to make the structured tags work, or if you are only moving the old code around. I suggest that we drop that functionality and care only about the current metadata at this stage.

Captura de pantalla de 2021-09-27 18-56-59

manuq (changes requested):
Looking better! Still with the Dance channel I'm not getting the filters and I see this in the console: `TypeError: Cannot convert undefined or null to object`

Captura de pantalla de 2021-09-28 07-46-37

I see that this is because you are missing this in the `CustomContentRenderer.vue` file:

```
      this.hashi.on(events.CHANNELFILTEROPTIONSREQUESTED, message => {
        this.sendChannelFilterOptions.call(this, message);
      });
```

danigm:
> I see that this is because you are missing this in the CustomContentRenderer.vue file:

>     this.hashi.on(events.CHANNELFILTEROPTIONSREQUESTED, message => {
>       this.sendChannelFilterOptions.call(this, message);
>     });

Good catch! I had that change on my local clone, but I didn't added to the commit correctly. Now I've fixed the other branch and it should work.

PR #280, opened 2021-09-22
Channel overrides: adjust representation of the SF49ers channel

erikos:
- make it a flat grid
- no carousel or filters needed as we only have 8 vidoes and the
full playbook

Change also the order of the tiles so that SF49ers and
Techbridge girls are in row one and two.

https://phabricator.endlessm.com/T32499

### DIFF
--- a/kolibri_explore_plugin/assets/src/customApps.js
+++ b/kolibri_explore_plugin/assets/src/customApps.js
@@ -29,6 +29,7 @@ export const CustomChannelApps = {
   '3fcffebc58d15175b948b140434ef6e6': 'sports',
   '2f95235c3709511fa12d007f31ed6a7b': 'steam',
   bf0260ed911f44cda27a263db93a8512: '49ers-edu-digital-playbook',
+  '1e755450dede5c40af1487e07984c260': 'techbridge-girls-home',
 };
 
 export function getAppNameByID(id) {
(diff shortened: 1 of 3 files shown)


kolibri_explore_plugin/assets/src/customApps.js
```diff
@@ -28,7 +28,8 @@ export const CustomChannelApps = {
   '59bb2e5a3d2e5e3b85b87d9ab4daa2f3': 'reading',
   '3fcffebc58d15175b948b140434ef6e6': 'sports',
   '2f95235c3709511fa12d007f31ed6a7b': 'steam',
-  e5b26962b0ee5d8487152d0987d32e44: '49ers-edu-digital-playbook',
+  bf0260ed911f44cda27a263db93a8512: '49ers-edu-digital-playbook',
+  '1e755450dede5c40af1487e07984c260': 'techbridge-girls',
```
manuq, line 32: For consistency, can you name the custom app `techbridge-girls-home` just like it's already used in other constants below? This means you will have to rename the new subfolder inside the `channel-overrides/` folder.

manuq (changes requested):
Looks good. Only a rename is needed to be merged.

kolibri_explore_plugin/assets/src/customApps.js
erikos: Sure, done. Thanks for catching

danigm (approved):
LGTM

manuq (approved):

PR #284, opened 2021-09-23
T32533 clean thumbs

manuq:
https://phabricator.endlessm.com/T32533


danigm (approved):
Looks good to me. I've just fixed one lint error with a if-else return, so we can merge.

manuq:
> Looks good to me. I've just fixed one lint error with a if-else return, so we can merge.

Ah I didn't notice that lint error. Thanks.

PR #287, opened 2021-09-24
T32512 flat

manuq:
https://phabricator.endlessm.com/T32512

### DIFF
--- a/packages/template-ui/src/views/Home.vue
+++ b/packages/template-ui/src/views/Home.vue
@@ -139,13 +139,18 @@ export default {
     },
     fetchContentNodes() {
       this.loadingContentNodes = true;
-      return window.kolibri.getContentByFilter({ parent: 'self', onlyContent: true })
+      const options = this.hasFlatGrid ? { onlyContent: true } : { parent: 'self', onlyContent: true }
+      return window.kolibri.getContentByFilter(options)
         .then((page) => {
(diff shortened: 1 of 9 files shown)


packages/template-ui/src/views/Home.vue
```diff
@@ -139,13 +139,18 @@ export default {
...
+      return window.kolibri.getContentByFilter(options)
         .then((page) => {
           this.contentNodes = page.results;
           this.loadingContentNodes = false;
         });
     },
     fetchSectionNodes() {
+      if (this.hasFlatGrid) {
+        this.loadingSectionNodes = false;
+        return;
```
danigm, line 152: This creates a lint error: https://eslint.org/docs/rules/consistent-return

This should return a promise to make it consistent with following calls, but it's only used in the `Promise.all` this could be fixed just with an explicit `return null`.

danigm (changes requested):
There's a lint error, but besides that, it looks good to me.

packages/template-ui/src/views/Home.vue
manuq: OK! Fixed, and also fixed the returns of the method above `fetchCarouselNodes`, which was not returning the promises.

But again, this doesn't produce a lint error for me. Have you checked your eslint settings? Also I don't see the CI running for this PR, that would be the source of truth.

PR #288, opened 2021-09-27
plugin: Add ancestors to nodes

manuq:
This is a backport adapted from the current code in the kolibri
development branch at:
https://github.com/learningequality/kolibri/blob/release-v0.15.x/kolibri/core/content/api.py#L393-L439

It adds the ancestors property to all nodes.

Also, make pre-commit hook for flake8 more compatible with
black. Otherwise I can't commit this.

https://phabricator.endlessm.com/T32586

### DIFF
--- a/kolibri_explore_plugin/viewsets.py
+++ b/kolibri_explore_plugin/viewsets.py
@@ -1,11 +1,40 @@
 from kolibri.core.content.api import ContentNodeSearchViewset
 from kolibri.core.content.api import ContentNodeViewset
+from kolibri.core.content.models import ContentNode
 from kolibri.core.content.models import ContentTag
 from rest_framework.decorators import list_route
 from rest_framework.response import Response
 
 
 class CustomContentNodeViewset(ContentNodeViewset):
+    values = (
+        "id",
+        "author",
+        "available",
+        "channel_id",
+        "coach_content",
+        "content_id",
+        "description",
+        "kind",
+        # Language keys
+        "lang__id",
+        "lang__lang_code",
+        "lang__lang_subcode",
+        "lang__lang_name",
+        "lang__lang_direction",
+        "license_description",
+        "license_name",
+        "license_owner",
+        "num_coach_contents",
+        "options",
+        "parent",
+        "sort_order",
+        "title",
+        "lft",
+        "rght",
+        "tree_id",
+    )
+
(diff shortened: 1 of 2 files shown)


kolibri_explore_plugin/viewsets.py
```diff
@@ -35,11 +64,62 @@ def _consolidate(self, items, queryset):
...
+
+        def ancestor_lookup(item):
+            lft = item.get("lft")
+            rght = item.get("rght")
+            tree_id = item.get("tree_id")
+            item["ancestors"] = list(
+                map(
+                    lambda x: {"id": x["id"], "title": x["title"]},
+                    filter(
+                        lambda x: x["lft"] < lft
```
danigm, line 109: I'm testing this with kolibri `v0.14.7` and I'm getting an error:

```
  File "/home/danigm/Projects/work/endless/kolibri/kolibri-explore-plugin/kolibri_explore_plugin/viewsets.py", line 109, in <lambda>
    lambda x: x["lft"] < lft
TypeError: '<' not supported between instances of 'int' and 'NoneType'
```

I'm trying with the dance channel.

danigm (changes requested):

kolibri_explore_plugin/viewsets.py
manuq: I doubt the plugin will work with anything else than kolibri v0.14.6,

kolibri_explore_plugin/viewsets.py
```diff
@@ -35,11 +64,62 @@ def _consolidate(self, items, queryset):
...
+
+            ancestors = sorted(ancestors_map.values(), key=lambda x: x["lft"])
+        else:
+            ancestors = list(
+                queryset.get_ancestors()
+                .values("id", "title", "lft", "rght", "tree_id")
+                .order_by("lft")
+            )
+
+        def ancestor_lookup(item):
```
danigm, line 101: If the call to `super().consolidate` works correctly, the item has the `ancestors` property but not the `lft` and `rght` properties.

https://github.com/learningequality/kolibri/blob/release-v0.14.x/kolibri/core/content/api.py#L375

```suggestion
        def ancestor_lookup(item):
            if "ancestors" in item:
                return item

```

danigm (changes requested):

PR #290, opened 2021-09-29
T32508 reimplement bundles

danigm:
https://phabricator.endlessm.com/T32508

This PR uses the new data in ContentNodes added in this commit: https://github.com/endlessm/kolibri/pull/15


manuq (approved):
Great work also updating the PBS-Kids override.

PR #294, opened 2021-10-19
Update api migration

danigm:
https://phabricator.endlessm.com/T32573

### DIFF
--- a/packages/template-ui/src/views/ListSection.vue
+++ b/packages/template-ui/src/views/ListSection.vue
@@ -87,7 +87,8 @@ export default {
       isFilterEmpty: 'filters/isEmpty',
     }),
     isInlineLevel() {
-      return this.section.children_count === this.section.topic_children_count;
+      // FIXME: Use API to query the amount of subtopics:
+      return this.sectionNodes.nodes.every((n) => n.kind === 'topic');
     },
     backgroundImageURL() {
       return this.getAssetURL('sectionBackgroundImage');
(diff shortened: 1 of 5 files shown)


packages/template-ui/src/views/Section.vue
```diff
@@ -38,7 +38,8 @@ export default {
       if (!this.isSimpleBundle) {
         return false;
       }
-      return this.node.topic_children_count === 0;
+      const hasChildTopics = this.sectionNodes.some((n) => n.kind === 'topic');
```
manuq, line 41: This is not ideal. Please add a FIXME comment line:
```suggestion
      // FIXME: Use API to query the amount of subtopics:
      const hasChildTopics = this.sectionNodes.some((n) => n.kind === 'topic');
```

packages/template-ui/src/views/ListSection.vue
```diff
@@ -87,7 +87,7 @@ export default {
       isFilterEmpty: 'filters/isEmpty',
     }),
     isInlineLevel() {
-      return this.section.children_count === this.section.topic_children_count;
+      return this.sectionNodes.nodes.every((n) => n.kind === 'topic');
```
manuq, line 90:
```suggestion
      // FIXME: Use API to query the amount of subtopics:
      return this.sectionNodes.nodes.every((n) => n.kind === 'topic');
```

manuq (approved):

PR #302, opened 2021-12-01
template: Don't show topic subtitle if it's zero

danigm:
With the new API we don't have all the children in the node object, so
the number is zero. This patch just don't show anything when it's zero,
but doesn't remove the logic in the case that we want to bring back this
with a new API call or if this data is added in the future to the node.

https://phabricator.endlessm.com/T32758


packages/eos-components/src/utils.js
```diff
@@ -94,6 +94,12 @@ export function getTopicCardSubtitle(node) {
   }
 
   const count = node.children_count || leaves.length;
+
+  // Don't show anything if the number is 0
+  if (count === 0) {
+    return '';
+  }
```
manuq, line 101: I would prefer if you comment the "N resources" line directly from the Vue template, since this will always be the case in master. And ideally add a FIXME comment explaining why it's commented.

manuq (changes requested):

packages/eos-components/src/utils.js
danigm: I've updated the PR to hide the subtitle, just with a comment and the information there.

manuq (approved):
Excellent!

PR #309, opened 2021-12-22
template: Sort main sections by title

danigm:
https://phabricator.endlessm.com/T32889

### DIFF
--- a/packages/template-ui/src/components/MainSections.vue
+++ b/packages/template-ui/src/components/MainSections.vue
@@ -1,7 +1,7 @@
 <template>
   <b-button-toolbar keyNav aria-label="Sections">
     <b-button-group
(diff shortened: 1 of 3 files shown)


manuq (commented):
Code-wise looks good but let's first discuss UX in the ticket: https://phabricator.endlessm.com/T32889#919154

manuq (changes requested):
This sorts by title in the frontend which we found is not what we want:

- We want to have the same order as Kolibri Studio (by `lft`)
- We want the fix to happen in the API, not in the frontend.

danigm:
I'm closing this PR, we should fix this in the API to make it work correctly with pagination and make it consistent in content, not just sections.

PR #310, opened 2022-01-03
T32921 channel navigation

manuq:
https://phabricator.endlessm.com/T32921

### DIFF
--- a/packages/eos-components/src/components/NavBar.vue
+++ b/packages/eos-components/src/components/NavBar.vue
@@ -1,6 +1,6 @@
 <template>
   <b-navbar class="navbar px-0" :class="{ shadow: hasScrolled }" fixed="top">
-    <b-container>
+    <b-container fluid class="mx-3">
       <img v-if="showLogo" class="logo mr-3" :src="logo" @click="$emit('click-logo')">
       <slot></slot>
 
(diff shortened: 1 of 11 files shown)


packages/template-ui/package.json
```diff
@@ -20,7 +20,8 @@
     "vue": "^2.6.11",
     "vue-clamp": "^0.3.1",
     "vue-router": "^3.2.0",
-    "vuex": "^3.5.1"
+    "vuex": "^3.5.1",
+    "vue-material-design-icons": "^4.12.1"
```
danigm, line 24: I think this is because these icons are better than the default ones that we've in the bootstrap-vue lib: https://bootstrap-vue.org/docs/icons

danigm (changes requested):
Overall looks good to me. But there are some issues that should be fixed

packages/eos-components/src/components/NavBar.vue
```diff
@@ -16,7 +16,7 @@
   import EndlessLogo from '../assets/EndlessLogo.svg';
 
   export default {
-    name: 'Header',
+    name: 'NavBar',
```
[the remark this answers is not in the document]
manuq: Good point! I just created https://phabricator.endlessm.com/T32938 to work specifically in sync'ing PBS-Kids override after the main work is complete. I think that doing it now will deviate from the navigation goal.

packages/template-ui/package.json
manuq: Yes, currently we have a mix of icons, some from Bootstrap (provided by bootstrap-vue) and some for Material. Joana started adding Material icons for the play buttons, then in more and more places. I think we should use only the Material ones. I just created https://phabricator.endlessm.com/T32939 for unifying this. Needs Joana's decision first.

PR #312, opened 2022-01-04
template: Add button to go back in Header

manuq:
- Click does the same as going 1 back in window history.
- Style header buttons. This applies to the more/less buttons too.
- Remove conditional for header image in buttons or header background.

https://phabricator.endlessm.com/T32923

### DIFF
--- a/packages/template-ui/src/components/mixins/headerMixin.js
+++ b/packages/template-ui/src/components/mixins/headerMixin.js
@@ -27,8 +27,5 @@ export default {
     headerImageURL() {
       return this.sectionImageURL || this.getAssetURL('headerImage');
     },
-    hasHeaderImage() {
-      return this.headerImageURL !== null;
-    },
   },
 };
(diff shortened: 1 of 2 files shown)


packages/template-ui/src/components/mixins/headerMixin.js
```diff
@@ -27,8 +27,5 @@ export default {
     headerImageURL() {
       return this.sectionImageURL || this.getAssetURL('headerImage');
     },
-    hasHeaderImage() {
```
danigm, line 30: This method is also used in the `ChannelNavBar` component, maybe the same style should be removed there too.

https://github.com/endlessm/kolibri-explore-plugin/blob/T32923-back-button/packages/template-ui/src/components/ChannelNavBar.vue#L5

danigm (changes requested):
There's a problem with the `ChannelNavBar` component that uses the mixing modified.

```
Property or method "hasHeaderImage" is not defined on the instance but referenced during render
```

packages/template-ui/src/components/mixins/headerMixin.js
manuq: I remove it in the other PR. See #310

PR #313, opened 2022-01-05
Display zim full iframe

danigm:
https://phabricator.endlessm.com/T32928

### DIFF
--- a/packages/template-ui/src/views/Home.vue
+++ b/packages/template-ui/src/views/Home.vue
@@ -24,15 +24,6 @@
 
       <template v-if="isFilterEmpty">
         <template v-if="contentNodes.nodes.length">
-          <b-container v-if="displayHeroContent">
-            <CarouselCard
-              v-for="node in contentNodes.nodes"
-              :key="'node-' + node.id"
-              :node="node"
-              class="template-ui-hero-card"
-              @click="goToContent(node)"
-            />
-          </b-container>
           <CardGrid
             :nodes="contentNodes.nodes"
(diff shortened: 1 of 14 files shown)


packages/template-ui/src/views/Home.vue
```diff
@@ -117,7 +108,13 @@ export default {
       this.fetchCarouselNodes(),
       this.fetchContentNodes(),
       this.fetchSectionNodes(),
-    ]);
+    ]).then(() => {
+      if (this.displayHeroContent) {
+        const [first] = this.contentNodes.nodes;
+        // Showing the first node
+        this.goToContent(first);
+      }
```
dylanmccall, line 116: It looks like `displayHeroContent` isn't really doing what it says any more, so we should rename it :)

Also, I wonder what happens here if a channel has more than one content node? Perhaps we can change `displayHeroContent` to something like `defaultContentNode="12345abcdef"`, or make this only happen if `this.contentNodes.nodes.length === 1`?

dylanmccall (approved):
This looks fine to me :) One small concern, but feel free to merge when you're ready.

packages/template-ui/src/views/Home.vue
danigm: It shows just the first content node if there's more than one. I think it's simpler than the `defaultContentNode` configuration, but maybe it's a good idea to add that option and just fallback to the first if it's not set, so this can work correctly even if the content id changes in the channel.

danigm: I've added a commit to rename that option to the proposed one.

PR #314, opened 2022-01-05
T32923 back button 2

manuq:
https://phabricator.endlessm.com/T32923

### DIFF
--- a/packages/template-ui/src/components/DetailView.vue
+++ b/packages/template-ui/src/components/DetailView.vue
@@ -1,6 +1,8 @@
 <template>
   <b-container class="bg-white wrapper" fluid>
     <b-container class="main-container">
+      <h1>{{ title }}</h1>
+      <p>{{ subtitle }}</p>
       <slot>
       </slot>
     </b-container>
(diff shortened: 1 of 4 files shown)


packages/template-ui/src/components/DetailView.vue
```diff
@@ -1,6 +1,16 @@
...
+      <b-button
+        href="#"
+        pill
+        variant="light"
+        size="sm"
+        class="my-3 pl-0"
+        @click="$router.go(-1)"
+      >
+        <span><ArrowLeft />Back</span>
+      </b-button>
```
danigm, line 13: Maybe it could be a good idea to create a new component with just this button if it should looks similar in different views.

danigm (approved):
Looks good to me.

packages/template-ui/src/components/DetailView.vue
manuq: Yes. Let's see if it's worth because I think Joana made them with different styles. The one in the header has some transparency.

PR #315, opened 2022-01-06
T32919 discovery navbar

manuq:
https://phabricator.endlessm.com/T32919


.pre-commit-config.yaml
```diff
@@ -17,10 +17,10 @@ repos:
   - id: reorder-python-imports
 - repo: local
   hooks:
-  - id: lint-frontend
+  - id: lint-frontend-format
     name: Linting of JS, Vue, SCSS and CSS files on the frontend
     description: This hook handles all frontend linting for Kolibri
-    entry: yarn run lint-frontend
+    entry: yarn run lint-frontend:format
```
danigm, line 23: maybe it's better to do this change in a different commit.

kolibri_explore_plugin/assets/src/components/DiscoveryNavBar.vue
```diff
@@ -2,23 +2,30 @@
 
   <NavBar class="discovery-navbar">
     <img class="logo mr-3" :src="logo">
-    <b-button-group>
-      <b-nav-text class="btn pl-0 text-primary" @click="goBack">
+    <b-button-group class="mx-auto">
```
danigm, line 5: This is not centering the content in the whole bar space, it's centering in the available space, so it's not really centered because the right widget is larger, so it's placed a bit to the left.

Maybe we can use this simple css positioning that I've found in [stackoverflow](https://stackoverflow.com/a/42181069):

```
.abs-center-x {
    position: absolute;
    left: 50%;
    transform: translateX(-50%);
}
```

danigm (approved):
Overall looks good to me. I've just found a problem with the centering of the tabs, but it's not a blocker, we can improve in a following PR.

Also a nitpick about the change in the lint configuration change to go to a different commit.

kolibri_explore_plugin/assets/src/components/DiscoveryNavBar.vue
manuq: Thanks for checking! I really dislike the absolute positioning for this. Let's ask Joana when she's back and polish accordingly.

.pre-commit-config.yaml
manuq: Yeah. The problem was that pre-commit wouldn't let me commit when the configuration is changed. But I'll split it!

PR #316, opened 2022-01-07
discovery: Improve going back the custom presentation

manuq:
Go back to the corresponding page, not always to the home page as
before.

- vue-router handler: Store the current page in the vuex state before
navigating to the custom presentation.

- Custom channel presentation: Use that state when the event from the
  custom presentation happens to navigate programatically to it.

Also fixes the Navigation cancelled error that happened always when
going back from a custom presentation.

https://phabricator.endlessm.com/T32915


danigm (approved):
Looks good to me

PR #317, opened 2022-01-07
Improve detail page

danigm:
https://phabricator.endlessm.com/T32916

### DIFF
--- a/packages/eos-components/src/components/PlayButton.vue
+++ b/packages/eos-components/src/components/PlayButton.vue
@@ -1,7 +1,8 @@
 <template>
   <b-button
     pill
-    size="sm"
+    :size="size"
+    :block="block"
     :variant="`${kind}-${variant}`"
     class="border-0 pb-1 pl-1 play-button pt-1 text-nowrap"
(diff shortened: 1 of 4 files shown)


packages/eos-components/src/components/PlayButton.vue
```diff
@@ -1,7 +1,8 @@
 <template>
   <b-button
     pill
-    size="sm"
+    :size="size"
+    :block="block"
```
manuq, line 5: :+1:

packages/template-ui/src/views/Content.vue
```diff
@@ -5,8 +5,15 @@
       :title="content.title"
       :subtitle="subtitle"
     >
+      <b-row class="my-3">
+        <b-col xs="12" md="6">
```
manuq, line 9: So with this the card occupies half of the width in large screens. I don't think this is Joana's intention but is a good start, so it's ready for merging. But please mention this implementation decision in the commit message before merging.

manuq (approved):

PR #318, opened 2022-01-10
discovery: keep the SearchPage state during navigation

danigm:
Use the keep-alive wrap in the SearchPage to keep the page state.

https://phabricator.endlessm.com/T32940


dylanmccall (approved):
Ooh, `keep-alive` is interesting! This looks good to me :)

PR #320, opened 2022-01-11
Discovery: Don't show debug info in iframes

danigm:
https://phabricator.endlessm.com/T32946


dylanmccall (approved):
Yay! Thanks for fixing this :)

PR #322, opened 2022-01-12
components: Detect bootstrap breakpoints in realtime

danigm:
The kolibri-explore-plugin module is built using kolibri-tools and the
import of styles.scss variables doesn't work there, so the
responsiveMixin is not giving the correct values.

This patch detects the breakpoint in the document styles in real time
instead loading from the scss in build time.

https://phabricator.endlessm.com/T32947

### DIFF
--- a/packages/eos-components/src/components/mixins/responsiveMixin.js
+++ b/packages/eos-components/src/components/mixins/responsiveMixin.js
@@ -1,5 +1,3 @@
-import { sm, md, lg, xl } from '../../styles.scss';
-
 export default {
   data() {
     return {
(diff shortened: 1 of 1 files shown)


packages/eos-components/src/components/mixins/responsiveMixin.js
```diff
@@ -58,5 +59,20 @@ export default {
     onXlChange(ev) {
       this.xl = ev.matches;
     },
+    getBreakpoints() {
+      if (!this.breakpointValues) {
+        // Get breakpoints defined by bootstrap styles. This code is based on:
+        // https://github.com/shaack/bootstrap-detect-breakpoint/blob/master/src/bootstrap-detect-breakpoint.js
```
manuq, line 65:
```suggestion
        // Get breakpoints defined by bootstrap styles. Ideally we should be able to just import the variables like:
        // import { sm, md, lg, xl } from '../../styles.scss';
        // But unfortunately kolibri-tools is not capable of doing such import. 
        // This code is based on:
        // https://github.com/shaack/bootstrap-detect-breakpoint/blob/master/src/bootstrap-detect-breakpoint.js
```

manuq (approved):
This looks good and it shouldn't affect the performance much. Still it would be much better to solve this in kolibri-tools than doing this runtime check. So could you add a comment about it?

And we have another import like this in cardMixin:
```
import { cardImageAspectRatio } from '../../styles.scss';
```
I wonder if that is problematic too.

PR #327, opened 2022-01-14
Update channel search

danigm:
https://phabricator.endlessm.com/T32930


packages/template-ui/src/components/SectionsSearchRow.vue
```diff
@@ -1,24 +1,14 @@
...
-      </b-col>
-    </b-row>
+    <MainSections>
+      <b-button
+        class="font-weight-bold my-1 text-nowrap"
+        variant="link"
+        to="/search"
+      >
+        <b-icon-search class="mr-1" />
+        Search Keywords
```
manuq, line 10: Great! Adding the link inside a slot works for me.

packages/eos-components/src/components/Keywords.vue
```diff
@@ -0,0 +1,34 @@
...
+      v-for="w in words"
+      :key="w"
+      pill
+      variant="outline-primary"
+      class="mx-1 pr-1 py-1"
+      @click="$emit('click', w)"
+    >
+      {{ w }}
+      <slot>
+        <CloseIcon :size="20" class="ml-2" />
```
manuq, line 13: The `size` here is a constant, so I think that it doesn't need to be dynamic. I. e. it doesn't need to be a v-bind (or in this case the `:` shorthand)

manuq (approved):
Excellent! Very well split in individual commits, which makes the review a piece of cake. Only a nit comment to be fixed.

packages/eos-components/src/components/Keywords.vue
danigm: The `:` is needed because the size on material icon components should be a number and not an string:
```
Invalid prop: type check failed for prop "size". Expected Number with value 20, got String with value "20".
```

In any case, I'm moving this to the `constants.js` file so it's not defined here.

manuq: Ah OK, good!

PR #329, opened 2022-01-14
Back to top

danigm:
https://phabricator.endlessm.com/T32922

### DIFF
--- a/packages/template-ui/src/App.vue
+++ b/packages/template-ui/src/App.vue
@@ -1,5 +1,6 @@
 <template>
   <div id="app" class="d-flex flex-column h-100">
+    <BackToTop />
     <!-- Wrapper needed to fix flexbox footer positioning on IE11 -->
     <div class="flex-fill flex-shrink-0">
       <router-view />


packages/eos-components/src/components/BackToTop.vue
```diff
@@ -0,0 +1,59 @@
...
+  }
+</script>
+
+<style lang="scss" scoped>
+  @import '../styles.scss';
+
+  .back-to-top {
+    position: fixed;
+    bottom: 10px;
+    right: 10px;
```
manuq, line 49: Let's use the `$spacer` variable here instead of hardcoding. Doing that ensures that all fits in the grid. Also, adding hardcoded values here and there makes the project very messy and hard to maintain. This comes up a lot in reviews, please remind it.

manuq (commented):
Looks great. Feel free to merge after fixing the hardcoded spacing.

packages/eos-components/src/components/BackToTop.vue
danigm: Good catch, I've just replaced that with `$spacer`.

PR #330, opened 2022-01-17
component: Update Card components

danigm:
The new Card component has the description as subtitle, and the author
information just when isEndlessApp.

The TopicCard goes with the subtitle and a new badge at bottom.

All tags are removed from cards.

https://phabricator.endlessm.com/T32924


manuq (commented):
Great code removal!

PR #331, opened 2022-01-17
component: Update CarouselCard components

danigm:
The new CarouselCard component has the description as subtitle, and the
author information just when isEndlessApp.

https://phabricator.endlessm.com/T32926


manuq (changes requested):
Looks great! Only thing I note is that the smooth transition in the title on mouse over is lost, turning from white to orange (now it's immediate). Can you fix it and then merge?

PR #332, opened 2022-01-18
force 'lft' ordering in custom API

danigm:
https://phabricator.endlessm.com/T32889

This is a second solution to the ordering problem with the new API. If LE doesn't accept the simple default ordering change we can try to propose this one: https://github.com/endlessm/kolibri/pull/22/files


danigm:
Not needed anymore: https://github.com/learningequality/kolibri/pull/9002

PR #333, opened 2022-01-18
Polish detail page

danigm:
https://phabricator.endlessm.com/T32942

### DIFF
--- a/packages/template-ui/src/components/ContentImage.vue
+++ b/packages/template-ui/src/components/ContentImage.vue
@@ -12,6 +12,7 @@
         size="lg"
         block
         :kind="node.kind"
+        :style="buttonStyle"
         class="mx-auto my-3 play-button py-2"
         @click="goToContent()"
       />
@@ -30,6 +31,28 @@ export default {
       type: Object,
       required: true,
     },
+  },
+  data() {
+    return {
+      buttonMaxWidth: '100%',
+    };
+  },
+  computed: {
+    buttonStyle() {
+      return {
+        maxWidth: this.buttonMaxWidth,
+        minWidth: '200px',
+      };
+    },
+  },
+  watch: {
+    thumbnail() {
+      const img = new Image();
+      img.src = this.thumbnail;
+      img.onload = () => {
+        this.buttonMaxWidth = `${img.width}px`;
+      };
+    }
   },
   methods: {
     goToContent() {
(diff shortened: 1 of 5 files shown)


manuq (changes requested):
As you mentioned in the ticket, this doesn't solve the thumbnail size issue, which I think is the main problem here. For the rest, see the comments.

packages/template-ui/src/components/ContentImage.vue
```diff
@@ -1,18 +1,23 @@
 <template>
   <b-link @click="goToContent()">
-    <b-card class="rounded-lg">
+    <b-card
+      class="content-image rounded-lg"
+      :class="{ 'mx-auto': bundleKind }"
```
manuq, line 5: This is taking advantage of the `bundleKind` being only `"simple"` or `null` today, but is not future-proof. Please use the `isSimpleBundle` getter instead of using the `bundleKind` state directly.

packages/template-ui/src/components/ContentImage.vue
```diff
@@ -31,6 +38,36 @@ export default {
...
+      const min = contentImageMinWidth;
+      const maxWidth = this.imageSize >= min ? this.imageSize : min;
+      return {
+        maxWidth,
+      };
+    },
+    cardStyle() {
+      const margin = cardSpacerX;
+      return {
+        maxWidth: `calc(${this.imageSize}px + ${margin})`,
```
manuq, line 58: Yeah, from the CSS is not possible to do things conditionally for the image size. This is not a good practice but is what design wants today.

packages/template-ui/src/components/DetailView.vue
```diff
@@ -11,16 +11,37 @@
...
-      <slot>
-      </slot>
+      <b-row>
+        <b-col>
+          <h1>{{ title }}</h1>
+        </b-col>
+      </b-row>
+      <b-row>
+        <b-col
+          v-if="!bundleKind"
```
manuq, line 21: Please use the `isSimpleBundle` getter instead of using the `bundleKind` state directly.

packages/template-ui/src/components/DetailView.vue
```diff
@@ -51,4 +75,8 @@ $main-container-bottom: $spacer * 3.5;
   padding-bottom: $main-container-bottom;
 }
 
+.wrapper {
+  padding: 0;
```
manuq, line 79: I think there is a Bootstrap class `.p-0` for this.

manuq (changes requested):
Small nitpicks but it looks good!

manuq (approved):

PR #334, opened 2022-01-20
T32927 row of cards

manuq:
https://phabricator.endlessm.com/T32927

### DIFF
--- a/packages/eos-components/src/components/CardGrid.vue
+++ b/packages/eos-components/src/components/CardGrid.vue
@@ -1,5 +1,8 @@
 <template>
-  <b-container class="mb-5 mt-4 section-container">
+  <b-container
+    class="mb-5 mt-4 section-container"
+    :class="{ 'no-container-padding': variant === 'slidable' }"
+  >
     <slot></slot>
 
     <component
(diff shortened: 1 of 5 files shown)


danigm (changes requested):
Looks almost good, I've just detected a minor issue with the `no-container-padding`. In the [channel search](https://github.com/endlessm/kolibri-explore-plugin/blob/master/packages/template-ui/src/views/Search.vue#L25) the text `Explore Topics` has no left padding.

This could be fixed just adding a container there. We should fix this kind of issues to avoid regressions.

manuq:
> Looks almost good, I've just detected a minor issue with the `no-container-padding`. In the [channel search](https://github.com/endlessm/kolibri-explore-plugin/blob/master/packages/template-ui/src/views/Search.vue#L25) the text `Explore Topics` has no left padding.
> 
> This could be fixed just adding a container there. We should fix this kind of issues to avoid regressions.

Good catch with the regression! Since "slidable" is the default card grid variant, I hope it's the only one. I've been navigating the channels and the discovery page and I couldn't find any other.

PR #335, opened 2022-01-21
Discovery: Reset search input on Search tab select

danigm:
The query attribute was not reset when showing the SearchPage, this
patch sync the query attribute in  the SearchPage with the searchTerm in
the store, and it's set to empty on clicking in the tab.

The searchTerm is not updated with the URL if the query parameter is not
passed as part of the url, so it's keeped when navigating to content and
back.

https://phabricator.endlessm.com/T32988


manuq (approved):
Great fix! One nit about the commit message: it's better to write about what the commit does, rather than explaining what was the previous situation.

PR #336, opened 2022-01-21
Discovery: Disable search input during search

danigm:
This patch disables the SearchBar input during the loading to avoid
edit the query during a search so the text input is disabled until the
API query is completely done.

After the search the input grab the focus again, so it's easy to change
the query without clicking anywhere.

https://phabricator.endlessm.com/T32917


manuq (approved):
Excellent.

PR #339, opened 2022-01-24
template: Replace route when receiving params

manuq:
In App we do a bit of routing when receiving parameters, to go to
either a specific content, topic, or the test page.

We shouldn't account this as two navigations in the browser
history. Use router replace instead of router push to do it.

https://phabricator.endlessm.com/T33012


danigm (approved):
Good catch! Let's get this merged.

PR #344, opened 2022-01-25
template: Move defaultContentNode logic to a NavigationGuard

danigm:
Instead doing the redirection in the component, this patch does the
redirection using a vue-router NavigationGuard so this is done before
loading the Home component.

With this change the ability of loading the first ContentNode setting
the defaultContentNode to true is lost, but it's not a big issue for the
Endless Key.

https://phabricator.endlessm.com/T33009


manuq (approved):
Definitely better! Also using router push was adding the home to the browser navigation history, breaking the Back button.

PR #345, opened 2022-01-25
template: Add DetailViewFullScreen component

danigm:
This new component was in the zim content channel overrides and it's now
in the template components to be able to reuse the same component in
different overrides.

The new option to use this component is showDetailViewFullScreen.

There's also a new option to do not show the full breadcrumb,
showBreadcrumb, that's set to false in all the zim content overrides.

This change make it possible to remove the custom widgets in these
overrides.

https://phabricator.endlessm.com/T33014


packages/template-ui/src/components/Breadcrumb.vue
```diff
@@ -20,19 +20,21 @@
...
-      <li
-        v-for="p in parentItems"
-        :key="p.id"
-        v-b-tooltip.hover
-        :title="p.title"
-        class="breadcrumb-item text-light text-truncate"
-      >
-        <b-link
-          :to="getTopicUrl(p)"
+      <template v-if="showBreadcrumb">
```
manuq, line 23: What about the "..." above? it seems you also don't want those when `showBreadcrumb` is false.

packages/template-ui/src/components/Breadcrumb.vue
```diff
@@ -20,19 +20,21 @@
...
-        </b-link>
-      </li>
+          <b-link
+            :to="getTopicUrl(p)"
+          >
+            {{ p.title }}
+          </b-link>
+        </li>
+      </template>
       <!-- Last li, current node or Search -->
```
manuq, line 38: So if you still want the current node in the breadcrumb, then you still have a breadcrumb . What you really want is to hide the parents from the breadcrumb. It would be better to rename the setting to "showParentsInBreadcrumb".

This breadcrumb component was a mess of conditionals and it took me a while to refactor it in 2b72fde1 . To avoid adding a template and more conditionals, I would do some restructuring:
- Move the `<b-navbar-nav>` and channel logo to the ChannelNavBar component
- Rename this component to BreadcrumbParents, with the "..." and the parent items only.
- Also move the last item (current node or search) to the ChannelNavBar component

manuq (approved):
The detail page variant looks great and uses composition. The breadcrumb part needs some restructuring.

packages/template-ui/src/components/Breadcrumb.vue
danigm: I've added a new commit renaming the option to `showParentsInBreadcrumb` and also do some of the restructuring suggestions:

 * [x] Move the <b-navbar-nav> and channel logo to the ChannelNavBar component
 * [ ]  Rename this component to BreadcrumbParents, with the "..." and the parent items only.
 * [ ] Also move the last item (current node or search) to the ChannelNavBar component

I was unable to implement the last two items, because it's not simple to have just the parents in one component and the last item in a different one. To implement the `BreadcrumbParents` as you said, we need to move the `<ol>` component to the `ChannelNavBar`, but it's not possible to have multiple `<li>` elements as root in a new `BreadcrumbParents` component, and if I try to wrap with a `span` or a `div`, all the bootstrap styles are broken, and I think that the HTML output is not good.

One way to bypass this could be to just keep the `ol` in the `Breadcrumb` component and add there an slot to add the parents and the last one, but again, the parents could not be a single component.

Another way could be to do not use `ol` and `li` and change the styles, but I don't like this option.

manuq: OK! Thanks for doing the effort and for the explanation.

PR #347, opened 2022-01-25
scripts: Add script to ingest highlighted content

manuq:
Add a script that:
- Reads a spreadsheet sheet
- Parses the rows, and
- writes a JSON file.

The sheet has sets of rows. The first set is content to be highlighted
in the Discovery page. The other sets must start with a channel ID,
and it's one set per channel.

The data in the JSON will be used later for filling the carousels with
content with a rotation logic.

Invalid IDs are ignored, except for the channel IDs. If a channel ID
is invalid, we raise an exception to stop execution.

Read spreadsheet key from a local, uncommited `.env` file. Pipenv
handles this automatically. Explain in the README. And add the file to
the list ignored by GIT.

https://phabricator.endlessm.com/T33016

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,6 @@ apps-bundle.zip
 
 # Symlink to template overrides
 packages/template-ui/src/overrides
+
+# Environment file with secrets
+/.env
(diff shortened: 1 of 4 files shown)


danigm (approved):
Looks good to me.

PR #351, opened 2022-01-27
Update default content node for WikiHow

dylanmccall:
https://phabricator.endlessm.com/T32799


manuq (changes requested):
Why this change? It's not explained in the commit message or ticket.

dylanmccall:
This change is simply that the main content node ID has changed in the Kolibri channel. (That happened because the zim file's source ID changed). I updated the commit message to explain that now :) Unfortunately, we're going to have to chase this kind of thing until channels can provide their own metadata.

danigm (approved):

PR #353, opened 2022-01-27
T33030 channels list

manuq:
https://phabricator.endlessm.com/T33030

### DIFF
--- a/kolibri_explore_plugin/assets/src/views/SearchPage.vue
+++ b/kolibri_explore_plugin/assets/src/views/SearchPage.vue
@@ -3,67 +3,74 @@
   <div class="d-flex flex-column min-vh-100 search-page">
     <AboutModal id="about-modal" />
     <DiscoveryNavBar />
-    <div class="flex-fill main">
+
+    <div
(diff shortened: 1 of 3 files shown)


kolibri_explore_plugin/assets/src/views/SearchPage.vue
```diff
@@ -3,67 +3,74 @@
...
+          </template>
+          <template v-else>
+            <template v-if="!isEmpty">
+              <Keywords :words="keywords" @click="removeKeyword" />
+            </template>
+            <div v-if="isNoResults" class="empty mb-4 mt-5 w-50">
+              <h1>Sorry, we can’t find any content that matches your search.</h1>
+              <h5>
+                You can try a different search, maybe use fewer words, or try one
+                of the topic suggestions below.
```
danigm, line 44: There's no topic suggestions below after this change, maybe we should change the wording. But it could be done as a separate task.

danigm (approved):
Looks good to me, just a minor issue with no results message inconsistency, but we can fix that in a different PR.

PR #356, opened 2022-01-28
T33017 highlighted

manuq:
https://phabricator.endlessm.com/T33017


kolibri_explore_plugin/assets/src/kolibriApi.js
```diff
@@ -110,6 +112,56 @@ class KolibriApi {
     return searchPromise;
   }
 
+  getHighlightedContent(options) {
```
danigm, line 115: I'm not sure if this function is okay here, this API is a "copy" of the upstream custom presentation API, using this API, as is, in the custom presentation will break the custom presentation outside of the Endless Key.

I think it's okay to do this right now, but we need to keep in mind that we should move this to a difference place when we want to use the upstream custom presentation, for example in the EOS flatpak app.

Maybe it could be a good idea to just import the highlighted content json in the javascript code, that will require to rebuild with every change, but that's something that we can assume.

danigm (approved):
Looks good.

PR #357, opened 2022-01-31
Add overrides for WikiHow category channels

dylanmccall:
https://phabricator.endlessm.com/T33064


erikos (approved):
Ok, I compared the overriedes with the ones of Wikipedia and they do match. Also all of the wikihow files have been added.

PR #361, opened 2022-02-02
Discovery: Initial search inside the zim files

danigm:
Iterates over all the defined ZimSearchChannels and look for all the zim
content nodes to search for articles there. To search it's used the
kolibri-zim-plugin JSON API.

https://phabricator.endlessm.com/T33078

### DIFF
--- a/kolibri_explore_plugin/assets/src/customApps.js
+++ b/kolibri_explore_plugin/assets/src/customApps.js
@@ -250,6 +250,24 @@ export const ChannelCardIcons = {
   f62db29be20453c4a267132e93a9e602: 'wikipedia.png',
 };
 
+export const ZimSearchChannels = {
+  Wikipedia: ['f62db29be20453c4a267132e93a9e602'],
+  WikiHow: [
+    '58828121cc115549bda7fb3f71f4795e',
+    '05a4060cf5175df78edec8dc09e396a1',
+    '69497db42937539e94fe3969f6995d2d',
+    '00b0d8dff13e518d801ef816227b2691',
+    'b9fe8fe9b92b5af5964feccdadb78413',
+    'e114fe9de66e513384e4ca2e9b529c18',
+    '9c494b7b879e512f96f60176df85968c',
+    '304d529d4b1c597ea7d3c59b30ac3c14',
+    'afbf0498da3c5db3947c291ddda8e21a',
+    '146efed5aa585229a49513219e73af66',
+    'e4e4d0e342f951b09aa9ef1622f5301f',
+    '6f6dd2720bd2574fa70a069ac9b75046',
+  ],
+};
+
 export const GameAppIDs = [
   '3160899a73564d8a8467284d9219b91c', // Terminal Two
 ];
--- a/kolibri_explore_plugin/assets/src/searchApi.js
+++ b/kolibri_explore_plugin/assets/src/searchApi.js
@@ -18,6 +18,7 @@ function zimToNode(zimArticle, node) {
     channelId: node.channel_id,
     description: zimArticle.snippet,
     kind: 'zim',
+    directContent: true,
     nodeUrl: `/topics/${node.channel_id}/c/${node.id}?zimPath=${zimArticle.path}`,
   };
 }
--- a/kolibri_explore_plugin/assets/src/searchApi.js
+++ b/kolibri_explore_plugin/assets/src/searchApi.js
@@ -1,0 +1,55 @@
+import { KolibriApi } from './kolibriApi';
+
+function searchInsideZimFile(zimfile, searchParams) {
+  const { origin } = window.location;
+  const url = new URL(`${origin}/zim/search/${zimfile.checksum}.zim`);
+  for (const k in searchParams) {
+    url.searchParams.append(k, searchParams[k]);
+  }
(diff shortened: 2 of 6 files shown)


kolibri_explore_plugin/assets/src/searchApi.js
```diff
@@ -0,0 +1,56 @@
+import { KolibriApi } from './kolibriApi';
+
+function zimSearch(zimfile, query, node, maxResults = 10, start) {
```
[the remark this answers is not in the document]
manuq: And if `node` can be removed from the parameters, it would be better to pass query, maxResults, start in a `searchParams` object directly.

kolibri_explore_plugin/assets/src/searchApi.js
```diff
@@ -0,0 +1,56 @@
...
+      const promises = results.map(node => {
+        const [zimfile] = node.files.filter(f => f.extension === 'zim');
+        return zimSearch(zimfile, query, node);
+      });
+      return Promise.all(promises);
+    })
+    .then(nodes => {
+      // Convert the result to a ContentNode like object
+      return nodes
+        .map(n => n.articles.map(a => zimToNode(a, n.node)))
```
manuq, line 53: This conversion will be expensive. I wonder if we can do it when clicking on the card instead of when fetching the results.

kolibri_explore_plugin/assets/src/views/SearchPage.vue
```diff
@@ -259,15 +288,28 @@
...
-            this.progress += 100 / kinds.length;
+            this.progress += 100 / this.totalSearchSections;
           });
         });
+
+        this.resultZim = [];
+        for (const k in ZimSearchChannels) {
+          wikiChannelSearch(ZimSearchChannels[k], query)
+            .then(r => {
+              this.resultZim.push([k, r]);
```
manuq, line 306: Please make `k` and `r` be readable, understandable names. I see that `ZimSearchChannels` is actually groups of channels, multiple WikiHow channels and the Wikipedia with only one ID in the group.

kolibri_explore_plugin/assets/src/searchApi.js
manuq, line 3: I would call this function `searchInsideZimFile`. Starting with the "search" verb is better to explain the function.

kolibri_explore_plugin/assets/src/searchApi.js
```diff
@@ -0,0 +1,56 @@
...
+    .getContentByFilter({
+      kinds: ['zim'],
+    })
+    .then(({ results }) => {
+      // Do the search for each zim node
+      const promises = results.map(node => {
+        const [zimfile] = node.files.filter(f => f.extension === 'zim');
+        return zimSearch(zimfile, query, node);
+      });
+      return Promise.all(promises);
```
manuq, line 48: You can return `node` here along with the promise returned by `Promise.all`. And avoid passing it to the `zimSearch` function.

manuq (changes requested):

kolibri_explore_plugin/assets/src/searchApi.js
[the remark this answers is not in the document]
danigm: The `node` is the `ContentNode` with the zim file attached, that way we can redirect to the real node and add the node information in the `zimToNode` function.

This is done that way because a `cotentNode` can have more than one zim file, so a `zimSearch` promise is created for each zim file in each ContentNode, so after all promises, the node could be different. But maybe this could be made with more than one group of promises.

kolibri_explore_plugin/assets/src/searchApi.js
danigm: every call to `zimSearch` has a different `node`. In any case I'll try to avoid to pass the node to the `zimSearch` function.

danigm:
I've just added a new commit with some fixes.

kolibri_explore_plugin/assets/src/customApps.js
```diff
@@ -188,6 +188,24 @@ export const ChannelCardIcons = {
   f62db29be20453c4a267132e93a9e602: 'wikipedia.png',
 };
 
+export const ZimSearchChannels = {
+  Wikipedia: ['f62db29be20453c4a267132e93a9e602'],
+  'Wiki How': [
```
manuq, line 193:
```suggestion
  'WikiHow': [
```

manuq (approved):
The changes in the last commit looks good. Please rebase before merging the PR, because the last commit is undoing things done in the previous commits.

kolibri_explore_plugin/assets/src/searchApi.js
manuq: I think you don't understand me but no problem, this whole search stuff needs a refactor anyways.

manuq (approved):
Looks good! Nedless to say that this adds one more technical debt, searching ZIM files should be upstreamed.

danigm:
> Looks good! Nedless to say that this adds one more technical debt, searching ZIM files should be upstreamed.

Upstreaming this will be challenging, because the zim search is done in the plugin, so the core kolibri doesn't know anything about what's inside the zim files, so upstream this will require to extend the plugin system to support some kind of search extension.

PR #364, opened 2022-02-03
template: New BackButton component

danigm:
Add a new component for the template Back Button. This patch also
modifies the default positioning of this new button to place it on the
left margin of the header.

https://phabricator.endlessm.com/T33081


dylanmccall (approved):
This all looks just fine to me :)

PR #366, opened 2022-02-04
template: Support dark background in BackButton

danigm:
https://phabricator.endlessm.com/T33081


manuq (approved):
Yes the color variant looks well implemented.

PR #370, opened 2022-02-10
Discovery: Show search results as soon as possible

danigm:
Show the results as soon as they are available, per kind search. The
placeholder is below the channels so it loads from top to bottom and the
order will be "random", depending on the speed of different calls to the
API.

https://phabricator.endlessm.com/T33127


manuq (changes requested):
At first test, I see a bug with the order of the words component:

Captura de pantalla de 2022-02-10 10-06-14

The words to remove should be always below the text input.

PR #372, opened 2022-02-10
Increase search debounce time and trigger with Enter

danigm:
Increase the debounce time to 2 seconds and capture the Enter key event
to trigger the search when it happens.

https://phabricator.endlessm.com/T33126

### DIFF
--- a/packages/eos-components/src/components/SearchBar.vue
+++ b/packages/eos-components/src/components/SearchBar.vue
@@ -42,6 +42,7 @@
   import _ from 'underscore';
   import CloseIcon from 'vue-material-design-icons/CloseCircleOutline.vue';
   import MagnifyIcon from 'vue-material-design-icons/Magnify.vue';
+  import { SearchBarDebounce } from '../constants.js';
 
   export default {
     name: 'SearchBar',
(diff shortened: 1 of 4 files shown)


packages/template-ui/src/views/Search.vue
```diff
@@ -4,7 +4,11 @@
     :style="{ backgroundImage: backgroundImageURL }"
   >
     <ChannelNavBar />
-    <SearchBar v-model="query" @clear-input="onClearInput" />
+    <SearchBar
+      v-model="query"
+      :debounce="2000"
```
manuq, line 9: This PR shows that this delay constant is hardcoded in two components. Let's take this opportunity to move it to the constants.js module, so next time it can be edited in a single place.

packages/eos-components/src/components/SearchBar.vue
```diff
@@ -23,6 +23,7 @@
           placeholder="Type keywords to search"
           :disabled="loading"
           :value="value"
+          @keyup.enter="inputUpdated($event.target.value)"
```
manuq, line 26: :+1:

manuq (approved):
Looks great, one refactor needed.

PR #377, opened 2022-02-15
Add override for wikiHow (PDF format) channel

erikos:
We are adding a wikiHow channel with conten tin PDF format. This reuses the existing wikihow override that was used for the full ZIM based wikiHow channel which is not in use anymore.

https://phabricator.endlessm.com/T33141


danigm (approved):
Looks good to me.

PR #378, opened 2022-02-15
template: avoid onload race condition

danigm:
This patch unbind the onload function when loading a new image to avoid
race conditions when the thumbnail changes.

https://phabricator.endlessm.com/T32942

### DIFF
--- a/packages/template-ui/src/components/ContentImage.vue
+++ b/packages/template-ui/src/components/ContentImage.vue
@@ -41,6 +41,7 @@ export default {
   data() {
     return {
       imageSize: 0,
+      image: new Image(),
     };
   },
   computed: {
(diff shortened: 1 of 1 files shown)


packages/template-ui/src/components/ContentImage.vue
```diff
@@ -64,10 +65,12 @@ export default {
   },
   watch: {
     thumbnail() {
-      const img = new Image();
-      img.src = this.thumbnail;
-      img.onload = () => {
-        this.imageSize = img.width;
+      // Cancel previous onload events
+      this.loadingImg.onload = null;
```
manuq, line 69: Please use explicit names: `loadingImg` should be `loadingImage`. Just 2 more characters, it shouldn't be much more typing :P Also we have smart editors nowadays.

Also, maybe starting the variable name with a verb is not a good idea. My first impression is that this is a boolean. What about `thumbnailImage`? Or just `image`, since we already have an `imageSize` variable.

manuq (changes requested):

manuq (approved):

PR #384, opened 2022-02-16
template: Same breadcrumb styles for pbs-kids

danigm:
https://phabricator.endlessm.com/T32950


manuq (commented):
This seems a rather inoffensive change. Great to have it.

PR #387, opened 2022-02-17
Discovery: Scroll to top when clicking the Search tab

manuq:
This is not needed when clicking on the Channels tab because that view
always refreshes it's content.

Also, fix the Javascript in the method which had commas instead of
semicolons.

https://phabricator.endlessm.com/T33179


danigm (approved):
LGTM

PR #389, opened 2022-02-18
T33197

erikos:
Polish for the Discovery page categories.

https://phabricator.endlessm.com/T33197


manuq (approved):
Yes, this recategorization looks good. There are no multiple wikiHow channels for now.

PR #390, opened 2022-02-18
Discovery: Fix initial query in SearchPage

manuq:
Initialize query when the component is created. We already have a
watcher for the next changes.

https://phabricator.endlessm.com/T33195


erikos:
Looks good to me.

PR #393, opened 2022-02-21
Discovery: Drop search results from previous queries

danigm:
This patch adds a new optional parameter to the searchChannels function
to be able to drop search results for old searches.

The SearchPage component increments the searchQueryIndex for every
search, so if there's a race condition and a search query done before
ends after a second search query has been launched, it will be dropped
because the searchQueryIndex returned will be different from the one
specified in the new search query.

https://phabricator.endlessm.com/T33202

### DIFF
--- a/kolibri_explore_plugin/assets/src/components/DiscoveryNavBar.vue
+++ b/kolibri_explore_plugin/assets/src/components/DiscoveryNavBar.vue
@@ -52,7 +52,6 @@
     },
     methods: {
       ...mapMutations({
-        setSearchResult: 'topicsRoot/SET_SEARCH_RESULT',
         setSearchTerm: 'SET_SEARCH_TERM',
       }),
       goToChannels() {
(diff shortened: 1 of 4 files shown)


manuq (changes requested):
This is customizing the API that we have already upstreamed. Please mention this in the code as a comment, and try to do it in a way that leaves the original API intact (see `getRandomNodesReal` for an example).

danigm:
> This is customizing the API that we have already upstreamed. Please mention this in the code as a comment, and try to do it in a way that leaves the original API intact (see `getRandomNodesReal` for an example).

This commit doesn't change the Kolibri API, the change is done in the `kolibri-explore-plugin` custom function `searchChannels` that's only used in our discovery search page. This problem only happens here because we've the component as `keep-alive`, so when going back to a different view, the old searches are running in the backend.

In any case, the index modifies the function extending it, but the interface is the same, because the new parameter is optional, if it's not provided, the result will be the same, because the `searchQueryIndex` will be always `0`, so no results will be dropped.

manuq (changes requested):

manuq (commented):
This seems to work, but doing such refactor at this point of the release cycle is not safe. It is better to release with the bug, IMO.

PR #397, opened 2022-02-22
discovery: Ignore previous search promises

manuq:
By keeping track of the last promise per each kind, we are able to
ignore the previous ones by not resolving them.

This fixes issues with old promises provoking inconsistent or
duplicate results coming from a previous search.

This is done in a new `searchLastChannels()` method which wraps
`searchChannels()` so it can be easily replaced.

Also, clear the progress bar and mark all promises as "old" when the
component is deactivated.

https://phabricator.endlessm.com/T33202

### DIFF
--- a/kolibri_explore_plugin/assets/src/modules/topicsRoot/handlers.js
+++ b/kolibri_explore_plugin/assets/src/modules/topicsRoot/handlers.js
@@ -181,40 +181,55 @@ export function showFilteredChannels(store) {
   );
 }
 
-export function searchChannels(store, search, kind) {
-  store.commit('CORE_SET_PAGE_LOADING', true);
-  return ContentNodeSearchResource.fetchCollection({
+export function searchChannelsOnce(store, search, kind) {
+  const searchPromise = ContentNodeSearchResource.fetchCollection({
     getParams: {
       search,
       kind,
       max_results: SEARCH_MAX_RESULTS,
     },
-  }).then(searchResults => {
-    const { rootNodes } = store.state.topicsRoot;
-    const { channel_ids, results } = searchResults;
-    const nodes = results.map(n => ({
-      ...n,
-      channel: rootNodes.find(c => c.id === n.channel_id),
-    }));
-    const promises = channel_ids.map(id => ChannelResource.fetchModel({ id }));
-    Promise.all(promises).then(collection => {
-      const channels = collection
-        .map(c => ({
-          ...c,
-          thumbnail: getChannelIcon(c),
-          title: c.name,
-          order: channel_ids.indexOf(c.id),
-        }))
-        .sort((a, b) => a.order - b.order);
-      store.commit('topicsRoot/SET_SEARCH_RESULT', {
-        ...store.state.topicsRoot.searchResult,
-        [kind]: {
-          ...searchResults,
-          channels: channels,
-          results: nodes,
-        },
-      });
-      store.commit('CORE_SET_PAGE_LOADING', false);
+  });
+  store.commit('topicsRoot/SET_LAST_SEARCH_PROMISE', { kind, searchPromise });
+  return new Promise((resolve, reject) => {
+    searchPromise.then(searchResults => {
+      const { lastSearchPromises } = store.state.topicsRoot;
+      if (searchPromise == lastSearchPromises[kind]) {
+        store.commit('topicsRoot/SET_LAST_SEARCH_PROMISE', { kind, searchPromise: null });
+        setSearchResults(store, searchResults, kind);
+        resolve(searchResults);
+      } else {
+        reject(new Error('Dropping a previous fetch while searching.'));
+      }
     });
+  });
+}
+
+function setSearchResults(store, searchResults, kind) {
+  store.commit('CORE_SET_PAGE_LOADING', true);
+  const { rootNodes } = store.state.topicsRoot;
+  const { channel_ids, results } = searchResults;
+  const nodes = results.map(n => ({
+    ...n,
+    channel: rootNodes.find(c => c.id === n.channel_id),
+  }));
+  const promises = channel_ids.map(id => ChannelResource.fetchModel({ id }));
+  Promise.all(promises).then(collection => {
+    const channels = collection
+      .map(c => ({
+        ...c,
+        thumbnail: getChannelIcon(c),
+        title: c.name,
+        order: channel_ids.indexOf(c.id),
+      }))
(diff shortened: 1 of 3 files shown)


kolibri_explore_plugin/assets/src/views/SearchPage.vue
```diff
@@ -250,6 +256,18 @@
           params: { channel_id: channelId },
         });
       },
+      searchLastChannels(store, query, kind) {
```
dylanmccall, line 259: This all makes sense to me, but the name of this function doesn't work for me. I'd love if it conveyed a bit more what it's doing. Maybe `searchChannelsOnce`?

dylanmccall (approved):
One nitpick, but all's good here :+1:

danigm (changes requested):
This doesn't solve the issue. Once the `searchChannels` is called, the request to the backend will be done outside of this component, and the results are stored in the state with the mutation `SET_SEARCH_RESULT`.

This component uses the `mapState` to keep track of `searchResult`, and that's what is rendered, so even if this promise is canceled here, the `searchResult` in the store will get updated with unwanted data.

Right now, we just fill the `resultKinds` array in the `SearchPage` component, but the real nodes and channels comes from the store state `topicsRoot.searchResult`, so even with this change, the race condition could happen and it's possible to appear content from previous queries.

That's why I removed all the state usage in this PR: https://github.com/endlessm/kolibri-explore-plugin/pull/393/files

kolibri_explore_plugin/assets/src/views/SearchPage.vue
manuq: Yeah `searchChannelsOnce` is much better. I'm not happy with the original `searchChannels` which is actually searching content per kind *in* channels. But I am preventing doing renames at this point.

manuq:
> This doesn't solve the issue. Once the `searchChannels` is called, the request to the backend will be done outside of this component, and the results are stored in the state with the mutation `SET_SEARCH_RESULT`.
> 
> This component uses the `mapState` to keep track of `searchResult`, and that's what is rendered, so even if this promise is canceled here, the `searchResult` in the store will get updated with unwanted data.
> 
> Right now, we just fill the `resultKinds` array in the `SearchPage` component, but the real nodes and channels comes from the store state `topicsRoot.searchResult`, so even with this change, the race condition could happen and it's possible to appear content from previous queries.
> 
> That's why I removed all the state usage in this PR: https://github.com/endlessm/kolibri-explore-plugin/pull/393/files

Good catch. In fact yes there are a number of issues and bad code practices in the SearchPage view, like using the store state and passing `this.$store` to the handler. As I said before, at this point we can't do aggresive changes as in #393. It has to be small enough to be predictable and to ensure that we are not introducing new bugs while fixing one bug. So I will try to adapt this PR to use the store state by moving the "only once" wrapper to the handler. If that doesn't work and no simple-enough ideas appear, then I'm afraid we'll have to release with this bug.

kolibri_explore_plugin/assets/src/modules/topicsRoot/handlers.js
```diff
@@ -181,7 +181,21 @@ export function showFilteredChannels(store) {
   );
 }
 
-export function searchChannels(store, search, kind) {
+export function searchChannelsOnce(store, search, kind) {
+  const searchPromise = searchChannels(store, search, kind);
```
danigm, line 185: This doesn't solve the problem of the `searchChannels` function updating the `topicsRoot/SET_SEARCH_RESULT`. This is the same as before, but done in the `handlers.js` instead of in the `SearchPage` component, but doesn't avoid the update of the store with search results for old searches.

This implementation will mitigate one of the problems in the `SearchPage` component, the one of duplicate rows of the same kind in the results, but won't fix the issue with content from a previous search appearing in the current one, mixed with the last search.

danigm (changes requested):
This just solves the problem partially adding a wrapper function and more data to the state.

I don't think that this is the way to go, it's a partial solution, if we are going to fix this, maybe it's better to apply a full fix.

I think that the main problem in the search process is the use of the vuex store to save the search results in async calls, so the better way to fix it is just don't do that.

The search results are not needed by any other component, so there's no reason to keep that in the store, it could be stored in the `SearchPage` component, and doing that, it's easier to drop old queries results, by storing the promise object (like in this PR) or with the integer index (like in the #393),  but without involving the store state.

kolibri_explore_plugin/assets/src/modules/topicsRoot/handlers.js
```diff
@@ -181,40 +181,53 @@ export function showFilteredChannels(store) {
...
+  });
+  store.commit('topicsRoot/SET_LAST_SEARCH_PROMISE', { kind, searchPromise });
+  return new Promise(resolve => {
+    searchPromise.then(searchResults => {
+      const { lastSearchPromises } = store.state.topicsRoot;
+      if (searchPromise == lastSearchPromises[kind]) {
+        store.commit('topicsRoot/SET_LAST_SEARCH_PROMISE', { kind, searchPromise: null });
+        setSearchResults(store, searchResults, kind);
+        resolve(searchResults);
+      }
```
danigm, line 200: Maybe it could be a good idea to add here an `else` clause with the promise `reject`, to free the resources.

danigm (approved):
Now this is similar to my [first proposal](https://github.com/endlessm/kolibri-explore-plugin/commit/edf52c7ad0c2b69346a2b31e25ff0592aaa63e9c), but storing the promise object instead of an integer index.

I think this is okay to be merged.

kolibri_explore_plugin/assets/src/modules/topicsRoot/handlers.js
manuq: @danigm after some investigation, we are not catching the promise so there is no need for a reject. Since we are not maintaining a reference to previous promises they will be garbage-collected. It's explained in [the article I've followed](https://codeburst.io/cancelling-async-operations-in-vue-js-3d0f3c2de598) and also [here](https://stackoverflow.com/questions/36734900/what-happens-if-you-dont-resolve-or-reject-a-promise#:~:text=A%20promise%20is%20just%20an,just%20a%20regular%20Javascript%20object.).

But I already added the reject. Please let me know if you want it or if I should revert the last change

danigm: Looks good to me.

PR #399, opened 2022-02-24
channel overrides: switch to new Chapter Books channel

erikos:
The new channel does have better quality Ebooks in suitable languages.

https://phabricator.endlessm.com/T33206


manuq (approved):
Code wise looks good, just an ID replacement.

PR #402, opened 2022-03-01
Auto authenticate

danigm:
This PR just adds a middleware and a custom settings file to extend the Kolibri default settings, so this new settings can be used in custom deployments like the Endless Key to auto authenticate the users with:

```
DJANGO_SETTINGS_MODULE = "kolibri_explore_plugin.middleware.AlwaysAuthenticatedMiddleware"
```

This is built on top of this PR: https://github.com/endlessm/kolibri-explore-plugin/pull/401

https://phabricator.endlessm.com/T33233

### DIFF
--- a/kolibri_explore_plugin/settings.py
+++ b/kolibri_explore_plugin/settings.py
@@ -1,0 +1,6 @@
+from kolibri.deployment.default.settings.base import *  # noqa @UnusedWildImport
+
+
+MIDDLEWARE = list(MIDDLEWARE) + [  # noqa F405
+    "kolibri_explore_plugin.middleware.AlwaysAuthenticatedMiddleware"
+]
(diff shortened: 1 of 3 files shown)


kolibri_explore_plugin/settings.py
```diff
@@ -0,0 +1,6 @@
+from kolibri.deployment.default.settings.base import *  # noqa @UnusedWildImport
+
+
+MIDDLEWARE = list(MIDDLEWARE) + [  # noqa F405
+    "kolibri_explore_plugin.middleware.AlwaysAuthenticatedMiddleware"
+]
```
dylanmccall, line 6: Instead of adding this settings file here, do you think we could add it to kolibri-app _itself_ in endlessm/kolibri-app#30? For example, that repo already has a `django_app_settings.py` which is used for the Android app.

PR #416, opened 2022-04-27
Filter out *.map files in custom channel bundles

dylanmccall:
https://phabricator.endlessm.com/T33430


danigm (approved):
Looks good to me

PR #417, opened 2022-04-27
Add an option to load channel apps from elsewhere

dylanmccall:
https://phabricator.endlessm.com/T33411

### DIFF
--- a/kolibri_explore_plugin/views.py
+++ b/kolibri_explore_plugin/views.py
@@ -19,9 +19,10 @@ from kolibri.core.content.zip_wsgi import get_embedded_file
 from kolibri.core.decorators import cache_no_user_data
 from kolibri.utils import conf
 
-APPS_BUNDLE_PATH = conf.OPTIONS["Explore"]["APPS_BUNDLE_PATH"] or os.path.join(
-    os.path.dirname(__file__), "apps"
-)
+APPS_BUNDLE_PATHS = []
+if conf.OPTIONS["Explore"]["APPS_BUNDLE_PATH"]:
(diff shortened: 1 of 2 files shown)


kolibri_explore_plugin/views.py
```diff
@@ -39,12 +44,10 @@ def options(self, request, *args, **kwargs):
         return HttpResponse()
 
     def _get_file(self, app, path):
-        base = os.path.join(os.path.dirname(__file__), "apps")
-
         if path.startswith("/"):
             path = path[1:]
 
-        filename = os.path.join(base, app, path)
+        filename = os.path.join(APPS_BUNDLE_PATH, app, path)
```
danigm, line 50: What do you think about making this a `PATH` resolution, with multiple paths with priority, instead of just a fixed path? For example:

```python
filename = None
for apps_path in APPS_PATHS:
    filename = os.path.join(apps_path, app, path)
    if os.path.exists(filename):
        break
else:
    raise Http404
```

And `APPS_PATHS` could be a composition of the option and the fallback path inside the python code.

danigm (approved):
Looks good to me. Just a design option comment, but feel free to merge as it is if you consider that it doesn't worth to implement that way.

kolibri_explore_plugin/views.py
dylanmccall: Thanks! That makes sense to me. It flows nicer :) It would help us with including a _default_ app in the explore plugin itself. This way we would have a reasonable guarantee that it exists.

manuq (approved):

PR #419, opened 2022-04-28
Filter out *.map files in custom channel bundles

danigm:
To support this, we need to use the zipfile module directly to create
each channel's zip file. In addition, we can use ZIP_LZMA compression
for improved performance. Lzma is supported in Python 3.3 and later.

https://phabricator.endlessm.com/T33430

### DIFF
--- a/scripts/_common.py
+++ b/scripts/_common.py
@@ -1,7 +1,8 @@
 #!/usr/bin/env python
 import os
-import shutil
 import subprocess
+import zipfile
+from pathlib import Path
 
 
 PROJECT_DIR = subprocess.check_output(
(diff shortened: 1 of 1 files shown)


scripts/_common.py
```diff
@@ -31,10 +32,21 @@
     PROJECT_DIR, "kolibri_explore_plugin", "static", "highlighted-content.json"
 )
 
+BUNDLE_EXCLUDE_GLOBS = ["*.map"]
 
-def bundle_zip(zip_name):
-    shutil.make_archive(zip_name, "zip", "dist/")
-    print(f"File ./{zip_name}.zip created.")
+
+def bundle_zip(zip_name, compression=zipfile.ZIP_LZMA):
```
danigm, line 38: This breaks the Explore Plugin in the my chromebook build: `RuntimeError: Compression requires the (missing) lzma module`

dylanmccall: Ah, thanks for testing that :) I updated endlessm/kolibri-installer-android#2 to include `liblzma` (https://github.com/kivy/python-for-android/blob/master/pythonforandroid/recipes/liblzma/__init__.py) in p4a's requirements list, which should solve that issue.

manuq:
I have cherry-picked @dylanmccall commit instead of merging this new PR to respect authoring :)

PR #420, opened 2022-04-29
package.json: Update kolibri-design-system version

manuq:
To fix the build, target the commit from Github for release tag v1.3.0.

https://phabricator.endlessm.com/T33412


dylanmccall:
This makes sense to me! Had a small hiccup from a dependency conflict that appeared on my end, but all's good here :)

PR #422, opened 2022-05-31
Import EK channels from USB if present

danigm:
This is a first step to import the content from USB if present instead
of from network, without importing the content, just the metadata.

The list of channels is getted from the network right now, waiting for
the metadata.json definition that could be present in the
KOLIBRI_DATA/content folder:
https://github.com/learningequality/kolibri/pull/9460

https://phabricator.endlessm.com/T33526

### DIFF
--- a/kolibri_explore_plugin/views.py
+++ b/kolibri_explore_plugin/views.py
@@ -6,6 +6,7 @@ import json
 import os
 
 import requests
+from django.core.management import call_command
 from django.http import FileResponse
 from django.http import Http404
 from django.http import HttpResponse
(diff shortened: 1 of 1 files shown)


kolibri_explore_plugin/views.py
```diff
@@ -144,10 +177,37 @@ def get(self, request):
             json.dumps(jobs_response), content_type="application/json"
         )
 
+    def import_job(self, task):
+        content_fallback = paths.get_content_fallback_paths()
+        if content_fallback:
+            task.update({"type": "DISKIMPORT"})
+            # Just getting the first one
+            directory = os.path.dirname(content_fallback[0])
```
manuq, line 185: Yes, getting the first one would work. At least in the Android app the content fallback path will be set only if there is an external device with a KOLIBRI_DATA folder and the corresponding folder structure.

kolibri_explore_plugin/views.py
```diff
@@ -109,6 +113,35 @@ def get(self, request, app):
             return HttpResponse(json_file, content_type="application/json")
 
 
+def _import_channel_from_usb(channel_id, directory):
```
manuq, line 116: Function name nitpick: this could be any fallback directory so I would change "usb" here.
```suggestion
def _import_channel_from_directory(channel_id, directory):
```

manuq (changes requested):
The change looks good. But getting the channels from the network goes against the goal of using an USB for offline content.

I would suggest either:
- read the channels from the sqlite database in `KOLIBRI_DATA/preseeded_kolibri_home`
- or wait for the manifest file in https://github.com/learningequality/kolibri/pull/9460, as copying the preseeded database file directly works offline for both Windows and Android.

PR #427, opened 2022-06-10
Remove map files from deployed loadingScreen

manuq:
The map files add some megabytes that aren't needed in production.

https://phabricator.endlessm.com/T33515


dylanmccall (approved):
This makes sense to me! I think we already do this elsewhere :)

PR #432, opened 2022-06-21
T33613 revert

manuq:
https://phabricator.endlessm.com/T33613


danigm (approved):
Looks good to me. We can remove this functionality for now so we can keep the plugin more stable

PR #437, opened 2022-06-30
Detect available space in device

danigm:
Disable collections that are bigger than the space available so the user
can't try to install a collection if there's no enought space.

https://phabricator.endlessm.com/T33572

### DIFF
--- a/packages/eos-components/src/components/CollectionSelectionModal.vue
+++ b/packages/eos-components/src/components/CollectionSelectionModal.vue
@@ -105,10 +105,6 @@
     border-radius: $border-radius-lg;
   }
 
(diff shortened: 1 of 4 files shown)


packages/eos-components/src/components/CollectionSelectionModal.vue
```diff
@@ -114,6 +105,10 @@
     border-radius: $border-radius-lg;
   }
 
+  .card.disabled {
+    opacity: 0.7;
```
manuq, line 109: Nit this should be in the bootstrap theme. Not hardcoded.

manuq (approved):
Looks great!

PR #439, opened 2022-07-01
Collection selection redesign

danigm:
https://phabricator.endlessm.com/T33631


packages/eos-components/src/components/GradeSelectionModal.vue
```diff
@@ -0,0 +1,134 @@
...
+        default: false,
+      },
+    },
+    data() {
+      return {
+        selection: null, // 'primary', 'intermediate', 'secondary'
+        grades: [
+          {
+            key: 'primary',
+            text: 'for students ages 6-9<br/>(in grades K-3)',
```
manuq, line 91: This text will be kept hardcoded, right? Please avoid passing HTML directly to the template. Use another variable for the second line. And in any case please ask Joana if this is really intended to be a single line. Usually designers break lines in their blueprints for aesthetic reasons but this goes against responsiveness.

packages/eos-components/src/components/GradeSelectionModal.vue
```diff
@@ -0,0 +1,134 @@
...
+
+        <b-card-group deck class="py-5">
+          <b-card
+            v-for="grade in grades"
+            :key="grade.key"
+            :title="grade.key"
+            :borderVariant="selection === grade.key ? 'primary' : 'default'"
+            titleTag="h6"
+            subTitleTag="h4"
+            class="welcome-card"
```
manuq, line 32: Since there are no other cards in this component you can just use the `.card` CSS class for referencing it. Thus this line is not needed.

packages/eos-components/src/components/GradeSelectionModal.vue
```diff
@@ -0,0 +1,134 @@
...
+
+  @import '../styles';
+
+  ::v-deep .modal-content {
+    color: black;
+    text-align: center;
+    background-color: white;
+  }
+
+  .welcome-card .card-title {
```
manuq, line 130: The CSS is scoped so you can just do:
```
  .card .card-title {
```

manuq (changes requested):
Looking great, only small changes needed.

packages/eos-components/src/components/GradeSelectionModal.vue
danigm: The `.welcome-card` styles are defined in [the styles](https://github.com/endlessm/kolibri-explore-plugin/pull/439/files#diff-3d1fd9dd539be4bf82ce6e5adb1a46c94d63b83ba763f2f5e5c71c9e260e67d1R98) that's why this is used.

packages/eos-components/src/index.scss
```diff
@@ -95,6 +95,22 @@
   white-space: pre-wrap;
 }
 
-.card.disabled {
-  opacity: 0.7;
+.welcome-card {
```
manuq, line 98: After a second thought, I think all this CSS should go to a WelcomeCard component. See all other styles in this `index.scss` files, they are not for a custom component in particular: they affect all buttons, all breadcrumbs, etc. If this were CSS for all cards then it should go here. But it's for a particular kind of cards.

manuq (changes requested):

packages/eos-components/src/index.scss
danigm: Yes, that's a better solution, we can implement the `WelcomeCard` to share the same css in a component instead of a class.

danigm: I'll create the `WelcomeCard` component to use here.

packages/eos-components/src/styles.scss
```diff
@@ -192,3 +192,5 @@ $navbar-height: $navbar-brand-height +
 
 // Bigger than 3 spacer:
 $big-spacer: $spacer * 3.5;
+
+$disabled-card-opacity: 0.7;
```
manuq, line 196: :+1:

packages/eos-components/src/components/WelcomeCard.vue
```diff
@@ -0,0 +1,105 @@
...
+    :borderVariant="selected ? 'primary' : 'default'"
+    :class="{ active: selected, disabled: disabled }"
+    titleTag="h6"
+    subTitleTag="h4"
+    imgBottom
+    :imgSrc="img"
+  >
+    <!-- eslint-disable vue/no-v-html -->
+    <p class="text-muted">
+      {{ text }} <br v-if="secondaryText"> {{ secondaryText }}
```
manuq, line 14: I think we should figure this out with Joana. Have you asked her?

manuq (approved):

packages/eos-components/src/components/WelcomeCard.vue
danigm: I didn't ask, It was clear for me that it's two text blocks, but in any case, we can ask and the change will be easy to apply.

manuq: Yes, we are hardcoding right now but it would be great to move this metadata to the manifest file. So if the manifest file format needs a title, subtitle and yet another sub-subtitle it would be great to know in advance. And please don't use HTML in those fields.

PR #443, opened 2022-07-08
Use _read_pid_file instead of get_status

danigm:
This patch uses the more simple function _read_pid_file instead of the
get_status to get the process PID so it does less operations and is
faster.

https://phabricator.endlessm.com/T33633


dylanmccall:
LGTM :+1+

PR #444, opened 2022-07-12
components: Shows the correct OS name in the banner by detecting it

starnight:
https://phabricator.endlessm.com/T33675


danigm (commented):
Let's just remove the platform specific text to simplify a bit. I think it's not too relevant to the text.

PR #447, opened 2022-07-13
Fix the size units

starnight:
The download size unit should be xx bytes. The bytes can be simplified
as capitalized 'B'. The lowercase 'b' represents bit.

https://phabricator.endlessm.com/T33685

### DIFF
--- a/kolibri_explore_plugin/views.py
+++ b/kolibri_explore_plugin/views.py
@@ -114,7 +114,7 @@ class AppMetadataView(AppBase):
 class EndlessLearningCollection(View):
     COLLECTIONS = {
         "small": {
-            "title": "3Gb",
+            "title": "3 GB",
             "subtitle": "Small",
             "channels": 10,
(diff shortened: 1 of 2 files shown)


kolibri_explore_plugin/views.py
```diff
@@ -114,7 +114,7 @@ def get(self, request, app):
 class EndlessLearningCollection(View):
     COLLECTIONS = {
         "small": {
-            "title": "3Gb",
+            "title": "3GB",
```
wjt, line 117: I hate to be a pedant but there should be a space between the number and the unit. If we are feeling really pedantic it should be a nonbreaking space or similar.

starnight: Knowledge learned.
I have not known the rule until I found https://english.stackexchange.com/questions/2794/punctuation-with-units

danigm (approved):
LGTM

PR #449, opened 2022-07-15
Change channel ID for TED Ed

hydrolet:
https://phabricator.endlessm.com/T33691


erikos (changes requested):
I think the patch is upside down :) The patch is trying to make the Discovery page pointing now to the wrong TED-Ed channel. We should however fix the manifests that we have created and point to the right TED-Ed channel.

Those have to be fixed and point to 2091ca47ff544c96b4ae02b3a92346e1

manuq:
Any follow-ups on this @erikos ? This PR was left around.

manuq:
No news so closing.

PR #456, opened 2022-08-12
Welcome screen: Better state management

manuq:
By implementing a simple store without vuex. This removes the
need to override setters in the components. Now the host app
(Android/Windows) can still call APIs like:
`window.setHasUSB(true)` and this is reactive to all the
welcome screen, not just a single component.

### DIFF
--- a/packages/loading-screen/src/App.vue
+++ b/packages/loading-screen/src/App.vue
@@ -8,6 +8,7 @@
 </template>
 
 <script>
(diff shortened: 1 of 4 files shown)


manuq:
@danigm what do you think? Is this a better approach to state management?

danigm:
> @danigm what do you think? Is this a better approach to state management?

Yes, it looks good to me!

PR #458, opened 2022-08-25
wording: Remove "the" Endless Key

manuq:
Except when referred to something like: "the Endless Key initiative"

https://phabricator.endlessm.com/T33672


danigm (approved):
LGTM

PR #460, opened 2022-08-30
Calculate the number of channels per collection

danigm:
Instead of use the hardcoded value, this patch makes the number of
channels dynamic and it's calculated in real time using the manifest
file.

https://phabricator.endlessm.com/T33744

This change depends on the collections metadata change: https://github.com/endlessm/endless-key-collections/pull/3


kolibri_explore_plugin/views.py
```diff
@@ -183,11 +183,20 @@ class EndlessLearningCollection(View):
...
 
-    def check_collection_availability(self):
+    def update_collection_json(self):
         free_space_gb = get_free_space() / 1024**3
-        for collections in self.grade_collections.values():
-            for v in collections.values():
+        for grade, collections in self.grade_collections.items():
+            for collection, v in collections.items():
+                # check collection availability
                 v["available"] = v["size"] < free_space_gb
```
manuq, line 191: What is "v" in this context? Please use a meaningful variable name.

kolibri_explore_plugin/views.py
```diff
@@ -183,11 +183,20 @@ class EndlessLearningCollection(View):
...
+                # calculate real number of channels
+                collection_manifest = os.path.join(
+                    COLLECTION_PATHS, f"{grade}-{collection}.json"
+                )
+                with open(collection_manifest) as f:
+                    manifest = json.load(f)
+                    channels = manifest.get("channels", [])
+                    v["channels"] = len(channels)
```
manuq, line 199: :+1:

manuq (changes requested):
Great, this PR removes the hardcoded availability and channels count number. But it doesn't remove the hardcoded values in https://github.com/endlessm/kolibri-explore-plugin/blob/325312d59890eb8a9cf8ade8fb8e6151bea1ec80/kolibri_explore_plugin/views.py#L126-L132. Why is that? Also there are more hardcoded fields. The only hardcoded thing should be the hierarchy of grade / collection that's part of the UI logic. Everything else like title, subtitle, can be in the JSON.

I guess this is to display the cards early. But the cards should not be displayed until the JSON information is fetched. Or better, they should be displayed in a loading card state.

Also a nit: I was checking the code yesterday and saw that the GradeSelectionModal and ContentSelectionModal are in eos-components. These are not reusable components. Can you move them out of the eos-components lib?

kolibri_explore_plugin/views.py
danigm: I've added a new commit using better variable names.

danigm:
> . Why is that? Also there are more hardcoded fields. The only hardcoded thing should be the hierarchy of grade / collection that's part of the UI logic. Everything else like title, subtitle, can be in the JSON.
>
> I guess this is to display the cards early. But the cards should not be displayed until the JSON information is fetched. Or better, they should be displayed in a loading card state.

This information is not currently in the JSON, but we can move to there to make it more consistent and easy to maintain.

manuq (changes requested):

kolibri_explore_plugin/assets/src/components/CollectionSelectionModal.vue
```diff
@@ -35,7 +35,7 @@
               variant="primary"
               :disabled="!collection.available"
             >
-              Download <strong>{{ collection.size }} GB</strong>
+              Download <strong>{{ collection.required_gigabytes }} GB</strong>
```
danigm, line 38: This should be `collection.metadata.required_gigabytes`.

PR #463, opened 2022-09-06
About versioning

danigm:
This PR adds more information to the `build-info.json` fiile and a new script, distributed with the package to update the `build-info.json` file during the specific platform app build.

https://phabricator.endlessm.com/T33828


scripts/build_info.sh
```diff
@@ -6,6 +6,9 @@ LAST_RELEASE=$(git describe --tags --abbrev=0)
 COMMITS=$(git log --no-merges --pretty=format:'%h|%s|%aN' $LAST_RELEASE..HEAD)
 
 echo '{
+    "platform": "online",
+    "platform_version": "1.0.0",
+    "kolibri": "0.15",
```
manuq, line 11: -1 from my side to maintain more hardcoded numbers and more handcraft work on each release.

manuq (changes requested):

scripts/build_info.sh
danigm: These are the default values. This PR also have a python script to update this automatically during each platform app build.

manuq: OK, anyways I don't think the plugin should know anything about which platform is going to use it. We don't want to have separate wheel files for each platform just for this info. Also the ticket is not about displaying the EK version in the UI. Each OS has a consistent way to check the app version in their app stores.

danigm: I've updated the PR to just add the Kolibri version and calculate it from the existing javascript object.

manuq (approved):

PR #464, opened 2022-09-06
Rebase kolibri 016

manuq:
https://phabricator.endlessm.com/T33793

### DIFF
--- a/kolibri_explore_plugin/views.py
+++ b/kolibri_explore_plugin/views.py
@@ -16,7 +16,7 @@ from django.views.decorators.clickjacking import xframe_options_exempt
 from django.views.decorators.csrf import csrf_exempt
 from django.views.generic.base import TemplateView
 from django.views.generic.base import View
-from kolibri.core.content.api import cache_forever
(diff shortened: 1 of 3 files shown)


kolibri_explore_plugin/views.py
```diff
@@ -77,7 +77,7 @@ def _get_file(self, app, path):
 
 
 class AppView(AppBase):
-    @metadata_cache
+    @method_decorator(metadata_cache, name="dispatch")
```
dbnicholson, line 80: I think you should squash this with the original `cache_forever` change.

dbnicholson (changes requested):
This seems OK to me, but I didn't really look at the Kolibri API changes in detail. The plan is to release this as 3.x, right?

kolibri_explore_plugin/views.py
manuq: Good point!

manuq:
> 

Yes, this will go to v3. Merging.

PR #470, opened 2022-10-18
Reimplement download of collections - backend part

manuq:
This is the backend part of the reimplementation.

### DIFF
--- a/kolibri_explore_plugin/collectionviews.py
+++ b/kolibri_explore_plugin/collectionviews.py
@@ -49,20 +49,23 @@ GRADES_METADATA = {
 
 
 class EndlessKeyContentManifest(ContentManifest):
-    def __init__(self, grade, name):
-        """The EK collections are organized by grade. Example: the
+    def __init__(self):
+        """Extended content manifest for Endless Key
+
+        The EK collections are organized by grade. Example: the
         "primary-small" collection has grade "primary" and name
         "small"
+
+        They also add more metadata and set availability according to
+        disk space.
         """
-        self.grade = grade
-        self.name = name
         self.metadata = None
         self.available = None
         super().__init__()
 
-    def read(self, validate=False):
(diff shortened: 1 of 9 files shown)


kolibri_explore_plugin/collectionviews.py
```diff
@@ -0,0 +1,621 @@
...
+
+
+def _remotechannelimport(user, channel_id):
+    """Create, validate and enqueue a remotechannelimport job."""
+    job = remotechannelimport.validate_job_data(
+        user,
+        {
+            "channel_id": channel_id,
+            # FIXME: Why is channel_name needed? It isn't known at this point.
+            "channel_name": "foo",
```
dylanmccall, line 420: That happened in this pull request:

https://github.com/learningequality/kolibri/pull/9438/commits/b53d7baa29374b4dbfb467573b664c9175bfb958#diff-0ebb74f1619d471f9ba176a2414e427e279a76fac0278f71fda6414a441ab0b5L109-L111

Looks like it used to set a default, but it was unused so it got removed. That old default looked like this:

```
def get_channel_name(channel_id, require_channel=False):
    try:
        channel = ChannelMetadata.objects.get(id=channel_id)
        channel_name = channel.name
    except ChannelMetadata.DoesNotExist:
        if require_channel:
            raise serializers.ValidationError("This channel does not exist")
        channel_name = ""

    return channel_name
```

Does the job validator accept an empty string for channel_name? That would be the same as the behaviour with Kolibri 0.15.

manuq: > Does the job validator accept an empty string for channel_name? That would be the same as the behaviour with Kolibri 0.15.

Nope, if I pass an empty string the validation fails with a "channel name can't be empty" error.

dylanmccall (approved):
👍️

PR #471, opened 2022-10-19
Reimplement download of collections - frontend part

manuq:

### DIFF
--- a/kolibri_explore_plugin/assets/src/components/InstallContentModal.vue
+++ b/kolibri_explore_plugin/assets/src/components/InstallContentModal.vue
@@ -15,7 +15,7 @@
         Downloading&hellip;
       </h1>
       <h5 class="text-muted">
-        You may close this window and explore content while
+        Please wait while
         the collection is downloading.
       </h5>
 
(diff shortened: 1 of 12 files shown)


kolibri_explore_plugin/assets/src/components/InstallContentModal.vue
```diff
@@ -73,156 +53,111 @@
...
     },
     computed: {
-      progress() {
-        const jobs = this.jobs.map(j => {
-          if (j.percentage > 1 || !j.channel_name) {
-            return 0;
+      titleLabel() {
+        if (this.status !== null) {
+          if (this.status.stage === 'IMPORTING_CONTENT') {
+            return 'Downloading…';
```
dylanmccall, line 74: There are a lot of strings here, so it would be nice if they were going through the localization system so we don't need to find them all later :) [`ContentUnavailablePage`](https://github.com/endlessm/kolibri-explore-plugin/blob/master/kolibri_explore_plugin/assets/src/views/ContentUnavailablePage.vue) has a nice straightforward example of all we need there for the moment.

```suggestion
            return this.$tr('titleLabelDownloading');
```

dylanmccall (changes requested):
I'm pretty happy with this! I left one small nitpick about translations, but I realize we haven't really worried about them here historically, so if it's too much trouble at the moment then feel free to merge :)

kolibri_explore_plugin/assets/src/components/InstallContentModal.vue
manuq: Excellent. Yes I'm happy to make this a new policy for upcoming frontend changes. All labels should use the translation system.

PR #472, opened 2022-10-22
Fix channel iframe loading before API is available

dylanmccall:
https://phabricator.endlessm.com/T33948


manuq (approved):
Thanks for figuring this out! It is a rather safe change to merge.

PR #474, opened 2022-11-08
T33835 download info logs

manuq:

### DIFF
--- a/kolibri_explore_plugin/collectionviews.py
+++ b/kolibri_explore_plugin/collectionviews.py
@@ -11,6 +11,7 @@ from kolibri.core.content.utils.content_manifest import ContentManifest
 from kolibri.core.content.utils.content_manifest import (
     ContentManifestParseError,
 )
+from kolibri.core.tasks.constants import DEFAULT_QUEUE
 from kolibri.core.tasks.job import Priority
 from kolibri.core.tasks.job import State as JobState
 from kolibri.core.tasks.main import job_storage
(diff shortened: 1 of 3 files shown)


kolibri_explore_plugin/collectionviews.py
```diff
@@ -430,7 +436,7 @@ def _remotechannelimport(user, channel_id):
             "channel_name": "foo",
         },
     )
-    job_id = remotechannelimport.enqueue(job=job)
+    job_id = remotechannelimport.enqueue(job=job, priority=Priority.HIGH)
```
dylanmccall, line 439: This change shouldn't be needed: the `remotechannelimport` task has priority=Priority.HIGH by default.

```suggestion
    job_id = remotechannelimport.enqueue(job=job)
```

kolibri_explore_plugin/collectionviews.py
```diff
@@ -446,7 +452,7 @@ def _remotecontentimport(user, channel_id, node_ids, exclude_node_ids):
             "exclude_node_ids": exclude_node_ids,
         },
     )
-    job_id = remotecontentimport.enqueue(job=job)
+    job_id = remotecontentimport.enqueue(job=job, priority=Priority.HIGH)
```
dylanmccall, line 455: This one … should work and be necessary, I think, but it appears to be a no-op:

https://github.com/learningequality/kolibri/blob/develop/kolibri/core/tasks/registry.py#L262-L273

It ends up using the priority from `register_task`, ignoring the keyword argument we pass. It doesn't look like anything in Kolibri expects this to work the way we are expecting here, so I think we need to either hack at a lower level in the tasks registry code (see the linked function) or add our own task for this instead.

dylanmccall (changes requested):

kolibri_explore_plugin/collectionviews.py
manuq: OK, it seems you found a way to make it work by passing the default queue. Thanks! Merging.

PR #482, opened 2022-12-19
T34157 version

manuq:
https://phabricator.endlessm.com/T34157


scripts/bump_version.py
```diff
@@ -0,0 +1,44 @@
...
+    if len(sys.argv) == 3:
+        assert sys.argv[1] == "major", usage_message
+        _update_version_name(sys.argv[2])
+        # Remove version name to let bump2version work:
+        del sys.argv[2]
+
+    elif len(sys.argv) == 2:
+        assert sys.argv[1] == "minor", usage_message
+    else:
+        raise AssertionError(usage_message)
```
dylanmccall, line 39: This is a clever solution. I like it :)

dylanmccall (approved):

scripts/bump_version.py
manuq: I started adding a argparse like in other scripts and realized that it was too much for this case since bump2version already has one. Maybe in the future we can improve releases with github actions?

PR #486, opened 2022-12-27
Backend: Add models for externally tagged content

manuq:
https://phabricator.endlessm.com/T34276


kolibri_explore_plugin/models.py
```diff
@@ -0,0 +1,25 @@
...
+        ContentNode, on_delete=models.CASCADE, primary_key=True
+    )
+    tags = models.ManyToManyField(
+        ExternalContentTag,
+        symmetrical=False,
+        related_name="tagged_content",
+    )
+
+    def __str__(self):
+        return self.content_node.title
```
dylanmccall, line 25: I'm wondering if you can explain why we're using two models here, as opposed to a `ForeignKey` in `ExternalContentTag` to `ContentNode`. As it is, this turns into a many-to-one relationship anyway, albeit with an extra table for the ManyToManyField. Is there something in particular this approach gives us?

manuq: We need a many-to-many relationship: a content node can have many tags, and a single tag can be in many content nodes.

Initially I thought about having a one-to-one intermediary model to associate more things we want downstream to a ContentNode. Like the structured metadata that are now overusing the upstream ContentTag with a `foo=bar` convention and being processed by the frontend.

But you are right! This can work with less tables and a single model:
```
class ExternalContentTag(models.Model):
    id = UUIDField(primary_key=True)
    tag_name = models.CharField(max_length=100, unique=True)
    content_nodes = models.ManyToManyField(
        ContentNode,
        symmetrical=False,
        related_name="tagged_content",
    )
```

I will give it a try and adapt the code to a single model.

manuq: @dylanmccall based on this I have provided an alternative PR using a single model: https://github.com/endlessm/kolibri-explore-plugin/pull/487

More details in the phabricator ticket!

PR #487, opened 2023-01-03
T34276 external tags alt

manuq:
https://phabricator.endlessm.com/T34276


dylanmccall:
Oh gosh, of course, didn't occur to me of course we would want the same tag to apply to more than one node 😊

The semantics of the original make more sense to me, then, but this _is_ a bit more lean. The [through table](https://docs.djangoproject.com/en/1.11/ref/models/fields/#django.db.models.ManyToManyField.through) this generates is effectively the same as the original TaggedContent model, and I think Django should attach some helpful reverse references to ContentNode itself this way. At any rate, I'm happy either way here - having the tag name unique makes it much clearer why it is the way it is :)

dylanmccall:
I went ahead and merged #486, so we can close this one.

PR #488, opened 2023-01-03
T34285 discovery collections

manuq:
https://phabricator.endlessm.com/T34285

### DIFF
--- a/kolibri_explore_plugin/assets/src/views/DiscoveryPage.vue
+++ b/kolibri_explore_plugin/assets/src/views/DiscoveryPage.vue
@@ -1,103 +1,30 @@
 <template>
 
-  <div class="channels-page d-flex flex-column min-vh-100">
-    <DiscoveryNavBar />
-
-    <b-container class="mb-2 mt-4">
-      <h5 class="mt-2 text-muted">
-        Content Suggestions
-      </h5>
-    </b-container>
-
-    <template v-if="loading || downloading">
-      <CarouselPlaceholder />
-    </template>
-
-    <template v-else>
-      <b-container class="mb-2 mt-2">
-        <ButtonsBar
-          class="mr-3 mt-1"
-          title="More Topics"
-          :buttons="Array.from(searchTerms.keys())"
-          @click="goToTerm"
-        />
-      </b-container>
-      <Carousel class="pt-3" :nodes="carouselNodes" :showChannelIcon="true" />
-    </template>
-
-    <div class="flex-fill main">
-
-      <template v-if="loading || downloading">
-        <CardGridPlaceholder />
-      </template>
-      <template v-else>
-        <CategorizedList />
-      </template>
-
-    </div>
-
-  </div>
+  <component :is="pageVariant" />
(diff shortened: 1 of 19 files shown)


kolibri_explore_plugin/viewsets.py
```diff
@@ -0,0 +1,28 @@
...
+class ContentNodeExtrasViewset(ContentNodeViewset):
+    @action(detail=False)
+    def by_external_tag(self, request):
+        tag = self.request.query_params.get("tag", None)
+        queryset = self.filter_queryset(self.get_queryset())
+        matching_extras = ContentNodeExtras.objects.filter(tags__tag_name=tag)
+        content_ids = [extra.content_node.pk for extra in matching_extras]
+        queryset = queryset & ContentNode.objects.filter_by_uuids(
+            content_ids, validate=False
+        )
```
dylanmccall, line 27: `ContentNode.objects.filter_by_uuids` accepts a QuerySet as a parameter, which would be more efficient than building a list here.

```suggestion
        queryset = queryset & ContentNode.objects.filter_by_uuids(
            matching_extras.values_list('content_node'), validate=False
        )
```

If we're feeling ambitious, we can even skip that `values_list` part because ContentNodeExtras is keyed by the content_node field:

```suggestion
        queryset = queryset & ContentNode.objects.filter_by_uuids(
            matching_extras, validate=False
        )
```

kolibri_explore_plugin/assets/src/views/DiscoveryPage.vue
```diff
@@ -15,8 +16,8 @@
     components: { DiscoveryPageCollections, DiscoveryPageFullContent },
     computed: {
       showCollectionsPage() {
-        // FIXME read this from a setting.
-        return false;
+        // FIXME return false if "use EK USB" was selected in the onboarding.
```
dylanmccall, line 19: Either way we should create a Phabricator task or two so there's a roadmap to converging these things, but how about in the Android app we skip that onboarding step when `newCollectionsPageFeature` is enabled? That way we can call this bit in the explore plugin _done_. The way I see it is eventually we'll want to have a similar experience either way, mostly for our own sanity. (So, the app can ship with metadata based on the content we know is in Endless Key USBs but then disable the content management UI).

dylanmccall (changes requested):
Okay, I went through this and it looks good! Sorry for the delay. One performance thing. Lots of nice little enhancements here :)

kolibri_explore_plugin/viewsets.py
manuq: Ah that's much better! Indeed keeping things in querysets is the proper way of querying the database. I copied the code from another plugin which probably needs this improvement too. I'll try to find it!

kolibri_explore_plugin/assets/src/views/DiscoveryPage.vue
manuq: I'll answer this in Phabricator as it involves design specs!

PR #490, opened 2023-02-06
Bump yarn to latest

manuq:
To avoid the "yarn is out of date" warning.


dylanmccall:
Oh yeah, that's a good idea :)

PR #495, opened 2023-02-13
T34310 app hook

manuq:
https://phabricator.endlessm.com/T34380


packages/welcome-screen/src/views/UsbDriveConnection.vue
```diff
@@ -11,7 +11,7 @@
         class="mb-4 mt-5"
         variant="primary"
         :disabled="!hasUSB"
-        @click="loadWithUSB"
+        @click="onClick"
```
dylanmccall, line 14: I have the same nitpick here as with `PermissionsWrongFolder.vue`

packages/welcome-screen/src/views/PermissionsWrongFolder.vue
```diff
@@ -12,7 +12,7 @@
       </h5>
       <b-button
         variant="primary"
-        @click="loadWithUSB"
+        @click="onClick"
```
dylanmccall, line 15: It would be nice if this click handler had a more specific name, since there are multiple components in this tree which could be clicked. Perhaps `onNextButtonClick`?

packages/welcome-screen/src/views/PackSelection.vue
```diff
@@ -16,18 +16,31 @@
...
         </b-row>
       </b-container>
+      <p class="mb-5">
+        <UsbIcon />
+        <SdIcon />
+        If you have an Endless Key USB device or SD card,
+        <b-link to="endless-key">
+          click here
+        </b-link>
+        to get your pack!
```
dylanmccall, line 26: I don't have the design handy to compare, but this sort of `[text] <CLICK HERE> [text]` pattern can be problematic for users with screen readers or other accessibility aids. In particular, a screen reader is going to just call out "click here" when that link is selected unless we give it a title attribute, and either way it makes localization tricky. It would help if all of "click here to get your pack" was a link.

Granted, it's a small nitpick since we aren't targeting that use case at the moment, so if it turns into more work to make it look nice, it's fine for now as it is :)

packages/welcome-screen/src/App.vue
```diff
@@ -11,78 +11,46 @@
...
-    window.firstLaunch = () => {
-      this.firstLaunch = true;
-      this.$router.push('/welcome');
-    }
-  },
-  beforeDestroy() {
-    function fallback() {
-      console.log('No welcome screen');
+    window.WelcomeApp = {
+      showLoadingRetry: () => {
```
dylanmccall, line 19: Did you explore using [provide / inject](https://vuejs.org/guide/components/provide-inject.html) here, instead of putting an object in `window`? I'm curious if that would work in a more Vue-ish way :)

dylanmccall (changes requested):
A couple of small nitpicks, but nothing big. This looks like a nice set of changes! :)

packages/welcome-screen/src/views/PermissionsWrongFolder.vue
manuq: You can see it was like that before: the handler was called "loadWithUSB". But this is the only click handler in this component so I think because of scoping is fine to call it just `onClick`.

packages/welcome-screen/src/views/PackSelection.vue
manuq: Totally agree! Also I think "press" is a better verb than "click" because it works for both desktop users and mobile/touchscreen users. I will change it in code and leave a comment in the design specs (tagging you).

packages/welcome-screen/src/App.vue
manuq: We are putting an object in `window` because it needs to be called from the Android/Windows apps.

From Android:
```
mWebView.evaluateJavascript("WelcomeApp.showWelcome()", null);
```

From Windows:
```
mainWindow.webContents.executeJavaScript('WelcomeApp.showWelcome()', true);
```

(See PRs https://github.com/endlessm/kolibri-installer-android/pull/98 and https://github.com/endlessm/endless-key-app/pull/96 which should be reviewed along this one as mentioned in the ticket https://phabricator.endlessm.com/T34380#966384)

I went through the provide/inject docs and I didn't see how it can replace this.

dylanmccall (approved):
Thanks @manuq, that explained everything I was wondering about!

PR #502, opened 2023-03-01
Content packs view: Fix setting loadingNodes

manuq:
This was a leftover line inside the map function. The loadingNodes variable should be set to false after all the collections are fetched. As it's already happening in the then() clause.

https://phabricator.endlessm.com/T34521


dylanmccall (approved):
Okay, makes sense to me.

PR #504, opened 2023-03-01
T34538 keep lights on

manuq:
https://phabricator.endlessm.com/T34538

### DIFF
--- a/kolibri_explore_plugin/assets/src/modules/topicsRoot/handlers.js
+++ b/kolibri_explore_plugin/assets/src/modules/topicsRoot/handlers.js
@@ -3,7 +3,7 @@ import { utils } from 'eos-components';
 import { ChannelResource, ContentNodeResource, ContentNodeSearchResource } from 'kolibri.resources';
 import { getContentNodeThumbnail } from 'kolibri.utils.contentNode';
 
-import { CarouselItemsLength, PageNames, SEARCH_MAX_RESULTS } from '../../constants';
+import { CarouselItemsLength, SEARCH_MAX_RESULTS } from '../../constants';
 import { CustomChannelApps, getBigThumbnail, getChannelIcon } from '../../customApps';
 import { _collectionState } from '../coreExplore/utils';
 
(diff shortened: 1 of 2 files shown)


kolibri_explore_plugin/assets/src/modules/coreExplore/actions.js
```diff
@@ -1,22 +1,18 @@
...
-    channels => {
-      return channels;
-    },
-    error => {
-      store.dispatch('handleApiError', error);
-      return error;
-    }
-  );
+export function setAndCheckChannels() {
+  return get(channels);
```
dylanmccall, line 15: In the [upstream function by the same name](https://github.com/learningequality/kolibri/blob/develop/kolibri/plugins/learn/assets/src/modules/coreLearn/actions.js#L15-L20), Kolibri is still redirecting to an error page if `!get(channels).length`. For reference, this function was originally the same as that code [from kolibri-v0.14.x](https://github.com/learningequality/kolibri/blob/release-v0.14.x/kolibri/plugins/learn/assets/src/modules/coreLearn/actions.js#L15-L28).

I'm wondering if we still want that behaviour. If we _don't_, maybe we should remove this function altogether? It, uh, isn't doing either of the things in its name :b Here's an example upstream where Kolibri is using `useChannels` directly inside a handlers.js, for an idea what the conventions are: https://github.com/learningequality/kolibri/blob/develop/kolibri/plugins/learn/assets/src/modules/recommended/handlers.js#LL10.

dylanmccall (changes requested):
One small thing. Nice assortment of cleanups! :)

kolibri_explore_plugin/assets/src/modules/coreExplore/actions.js
manuq: Yes, I copied from upstream but I intentionally removed that `if` for now. Because our TOPICS_ROOT (aka landing page) may start the collection download if there aren't any channels. But I think that will be useful when we move the collection download to a separate page:
- `if !get(channels).length` and an starter pack was passed in config.ini: redirect to download page.
- else: redirect to CONTENT_UNAVAILABLE page.

PR #506, opened 2023-03-02
Build: Remove collections from build

manuq:
https://phabricator.endlessm.com/T34548


dylanmccall (approved):
:+1: - this manifest file is getting a bit out of hand, but that's for another day :)

manuq:
> +1 - this manifest file is getting a bit out of hand, but that's for another day :)

Yeah.. this is getting worse and worse.

PR #509, opened 2023-03-10
T34593 onboarding pass

manuq:
https://phabricator.endlessm.com/T34593


packages/eos-components/src/components/PackCard.vue
```diff
@@ -50,12 +44,20 @@
...
       onClick() {
         this.$emit('click');
       },
-      onChoosePack() {
-        this.$emit('choosePack');
+      onMouseOver() {
+        this.isHovered = true;
+      },
+      onMouseLeave() {
+        this.isHovered = false;
```
dylanmccall, line 60: I'm unclear why we're doing this _this_ way, so it could use a quick explanation in the comments. It kind of looks like we could use [v-b-hover](https://bootstrap-vue.org/docs/directives/hover) to save some boilerplate, or even add `.card:hover` in our CSS?

dylanmccall (approved):
This looks good to me! One small bit I wasn't sure about, but feel free to merge as needed :)

packages/eos-components/src/components/PackCard.vue
manuq: Oh right! I was copying code from another of our cards which do the same for adding these Bootstrap CSS classes. There is no other way of adding classes dynamically on the template side. But I went to the bootstrap code and found what these classes do:

```
.card {
  box-shadow: $box-shadow-sm !important;
  &:hover {
    box-shadow: $box-shadow !important;
  }
}
```

PR #510, opened 2023-03-14
Channels presentation: Add overrides for SciGirls

manuq:
Two channels point to the same theme: SciGirls English and SciGirls Spanish.

Option: display last topic as bundle.

https://phabricator.endlessm.com/T34204


manuq:
I'm auto-merging this since checks have passed and it's just data/assets.

PR #511, opened 2023-03-15
Template: Disable BackButton when no history

manuq:
Check window.history because the VueJS router doesn't have the history list.

Also, add styling for the disabled state. Because this component isn't respecting our theme yet.

https://phabricator.endlessm.com/T34612


dylanmccall (approved):
I left a comment in Phabricator about where we're going with this, but I think this is great for now! I'll merge it :)

PR #513, opened 2023-03-22
Utils: Add function to check if user is in the US

manuq:
By getting the browser current timezone, and checking if it's one of the US timezones.

https://phabricator.endlessm.com/T34589

### DIFF
--- a/kolibri_explore_plugin/collectionviews.py
+++ b/kolibri_explore_plugin/collectionviews.py
@@ -51,6 +51,11 @@ PROGRESS_STEPS = {
     "downloading": 0.95,
 }
 
+# FIXME, remove quick patch with a proper per-region content solution:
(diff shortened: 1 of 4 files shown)


kolibri_explore_plugin/collectionviews.py
```diff
@@ -618,6 +630,14 @@ def get_should_resume(request):
     )
 
 
+# FIXME, remove quick patch with a proper per-region content solution:
```
dylanmccall, line 633: And here…

```suggestion
# FIXME, remove quick patch with a proper per-region content solution:
# https://phabricator.endlessm.com/T34589
```

dylanmccall (approved):
Looks good. One tiny suggestion with the FIXME comments :)

PR #521, opened 2023-04-24
Download: Skip channels cache after download

manuq:
The long running caching is enabled in the channels resource by default. Disable it during the showChannels handler.

https://phabricator.endlessm.com/T34704

### DIFF
--- a/kolibri_explore_plugin/assets/src/modules/topicsRoot/handlers.js
+++ b/kolibri_explore_plugin/assets/src/modules/topicsRoot/handlers.js
@@ -188,39 +188,46 @@ export function decideDownload(store) {
 export function showChannels(store) {
   store.commit('CORE_SET_PAGE_LOADING', true);
 
-  return store.dispatch('setAndCheckChannels').then(
-    channels => {
-      if (!channels.length) {
-        router.replace({ name: PageNames.CONTENT_UNAVAILABLE });
-        store.commit('CORE_SET_PAGE_LOADING', false);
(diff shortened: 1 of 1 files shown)


kolibri_explore_plugin/assets/src/modules/topicsRoot/handlers.js
```diff
@@ -217,6 +220,7 @@ export function showChannels(store) {
         });
     },
     error => {
+      ChannelResource.useContentCacheKey = true;
       store.dispatch('handleApiError', error);
       return error;
     }
```
dylanmccall, line 226: I think this could be clearer if we used `Promise.prototype.finally()`:

```js
    error => {
      store.dispatch('handleApiError', error);
      return error;
    }
  ).finally(
    () => ChannelResource.useContentCacheKey = true;
  );
```

dylanmccall (approved):
I have a small suggestion if you'd like, but it seems good to me :)

kolibri_explore_plugin/assets/src/modules/topicsRoot/handlers.js
manuq: Ah absolutely @dylanmccall , this is better for what I meant, to reset the value anyways. Thanks, fixing and merging.

manuq: Unfortunately the pre-commit hook changed the spacing of the whole function with the `finally`. Merging!

PR #522, opened 2023-04-28
T34778 hide discovery

manuq:
https://phabricator.endlessm.com/T34778

### DIFF
--- a/kolibri_explore_plugin/assets/src/components/DiscoveryNavBar.vue
+++ b/kolibri_explore_plugin/assets/src/components/DiscoveryNavBar.vue
@@ -4,6 +4,7 @@
     <img class="logo mr-3" :src="logo">
     <b-button-group class="mx-auto">
       <b-nav-text
+        v-if="showDiscoveryTab"
         class="btn discovery-tab py-3 rounded-0 text-primary"
         :class="{ active: currentIsChannels() }"
         @click="goToChannels"
@@ -46,6 +47,7 @@
 
   import ViewDashboardOutlineIcon from 'vue-material-design-icons/ViewDashboardOutline.vue';
   import MagnifyIcon from 'vue-material-design-icons/Magnify.vue';
+  import plugin_data from 'plugin_data';
 
   import { mapMutations } from 'vuex';
   import { assets } from 'eos-components';
@@ -57,6 +59,9 @@
     computed: {
       logo() {
         return assets.EndlessLogo;
+      },
+      showDiscoveryTab() {
+        return !plugin_data.hideDiscoveryTab;
       },
     },
     methods: {
--- a/kolibri_explore_plugin/kolibri_plugin.py
+++ b/kolibri_explore_plugin/kolibri_plugin.py
@@ -54,6 +54,7 @@ class ExploreAsset(webpack_hooks.WebpackBundleHook):
             "initialContentPack": conf.OPTIONS["Explore"][
                 "INITIAL_CONTENT_PACK"
             ],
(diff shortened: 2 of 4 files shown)


jprvita (approved):
This makes sense to me, although I'm still not very familiar with this codebase. Leaving for @manuq to merge once his back from time off.

kolibri_explore_plugin/kolibri_plugin.py
```diff
@@ -54,6 +54,7 @@ def plugin_data(self):
             "initialContentPack": conf.OPTIONS["Explore"][
                 "INITIAL_CONTENT_PACK"
             ],
+            "hideDiscoveryTab": conf.OPTIONS["Explore"]["HIDE_DISCOVERY_TAB"],
```
GeorgesStavracas, line 57: I have no objections, this seems fine, but I wonder about using either a blocklist or an allowlist for the tabs that should be hidden or shown. That way this can be fine-tuned more easily in the future, for more cases.

kolibri_explore_plugin/assets/src/components/DiscoveryNavBar.vue
```diff
@@ -58,6 +60,9 @@
       logo() {
         return assets.EndlessLogo;
       },
+      showDiscoveryTab() {
+        return !plugin_data.hideDiscoveryTab;
```
GeorgesStavracas, line 64: If internally the semantics of this parameter is positive, perhaps it's better to switch the parameter itself to a positive semantic as well? I.e. `plugin_data.showDiscoveryTab` and the preference in the previous commit.

GeorgesStavracas (commented):
I'm not going to pretend I fully understand what's going on here - especially the last commit! - but left a couple of thoughts nonetheless

kolibri_explore_plugin/assets/src/components/DiscoveryNavBar.vue
manuq: Good point, I started switching the parameter and then realized that the internal semantic of the parameter in the last commit is negative. So the redirects would have to be `redirect: !plugin_data.showDiscoveryTab ? ...` ie "redirect if not show discovery tab".

Here is the context of this requirement: https://phabricator.endlessm.com/T34748#974595 . According to it, I think is better to have a "please hide the tab" setting as an exception for the online instance. Rather than a "show the tab" that's generally what we want.

kolibri_explore_plugin/kolibri_plugin.py
manuq: Interesting idea. I asked Design (Joana) about it and there aren't more tabs planned in the future. Quite the opposite: ideally we should use Kolibri's Library page in the future instead of our current fork. So a blocklist would be a good solution for generalizing the logic but at this point it will be against the KISS principle. Anyways thanks so much for your feedback and thinking about potential improvements!

kolibri_explore_plugin/assets/src/components/DiscoveryNavBar.vue
```diff
@@ -58,6 +60,9 @@
...
+      },
```
GeorgesStavracas, line 65: It seems to me that this method is not used anywhere? Apologies if this is unhelpful.

jprvita: It is called in ` <b-nav-text v-if="showDiscoveryTab"`, which in my understanding means that the whole tag will only be rendered when that method returns true.

I'm also completely new to Vue, but this is what I'm basing my understanding on: https://vuejs.org/guide/essentials/conditional.html

PR #523, opened 2023-05-04
Channels presentation: Add overrides for CAS

manuq:
California Academy of Sciences

https://phabricator.endlessm.com/T34364


packages/template-ui/channel-overrides/cas/assets/header-background.jpg
jprvita, line 1: This has the executable bit set (`rwxr-xr-x`), which is probably because this file came from a windows machine. Although it believe it does not pose a problem in practice, it may be nice to avoid it. (There are a few other files in the same state, we may want to clean them up too eventually).

packages/template-ui/channel-overrides/cas/options.json
```diff
@@ -0,0 +1,5 @@
+{
+  "displayLogoInHeader": false,
+  "isEndlessApp": true,
```
jprvita, line 3: I believe `isEndlessApp` matches the `Is an Endless curated channel?` question from the checklist, correct? If so, :+1:

jprvita (approved):
LGTM. I left one minor nitpick comment, and a question to clarify my understanding. Feel free to merge it after addressing the nitpick, or without addressing if you think it is not worth it (we can always clean it up later with other occurrences of the same thing).

packages/template-ui/channel-overrides/cas/assets/header-background.jpg
manuq: Ah good point! Yes that comes from the asset downloaded from Dropbox. I will fix this (and maybe others) and merge. Thanks!

packages/template-ui/channel-overrides/cas/options.json
manuq: Yes, exactly. That flag adds a "curated by" and the Endless logo in the channel page footer.

PR #526, opened 2023-05-09
wip

manuq:
https://phabricator.endlessm.com/T34740


manuq:
Asking for early feedback! Marking it again as draft because I don't think we want this until we sort out the Kolibri side. See ticket for details.

PR #527, opened 2023-05-10
Landing page: Add feedback button

jprvita:
https://phabricator.endlessm.com/T34665

### DIFF
--- a/kolibri_explore_plugin/assets/src/views/DiscoveryPageContenPacks.vue
+++ b/kolibri_explore_plugin/assets/src/views/DiscoveryPageContenPacks.vue
@@ -59,6 +59,27 @@
             :nodes="sectionNodes['curious']"
           />
         </template>
+        <b-container class="mb-5">
+          <b-row class="d-flex justify-content-center" cols="3">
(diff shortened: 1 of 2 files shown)


kolibri_explore_plugin/assets/src/views/DiscoveryPageContenPacks.vue
```diff
@@ -122,6 +152,9 @@
       // channelLabel: "This week's channel: {channel}",
       careerLabel: 'Explore careers',
       curiousLabel: 'Feeling curious?',
+      aboutLabel: 'About the Endless Key',
```
manuq, line 155: Nit: I think Heather prefers it without the article: "About Endless Key" . And the designs are always carring the error :) . Feel free to ask Heather.

jprvita: That makes sense! I went and changed it, we can double-check with Heather next week.

jprvita:
Updated to the final URL and squashed fixups, this is ready for final review and merging.

PR #528, opened 2023-05-11
Add LICENSE and update author details

wjt:
~~I actually don't know where key@endlessos.org goes but it autocompletes in GMail and is more likely to work than danigm@endlessos.org.~~

key@ goes to our engineering team.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -14,7 +14,7 @@ setup(
     name=dist_name,
     description=description,
     version=kolibri_explore_plugin.__version__,
-    author="EndlessOS",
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -14,8 +14,8 @@
     name=dist_name,
     description=description,
     version=kolibri_explore_plugin.__version__,
-    author="EndlessOS",
-    author_email="danigm@endless.org",
```
manuq, line 18: Oh wow, how this ever happened?!

manuq (approved):

PR #530, opened 2023-05-15
Navigate unavailable channels and content

### LINKED ISSUE #545: Add option to navigate unavailable channels and content
Add a new plugin boolean option `NAVIGATE_UNAVAILABLE` that defaults to false. If true:
- The unavailable channels should be displayed in the Search/Library tab. That is, channels with imported metadata but without any imported content.
- The unavailable content should be displayed. Channels with partial content imported (the case of all channels in a starter pack) will also show the content not imported. The search will also show unavailable content in results.

manuq:
Fixes #545

### DIFF
--- a/kolibri_explore_plugin/assets/src/kolibriApi.js
+++ b/kolibri_explore_plugin/assets/src/kolibriApi.js
@@ -1,4 +1,5 @@
 import urls from 'kolibri.urls';
+import plugin_data from 'plugin_data';
 import { ChannelResource, ContentNodeResource, ContentNodeSearchResource } from 'kolibri.resources';
 
 import { ContentNodeKinds } from 'kolibri.coreVue.vuex.constants';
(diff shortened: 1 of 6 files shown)


jprvita (approved):
LGTM, although I confess I did not understand how this `...(NO_AVAILABLE_FILTERING && { no_available_filtering: true })` works. But I tested this PR and it looks like it is working as expected.

manuq:
> LGTM, although I confess I did not understand how this `...(NO_AVAILABLE_FILTERING && { no_available_filtering: true })` works. But I tested this PR and it looks like it is working as expected.

That is adding the attribute to the object dynamically:
```js
> let x = true
> { foo: 123, ...(x && {bar: x})}
{foo: 123, bar: true}

> let x = false
> { foo: 123, ...(x && {bar: x})}
{foo: 123}
```

As I explained in the commit message It was needed because the Kolibri backend doesn't mind the value. So if I passed:
```js
{
  no_available_filtering: NO_AVAILABLE_FILTERING,
}
```
it was filtering regardless of `NO_AVAILABLE_FILTERING` being `true` or `false`. This can be considered a bug upstream and I should propose a fix.

PR #546, opened 2023-05-25
Re-implement the build and release CI

starnight:
Have:
* `main.yml` renamed from release-apps.yml: build, including release if it is on a tag
* `bump2release.yml`: bump the version with a new commit, and have a corresponding tag

The new commit and tag generated by `bump2release.yml` will be pushed to this repository.  And, the push action will trigger `main.yml` action.

https://phabricator.endlessm.com/T34804


.github/workflows/bump2release.yml
```diff
@@ -0,0 +1,82 @@
...
+            exit 22
+          fi
+
+          pipenv run yarn bump-version major ${{ inputs.VERSION_NAME }}
+
+      - name: Push bump commit & tag to stable branch
+        run: |
+          git show
+
+          repo_url=https://$GIT_USERNAME:$GITHUB_TOKEN@github.com/$GITHUB_REPOSITORY.git
```
dbnicholson, line 75: I don't think you need this. `actions/checkout` already configures all of this so you can just [push using the actions token](https://github.com/actions/checkout#Push-a-commit-using-the-built-in-token). The remote is named `origin` like normal.

.github/workflows/bump2release.yml
```diff
@@ -0,0 +1,82 @@
...
+          echo "Going to push the modification & tag to " $repo_url
+          git push $repo_url
+          tagname=$(git describe --always --tags --match 'v*')
+          echo $tagname
+          git push $repo_url $tagname
```
dbnicholson, line 80: Instead of pushing twice, I think it would be better to push once with both the ref and the tag. The ref should be available as `${{ github.ref_name }}`. So:

```
git push origin ${{ github.ref_name }} $tagname
```

.github/workflows/bump2release.yml
```diff
@@ -0,0 +1,82 @@
...
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
dbnicholson, line 82: I don't believe this is needed as actions/checkout already sets the HTTP authentication in the git configuration using the actions token by default.

.github/workflows/bump2release.yml
```diff
@@ -0,0 +1,82 @@
...
+        type: string
+        description: major version name
+        default: ''
+
+jobs:
+  bump_version:
+    runs-on: ubuntu-latest
+    env:
+      GIT_USERNAME: "No-Reply"
+      GIT_USEREMAIL: "no-reply@endlessos.org"
```
dbnicholson, line 24: I think you can use `${{ github.actor }}` and you'll get the github username of the person triggering the workflow. Then I think you can use that as the username and probably it could be used in the made up email address, too.

See https://docs.github.com/en/actions/learn-github-actions/contexts#github-context for details.

dbnicholson (changes requested):
I left some comments, but nothing major. However, I think @manuq should at least have a look over this since I believe he's the one that's handle most/all of the releases in this repo so far. The deprecation of the `stable` branch likely warrants some consideration.

.github/workflows/bump2release.yml
starnight: good idea!

.github/workflows/bump2release.yml
starnight: cool!

starnight:
Fixed the typo in README

dylanmccall:
I'm not 100% sure this is correct, but I added a trusted publisher in the pypi project (https://pypi.org/manage/project/kolibri-explore-plugin/settings/publishing/) to refer to this workflow, based on a very quick read of this documentation: https://docs.pypi.org/trusted-publishers/.

dbnicholson:
> I'm not 100% sure this is correct, but I added a trusted publisher in the pypi project (https://pypi.org/manage/project/kolibri-explore-plugin/settings/publishing/) to refer to this workflow, based on a very quick read of this documentation: https://docs.pypi.org/trusted-publishers/.

I also created a PyPI `endlessci` user that is now a maintainer of the kolibri-explore-plugin PyPI project. A token for that user is available in the `PYPI_API_TOKEN` secret. So, I think we can use either method. The trusted publisher method is pretty neat.

dbnicholson (approved):
To me this looks good. It doesn't tackle the PyPI upload yet, but it might be better to start with this part that just does the version bumping and tagging. We can validate this works and then add the PyPI parts as documented in https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/.

starnight:
When I try to build my own Linux image with GitHub actions in [starnight/build-image](https://github.com/starnight/build-image), I notice document [Using the GITHUB_TOKEN in a workflow](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#using-the-github_token-in-a-workflow).  And, that reminds me why we add [build_for_tag](https://github.com/endlessm/endless-key-app/blob/e72280364561bfd7cb5a23235ac92ce4e4eb52f2/.github/workflows/tag.yml#LL48C3-L48C16) job into [endless-key-app](https://github.com/endlessm/endless-key-app).  So, I pushed a new commit ("CI: Call the main workflow after bump the version for release").

I think this PR is good to be merged now :)

starnight:
I merge this PR in advance to keep automation moving on!

Issue #534, opened 2023-05-16
Style card of unavailable content


manuq:
The play button ("read", "watch") should be disabled when the content is not available. And a cloud icon should be present:

![Image](https://github.com/endlessm/kolibri-explore-plugin/assets/83944/aa3d6d66-0845-4330-a463-e03db871bd5b)

Current Explore plugin can't display unavailable content, but this draft PR can: https://github.com/endlessm/kolibri-explore-plugin/pull/530/files

So this can be done now based on the PR above assuming that unavailable content will be navigated in the future.

Hint: content cards have the content node as parameter already. The play button ("read", "watch") should be disabled when `node.available` is false.

manuq:
@GeorgesStavracas, I reviewed: https://github.com/endlessm/kolibri-explore-plugin/pull/547#pullrequestreview-1451962270

manuq:
Released internally:
- Endless Key Android: Ladybird 6.2-329
- Endless Key Windows: v1.0.52
- kolibri-explore-plugin: v6.2.0

To verify this a feature flag must be enabled. For Android see https://phabricator.endlessm.com/w/apps/endlesskey/chromebook/#edit-the-kolibri-options

manuq:
@GeorgesStavracas reopening because I've found some errors in the EkCardBody rearrangemnt (commit 6482dd174f0edf5be37d1c8108714c9b1044405b): There is a typo in BootstrapVue components: they should be `<b-container>`, `<b-row>`, `<b-col>`. Vue is raising warnings for unknown components `<v-container>`, `<v-row>`, `<v-col>`. (Note "v" instead of "b"). This suggests that the components aren't actually needed, they are probably treated as a <div>?. Sorry that I didn't saw this during review.

![Captura de pantalla de 2023-06-05 20-09-20](https://github.com/endlessm/kolibri-explore-plugin/assets/83944/d46362c1-d582-419b-82cc-2357553de569)

manuq:
Here is a PR that would fix the cards.

erikos:
https://github.com/endlessm/kolibri-explore-plugin/pull/566 the PR.

erikos:
Landed in Ladybird 6.8-338

erikos:
The new card style can not be seen in the latest release Ladybird 6.8-338 when pulling in cloud content with NAVIGATE_UNAVAILABLE turned on.

manuq:
[...]

PR #547, opened 2023-05-26
Various improvements to cards

GeorgesStavracas:
Starting with the easy bit:

image

Related: https://github.com/endlessm/kolibri-explore-plugin/issues/534

### DIFF
--- a/packages/ek-components/src/components/EkCardBody.vue
+++ b/packages/ek-components/src/components/EkCardBody.vue
@@ -1,27 +1,49 @@
 <template>
   <div class="card-content">
-    <h5 class="mb-1 title">
-      <VClamp
-        autoresize
-        :maxLines="titleLines"
-      >
-        {{ node.title }}
-      </VClamp>
-    </h5>
-    <!-- FIXME only for debugging: -->
-    <span class="debug-node-id sr-only">{{ node.id }}</span>
-    <div v-if="showChannelIcon" class="align-items-center d-flex mb-3">
-      <EkChannelLogo class="mr-2" :channel="node.channel" size="sm" />
-      <span class="channel-title text-muted text-truncate">{{ node.channel.title }}</span>
-    </div>
-    <p v-else class="mb-1 subtitle text-muted">
-      <VClamp
-        autoresize
-        :maxLines="3"
-      >
-        {{ subtitle }}
-      </VClamp>
-    </p>
+    <v-container>
+
+      <!-- Channel -->
+      <v-row v-if="showChannelIcon">
+        <v-col>
+          <!-- FIXME only for debugging: -->
+          <span class="debug-node-id sr-only">{{ node.id }}</span>
+          <div class="align-items-center d-flex mb-3">
+            <EkChannelLogo class="mr-2" :channel="node.channel" size="sm" />
+            <span class="channel-title text-muted text-truncate">{{ node.channel.title }}</span>
+          </div>
+        </v-col>
+      </v-row>
+
+      <!-- Title -->
+      <v-row>
+        <v-col>
+          <h5 class="align-items-center mb-1 title">
+            <VClamp
+              autoresize
+              :maxLines="titleLines"
+            >
+              {{ node.title }}
+            </VClamp>
+          </h5>
+        </v-col>
+      </v-row>
+
+      <!-- Description -->
+      <v-row v-if="!showChannelIcon">
+        <v-col>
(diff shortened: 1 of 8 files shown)


jprvita:
I tested this and the play button was grey on the page listing the resource:
Screenshot from 2023-05-26 16-31-30
But clicking on the resource to go to the resource page (not sure if that is the right name) the play button is still enabled:
Screenshot from 2023-05-26 16-32-04

In addition, there are two things that seem wrong with the UX here:
1. If I try to click on the disabled button on the Library page, the click "falls through" and the effect is the same as if I had clicked on the card (navigating to the resource page), even if the cursor does not turn to the little hand (aka pointer) indicating that area is clickable;
2. In the resource page, the whole card is enabled -- perhaps this is just a side-effect of the button there being enabled too, but we should keep that in mind and address it at the same time.

jofilizola:
> I tested this and the play button was grey on the page listing the resource: Screenshot from 2023-05-26 16-31-30 But clicking on the resource to go to the resource page (not sure if that is the right name) the play button is still enabled: Screenshot from 2023-05-26 16-32-04
Great catch!
In the future we want the resource page to be the same as upstream, so the behavior here would be:
Screenshot 2023-05-30 at 16 33 27

For now, two options:
Screenshot 2023-05-30 at 16 34 22

packages/ek-components/src/components/EkCardBody.vue
```diff
@@ -1,27 +1,49 @@
...
-        {{ subtitle }}
-      </VClamp>
-    </p>
+    <v-container>
+
+      <!-- Channel -->
+      <v-row v-if="showChannelIcon">
+        <v-col>
+          <!-- FIXME only for debugging: -->
+          <span class="debug-node-id sr-only">{{ node.id }}</span>
```
manuq, line 9: I was about to suggest "remove this" but I see it was there before.. so I guess nevermind :)

packages/ek-components/src/components/EkContentCard.vue
```diff
@@ -99,9 +106,18 @@ export default {
...
 .card {
   transition: all ease .4s;
-  &:hover {
+  &:hover:not(.disabled) {
     color: $primary;
   }
+  &.disabled {
+    background: transparent;
+    outline: solid 1px $gray-400;
+    outline-offset: -1px;
```
manuq, line 115: :+1:

packages/ek-components/src/components/EkPlayButton.vue
```diff
@@ -105,4 +113,9 @@ export default {
   text-transform: uppercase;
 }
 
+button:disabled {
+  opacity: 0.33;
+  background-color: rgba(0, 0, 0, 0.25);
```
manuq, line 118: Two options for de-hardcoding these values:
- Add new variables to the theme at `packages/ek-components/src/styles.scss` (if this is wanted only for EkPlayButton)
- Change the style of all disabled buttons in the theme at `packages/ek-components/src/index.scss` (if this is wanted for all disabled buttons)

packages/ek-components/src/styles.scss
```diff
@@ -32,13 +32,13 @@ $modal-backdrop-opacity: 0.8;
...
-  "html5": $yellow,
-  "zim": $yellow,
-  "video": $orange
+  "bundle": $gray-300,
+  "audio": $gray-300,
+  "document": $gray-300,
+  "exercise": $gray-300,
+  "html5": $gray-300,
+  "zim": $gray-300,
+  "video": $gray-300
```
manuq, line 41: :+1: I guess it's not worth the effort refactoring this now as design decision to make all the same gray may change.

packages/ek-components/src/components/EkContentLink.vue
```diff
@@ -17,14 +18,24 @@ export default {
...
   },
   emits: ['isHovered'],
   methods: {
     handleHover(hovered) {
       this.$emit('isHovered', hovered);
     },
     onClick() {
-      this.$router.push(this.url);
+      if (this.enabled)
+        this.$router.push(this.url);
```
manuq, line 38: :+1:

manuq (approved):
All looks good! Feel free to accept the de-hardcoding suggestion and merge.

packages/ek-components/src/components/EkPlayButton.vue
GeorgesStavracas: Done!

manuq (approved):

PR #564, opened 2023-06-05
CI: Trigger Endless Key app build with new kolibri-explore-plugin release

starnight:
Add trigger_endless-key-app job to trigger Endless Key app build with the new kolibri-explore-plugin release with the webhook.

https://phabricator.endlessm.com/T34806

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -114,3 +114,34 @@ jobs:
         uses: pypa/gh-action-pypi-publish@release/v1
         with:
           password: ${{ secrets.PYPI_API_TOKEN }}
+
+
+  trigger_endless-key-app:
+    runs-on: ubuntu-latest
+    if: startsWith(github.ref, 'refs/tags/') || startsWith(inputs.tagname, 'v')
+    needs: [build_apps-bundle, build_kolibri-explore-plugin_whl]
+
+    steps:
+      - name: Get tag name as the release version when at a tag
+        if: startsWith(github.ref, 'refs/tags/')
+        run: |
+          echo "VERSION=${{ github.ref_name }}" >> $GITHUB_ENV
+
+      - name: Get tag name as the release version from input
+        if: ${{ !startsWith(github.ref, 'refs/tags/') && startsWith(inputs.tagname, 'v') }}
+        run: |
+          echo "VERSION=${{ inputs.tagname }}" >> $GITHUB_ENV
+
+      - name: Trigger building endless-key-app
+        uses: actions/github-script@v6
+        with:
+          github-token: ${{ secrets.REPOSITORY_DISPATCH_TOKEN }}
+          script: |
+            await github.rest.repos.createDispatchEvent({
+              owner: "endlessm",
+              repo: "endless-key-app",
+              event_type: "kolibri-explore-plugin-release",
+              client_payload: {
+                VERSION: "$VERSION",
+              }
+            })


.github/workflows/main.yml
```diff
@@ -85,3 +85,28 @@ jobs:
...
+    steps:
+      - name: Get tag name as the release version when at a tag
+        if: startsWith(github.ref, 'refs/tags/')
+        run: |
+          echo "VERSION=${{ github.ref_name }}" >> $GITHUB_ENV
+
+      - name: Get tag name as the release version from input
+        if: ${{ !startsWith(github.ref, 'refs/tags/') && startsWith(inputs.tagname, 'v') }
+        run: |
+          echo "VERSION=${{ inputs.tagname }}" >> $GITHUB_ENV
```
manuq, line 104: Side note: there is a "friendly" version name that changes with each major version. It lives in `kolibri_explore_plugin/VERSION`. Current is "Ladybird".

We use that to form the Android version: Ladybird 6.2-329

where:
- "Ladybird" comes from the VERSION file
- 6.2 is the tag minus the patch (we don't use patch number)
- 329 is the Android specific version

It would be great if the Windows app can follow the same versioning!

wjt:
Removing myself as a reviewer & leaving it in the capable hands of @manuq and @dbnicholson :)

.github/workflows/main.yml
starnight: Microsoft defines MSIX/APPX's version number as `Major.Minor.Build.Revision`.
https://github.com/microsoft/WindowsAppSDK/blob/main/specs/Deployment/MSIXPackageVersioning.md#122-version-number--msix

* We can sync the Major and Minor as kolibri-explore-plugin's version minus the patch
* Build: increase number simply for Windows app build
* Revision: is `0`

However, there is no slot for the VERSION name, "Ladybird".

But, synchronizing the number part into APPX's version number is a good suggestion.

To keep this PR moving, I split this idea as ticket https://github.com/endlessm/endless-key-app/issues/119

starnight: Then, this PR can focus on the webhook.

starnight: Pushed commit ("CI: Sync the tag as kolibri-explore-plugin's major & minor version") into PR [CI: Add bumpversion.yml workflow as a webhook target](https://github.com/endlessm/endless-key-app/pull/118).

manuq: @starnight I wouldn't tie the kolibri-explore-plugin version to the Windows app version. At least on Android there are cases where the app's wrapping code needs a fix that should be released without any changes in the plugin.

In Android the real version (versionCode) is just an integer that must always increase. The Jenkins CI is automatically bumping that version number. Then there is a version name (versionName) that can be whatever string, that's where we put "Ladybird 6.2-329". https://developer.android.com/studio/publish/versioning#versioningsettings

If Windows uses semantic versioning, could this CI automatically increase the patch or minor version?

.github/workflows/main.yml
```diff
@@ -85,3 +85,28 @@ jobs:
...
+
+      - name: Trigger building endless-key-app
+        run: |
+          curl -H "Accept: application/vnd.github.everest-preview+json" \
+               -H "Authorization: token ${{ secrets.BUILD_ENDLESSKEYAPP_TOKEN }}" \
+               --request POST \
+               --data '{"event_type": "kolibri-explore-plugin", "client_payload": { "VERSION": "$VERSION"}}' \
+               https://api.github.com/repos/endlessm/endless-key-app/dispatches
```
dbnicholson, line 112: I was looking around for how to do this and found [this comment](https://github.com/orgs/community/discussions/26323#discussioncomment-3251451) suggesting to use the [github-script action](https://github.com/actions/github-script). That seems a little nicer since it knows how to construct GitHub API requests with proper authentication. Then you'd use the [`rest.repos.createDispatchEvent`](https://octokit.github.io/rest.js/v19#repos-create-dispatch-event) API.

.github/workflows/main.yml
```diff
@@ -85,3 +85,28 @@ jobs:
...
+  trigger_endless-key-app:
+    runs-on: ubuntu-latest
+    if: startsWith(github.ref, 'refs/tags/') || startsWith(inputs.tagname, 'v')
+    needs: [build_apps-bundle, build_kolibri-explore-plugin_whl]
+
+    steps:
+      - name: Get tag name as the release version when at a tag
+        if: startsWith(github.ref, 'refs/tags/')
+        run: |
+          echo "VERSION=${{ github.ref_name }}" >> $GITHUB_ENV
```
dbnicholson, line 99: Both this and the step below add `VERSION` to the environment. Which one is supposed to be used?

.github/workflows/main.yml
starnight: That looks nice with the [token](https://github.com/actions/github-script#using-a-separate-github-token)!

.github/workflows/main.yml
starnight: It is according the "if" condition.

* `startsWith(github.ref, 'refs/tags/')` is for a tag push
* `${{ !startsWith(github.ref, 'refs/tags/') && startsWith(inputs.tagname, 'v') }` is for the workflow_call

.github/workflows/main.yml
starnight: Updated by the new commit.

dbnicholson (approved):
This looks good. I'm going to hold off on merging it since the token doesn't exist and this will just fail in the meantime.

starnight:
This might conflict to https://github.com/endlessm/kolibri-explore-plugin/pull/640
One has to do the rebase according which PR is merged first.

.github/workflows/main.yml
```diff
@@ -86,3 +86,34 @@ jobs:
...
+
+      - name: Get tag name as the release version from input
+        if: ${{ !startsWith(github.ref, 'refs/tags/') && startsWith(inputs.tagname, 'v') }}
+        run: |
+          echo "VERSION=${{ inputs.tagname }}" >> $GITHUB_ENV
+
+      - name: Trigger building endless-key-app
+        uses: actions/github-script@v6
+        with:
+          github-token: ${{ secrets.BUILD_ENDLESSKEYAPP_TOKEN }}
```
dbnicholson, line 110: Per https://github.com/endlessm/eos-administration/commit/e53ba80fd359645f1cb3017f46339c438e2dafb2, I decided to name the secret `REPOSITORY_DISPATCH_TOKEN` since it's not actually scoped to the endless-key-app repo but really to the `eos-shell-build` user.

starnight:
Verified with the comment https://github.com/endlessm/endless-key-app/pull/118#issuecomment-1608765587

Issue #570, opened 2023-06-06
Make control to display offline content available to the UI


jprvita:
We want to allow the user to control if offline content is displayed or not when navigating or searching through their content. For that to be possible we need to expose this control to the UI.

jprvita:
This is not needed anymore, it has been addressed by #569, #572 and #596.

erikos:
Great, archiving.

PR #575, opened 2023-06-07
Parametrize navigate unavailable

jprvita:
https://github.com/endlessm/kolibri-explore-plugin/issues/570

### DIFF
--- a/kolibri_explore_plugin/assets/src/kolibriApi.js
+++ b/kolibri_explore_plugin/assets/src/kolibriApi.js
@@ -1,5 +1,4 @@
 import urls from 'kolibri.urls';
-import plugin_data from 'plugin_data';
 import { ChannelResource, ContentNodeResource, ContentNodeSearchResource } from 'kolibri.resources';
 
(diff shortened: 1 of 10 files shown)


jprvita:
The last two commits here are not working, because `template-ui` cannot find `plugin_data` (which I understand why, but don't know the best approach to fix -- and maybe we actually won't need `plugin_data` there)

jprvita:
> The last two commits here are not working, because `template-ui` cannot find `plugin_data` (which I understand why, but don't know the best approach to fix -- and maybe we actually won't need `plugin_data` there)

I have fixed that now simply by initializing the switch state to `true` instead of using the value from the configuration switch. The next steps here I think are adjusting `setEkChannelInfo` to not use `plugin_data.navigateUnavailable` and have the callers pass that as a parameter, based on the switch state in their views.

Also, at some point I think we want to make all switches share the same state, which is not the case in they way this draft is coming along so far.

jprvita:
What this PR was trying to do has been achieved through different PRs. This isn't needed anymore.

Issue #572, opened 2023-06-07
Add filtering functionality for online/offline content to the search results


jprvita:
Search results, unfiltered:
![image](https://github.com/endlessm/kolibri-explore-plugin/assets/51095367/ecb43837-e6b6-4308-a872-efae6e286eb3)

Search results, filtered:
![image](https://github.com/endlessm/kolibri-explore-plugin/assets/51095367/e300b397-f1ea-489a-9dc9-2c33307b6168)

_Originally posted by @jofilizola in https://github.com/endlessm/kolibri-explore-plugin/issues/569#issuecomment-1579260526_

- [ ] #570

jprvita:
@manuq provided some insight on https://github.com/endlessm/kolibri-explore-plugin/issues/569#issuecomment-1579510036

> The control itself could be a BootstrapVue switch style checkbox: https://bootstrap-vue.org/docs/components/form-checkbox#individual-checkbox-switch-style
> 
> The spec from @jofilizola shows a checkbox control from Material, but I'm afraid we are using Bootstrap through BootstrapVue. The only exception is that we use icons from Material. Switching entirely to Material would be a bigger change, and mixing UI frameworks is not convenient.

jofilizola:
The control can be the Bootstrap one!

jprvita:
This is what I get with #576:
![Screenshot from 2023-06-07 17-02-51](https://github.com/endlessm/kolibri-explore-plugin/assets/73894/b37a0900-e285-4dcb-9618-2afe8c5bb8a3)
![Screenshot from 2023-06-07 17-03-09](https://github.com/endlessm/kolibri-explore-plugin/assets/73894/b3788c6b-da91-4b80-a6c9-fe56364d0331)

manuq:
Reviewed https://github.com/endlessm/kolibri-explore-plugin/pull/576#pullrequestreview-1469821904

jprvita:
Addressed changes, ready for review again.

jofilizola:
I noticed that when the switch is ON, it shows unavailable content.
It seems more intuitive to me to filter out unavailable content when the switch is on, so the other way round.

![image](https://github.com/endlessm/kolibri-explore-plugin/assets/51095367/05d3dda4-8f9d-41c1-9b53-0622662e7337)

erikos:
In the Ladybird 6.8-338 release we do not see the slider in the Library view.

erikos:
[...]

PR #576, opened 2023-06-07
Search page: Add switch to control showing unavailable content

jprvita:
This adds a switch to the search results page that controls whether the
results should include content that is not available on the device
(i.e., needs to be downloaded).

Fixes: https://github.com/endlessm/kolibri-explore-plugin/issues/572

### DIFF
--- a/kolibri_explore_plugin/assets/src/views/SearchPage.vue
+++ b/kolibri_explore_plugin/assets/src/views/SearchPage.vue
@@ -109,6 +109,7 @@
   import _ from 'lodash';
   import { mapMutations, mapState } from 'vuex';
   import { utils, constants, responsiveMixin } from 'ek-components';
+  import plugin_data from 'plugin_data';
 
   import { searchChannelsOnce } from '../modules/topicsRoot/handlers';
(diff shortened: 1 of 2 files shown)


jprvita:
The first commit here is from #575, but it may make more sense to merge this PR first, as-is, and later rebase #575.

kolibri_explore_plugin/assets/src/views/SearchPage.vue
```diff
@@ -21,7 +21,16 @@
 
       <template v-else>
         <b-container>
-          <EkKeywords :words="keywords" @click="removeKeyword" />
+          <b-row alignH="between">
+            <b-col>
+              <b-form-checkbox v-model="showUnavailable" name="check-show-unavailable" switch>
+                Show unavailable content
```
manuq, line 27: At one point we (Dylan and me) decided to start marking all new UI text as translatable. See how other components use the `{{ $tr('foo') }}` in the template part, and define a `$trs: { ... }` in the script part.

manuq (changes requested):
Only needs to mark a string as translatable. The rest looks good!

Issue #548, opened 2023-05-26
Download metadata of additional collections


wjt:
On first launch, the app downloads the metadata, and content, for the collection the user selects.

In order to make it possible to browse the other content in the Endless Key universe (and later, download it), we will make the app go on to download (in the background) the metadata for the other collections.

Check if there is already API in Kolibri to check for and update metadata of channels.

dylanmccall:
I've been working on this one today. Should have something ready to test by the end of the week. The way I'm hooking this up for now is an additional stage in our ek-collections download API that runs after `APPLYING_EXTERNAL_TAGS` and before `COMPLETED` (along with some small adjustments so the frontend understands it can start showing things for any stage after `APPLYING_EXTERNAL_TAGS`).

At some point we'll need to come back over here and add a system for checking for / applying updates to content manifests. At that point it would probably make more sense to stick this functionality over there, but there could be some other fluctuation before then so I figure I can glue it into the same machinery for now :)

dylanmccall:
Some technical limitations mean that my initial solution (#580) _does not_ download channel metadata in the background. It looks like background downloading is being explored via #568, but I'll follow up on Slack about where that is going. In theory, #579 is my attempt to do what we want here before realizing it won't work without rewriting stuff.

manuq:
Is it too bad to download the metadata of additional collections in the foreground? As a first approximation. Ideally we would move all the collections download to use the task manager (it is creating tasks directly now) which will move all to the background. Regardless of the collections progress bar being shown in the frontend or not.

dylanmccall:
> Is it too bad to download the metadata of additional collections in the foreground? As a first approximation. Ideally we would move all the collections download to use the task manager (it is creating tasks directly now) which will move all to the background. Regardless of the collections progress bar being shown in the frontend or not.

In my testing it doesn't take long to download these extra channels, to the point that it's probably better if we don't do it in the background here: if we do, we have to deal with the frontend having a stale channels list while that information rapidly changes in the backend. So let's solve background downloads elsewhere and circle back here after that, but perhaps as a broader effort of making content updates happen :)

manuq:
I gave it another review, only a suggestion left, looks good to merge: https://github.com/endlessm/kolibri-explore-plugin/pull/580#pullrequestreview-1474627916

dylanmccall:
I created a follow-up task for the background download:

- #592

PR #579, opened 2023-06-09
WIP: Download: Import extra channels in the background

dylanmccall:
Issue #548

### DIFF
--- a/kolibri_explore_plugin/assets/src/modules/topicsRoot/handlers.js
+++ b/kolibri_explore_plugin/assets/src/modules/topicsRoot/handlers.js
@@ -124,8 +124,12 @@ function _fetchCarouselNodes(store) {
   );
 }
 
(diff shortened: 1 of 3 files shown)


dbnicholson:
FYI, I'm going to steal the blocking/completed difference for #584.

dbnicholson:
I also wanted to say that after reading #608, I think that's the direction all of this downloading should move in. I.e., only use the backend helper for computing the tasks but have the frontend manage the tasks and the download state transitions.

dbnicholson:
I tried out the method used here and it didn't work. Without directly enqueing the tasks you rely on the frontend to iterate the state machine, but then it's by definition not in the background.

dylanmccall:
I believe this has been superseded by #650 (that one needs a bit of work still, but this one would need _a lot_ of work).

Issue #568, opened 2023-06-06
Implement download of single resource


manuq:
Add a "cloud" button to resource cards that are unavailable. When clicked, the download should start in the background (see #533). The task ID must be stored somewhere in order to keep track of progress and query until the task finished. When the download is complete the card should refresh to be available for playing.

Design reference: https://github.com/endlessm/endless-key-content-private/issues/17#issuecomment-1551338721

Challenge: how would this work in custom presentations? Since they are separate webapps it could be problematic.

**Update:** A number of subtasks have popped up:
- [x] #600 
- [x] #601 
- [x] #602 
- [x] #603

jofilizola:
Card states, when online
![image](https://github.com/endlessm/endless-key-content-private/assets/51095367/bed2c056-e2a5-47a8-a1e6-19042a66f861)

when offline
![image](https://github.com/endlessm/endless-key-content-private/assets/51095367/061f0642-faa7-44e0-8616-6e2becd5df7f)

jofilizola:
Link to design : 
https://www.figma.com/file/ItL7d0N9yJALcu12S50dQn/EK-Ladybird?type=design&node-id=468%3A176283&t=0QSUjCPRYQaJWGM2-1

erikos:
`Challenge: how would this work in custom presentations? Since they are separate webapps it could be problematic.
`
@manuq this sounds concerning. Have you found a way to make this work for our custom presentations?

manuq:
Update: The UI is in place:
https://github.com/endlessm/kolibri-explore-plugin/assets/83944/59d32b58-e25c-4201-add7-5a79c986b978

I had to do lateral things:
- Resurrect the test page that had regressed. This is to work on the theme and UI without actual logic (video above).
- Workaround i18n in custom presentations.

Currently I'm working on a global download manager. Since resource cards are in multiple views there has to be a single place to manage the downloads. It will handle full channels also. The download of collections was simpler because all happens in its own view (DownloadPage).

> `Challenge: how would this work in custom presentations? Since they are separate webapps it could be problematic. ` @manuq this sounds concerning. Have you found a way to make this work for our custom presentations?

[...]

PR #580, opened 2023-06-09
Download: Add a stage to import all channels

dylanmccall:
After the `APPLYING_EXTERNAL_TAGS` stage, the DownloadManager will enter a new `IMPORTING_EXTRA_CHANNELS` stage. In this stage, it should download metadata for any channels which are listed in another content collection.

Issue #548

-----

The parent issue here wants this stage to happen in the background, and I took a look at that in #579. But I think there's some overlap with #568, so I'm opting for something quick and dirty in this pull request and we can deal with downloading channel metadata in the background _separately_.

### DIFF
--- a/kolibri_explore_plugin/collectionviews.py
+++ b/kolibri_explore_plugin/collectionviews.py
@@ -48,7 +48,9 @@ COLLECTION_NAMES = ["0001"]
 
 PROGRESS_STEPS = {
     "importing": 0.1,
-    "downloading": 0.95,
+    "downloading": 0.8,
+    "tagging": 0.9,
+    "completed": 1,
 }
 
 
@@ -68,6 +70,10 @@ class EndlessKeyContentManifest(ContentManifest):
         self.metadata = None
         self.available = None
         super().__init__()
+
+    def get_extra_channel_ids(self):
+        all_channel_ids = _get_channel_ids_for_all_content_manifests()
+        return all_channel_ids.difference(self.get_channel_ids())
 
     def read_from_static_collection(self, grade, name, validate=False):
         self.grade = grade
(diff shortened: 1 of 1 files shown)


manuq (approved):
I think this is a good first approach, as long as the extra collections is not a ridiculous download like Khan channel metadata.

dylanmccall:
I made a small change here based on our talk earlier, so the progress bar updates correctly as additional channels are downloaded (and that part of the operation is between 90% and 100%). It could do with another quick look to be sure it's still all right :)

kolibri_explore_plugin/collectionviews.py
```diff
@@ -421,8 +415,33 @@ def get_status(self):
...
+                    + (PROGRESS_STEPS["completed"] - PROGRESS_STEPS["tagging"])
+                    * current_task_number
+                    / total_tasks_number
+                )
+            else:
+                progress = 0
+
         elif self._stage == DownloadStage.COMPLETED:
-            progress = 1
+            progress = PROGRESS_STEPS["completed"]
```
manuq, line 444: :+1:

kolibri_explore_plugin/collectionviews.py
```diff
@@ -421,8 +415,33 @@ def get_status(self):
...
+                    PROGRESS_STEPS["downloading"]
+                    + (
+                        PROGRESS_STEPS["tagging"]
+                        - PROGRESS_STEPS["downloading"]
+                    )
+                    * current_task_number
+                    / total_tasks_number
+                )
+            else:
+                progress = 0
```
manuq, line 430: I would say "else, progress is the previous step progress" instead of zero. Trying to prevent the progress bar to go back. Note, this was wrong before, you are just carrying the zero from the previous code.
```suggestion
                progress = PROGRESS_STEPS["downloading"]
```

kolibri_explore_plugin/collectionviews.py
```diff
@@ -421,8 +415,33 @@ def get_status(self):
...
+        elif self._stage == DownloadStage.IMPORTING_EXTRA_CHANNELS:
+            if total_tasks_number > 0:
+                progress = (
+                    PROGRESS_STEPS["tagging"]
+                    + (PROGRESS_STEPS["completed"] - PROGRESS_STEPS["tagging"])
+                    * current_task_number
+                    / total_tasks_number
+                )
+            else:
+                progress = 0
```
manuq, line 441: Same here.
```suggestion
                progress = PROGRESS_STEPS["tagging"]
```

kolibri_explore_plugin/collectionviews.py
```diff
@@ -573,6 +606,13 @@ def _get_collections_info_by_grade_name(grade, name):
     }
 
 
+def _get_channel_ids_for_all_content_manifests():
+    channel_ids = set()
+    for content_manifest in _content_manifests:
+        channel_ids.update(content_manifest.get_channel_ids())
+    return channel_ids
```
manuq, line 613: :+1:

manuq (approved):

erikos:
Available in Ladybird 6.8-338

Issue #549, opened 2023-05-26
Download thumbnails for content that is not available locally


wjt:
In Kolibri's terms, these are not part of the channel's metadata but are content items in their own right.

When we download metadata for additional collections besides the one the user picks (#548) we'll then want to download thumbnails for the content in those collections, so that users can explore it (#545) visually.

wjt:
I wrote:

> Can someone more experienced than me check how many files and how much data this corresponds to for all the thumbnails on key.endlessos.org (or give me a crash course on the Kolibri data model and how to get an interactive Python console with the Django ORM models loaded? :) )

dbnicholson:
I ran the following script on the prod key instance:
```
#!/usr/bin/env python3

from collections import defaultdict
from kolibri.core.content.models import File, ChannelMetadata
from operator import itemgetter


def nice_size(num):
    for unit in ('bytes','KB','MB','GB'):
        if num < 1024.0:
            return "%3.1f %s" % (num, unit)
        num /= 1024.0
    return "%3.1f %s" % (num, 'TB')


total = 0
channels = defaultdict(int)
for thumbnail in File.objects.filter(thumbnail=True):
    size = thumbnail.local_file.file_size
    total += size
    channels[thumbnail.contentnode.channel_id] += size

nice_total = nice_size(total)
print(f'Total {nice_total} ({total})')
for channel_id, thumbnail_size in sorted(channels.items(), key=itemgetter(1), reverse=True):
    channel_name = ChannelMetadata.objects.get(id=channel_id).name
    channel_nice = nice_size(thumbnail_size)
    print(f'{channel_id} ({channel_name}) {channel_nice} ({thumbnail_size})')
```

And I ran that by having it be read on stdin like `kolibri manage shell < sizes.py` when run as the `kolibri` user with all the environment variables from `/etc/default/kolibri` set.

Here's what it came up with:

```
Total 1.7 GB (1844833202)
c9d7f950ab6b5a1199e3d6c10d7f0103 (Khan Academy (English - US curriculum)) 1.1 GB (1186440792)
7aca54975a2c415c888d5fe73e0e8163 (हिन्दी) 166.5 MB (174574651)
59b8deeb90f544da923187e77c8d3820 (wikiHow) 88.1 MB (92409113)
914fee213ee146de869016c287116b23 (Chapter Books) 55.2 MB (57849018)
[...]

PR #584, opened 2023-06-09
Download backend: Request all thumbnails for channel

dbnicholson:
In order to provide a rich display of unavailable content, thumbnails for the missing content is desired. When downloading the starter pack content, request all thumbnails for each channel. This is a new option for the `remotecontentimport` Kolibri task added after 0.16.0-alpha14. The option is ignored on older releases.

#549

### DIFF
--- a/kolibri_explore_plugin/__init__.py
+++ b/kolibri_explore_plugin/__init__.py
@@ -1,1 +1,3 @@
 __version__ = "6.16.0"
+
+default_app_config = "kolibri_explore_plugin.apps_config.ExploreConfig"
--- a/kolibri_explore_plugin/apps_config.py
+++ b/kolibri_explore_plugin/apps_config.py
@@ -1,0 +1,11 @@
+from django.apps import AppConfig
+
+
+class ExploreConfig(AppConfig):
+    name = "kolibri_explore_plugin"
+    label = "explore"
+
+    def ready(self):
+        from .tasks import restart_failed_background_jobs
+
+        restart_failed_background_jobs()
--- a/kolibri_explore_plugin/collectionviews.py
+++ b/kolibri_explore_plugin/collectionviews.py
@@ -22,6 +22,7 @@ from rest_framework.exceptions import APIException
 from rest_framework.response import Response
 
 from .tasks import applyexternaltags
+from .tasks import BACKGROUND_QUEUE
 from .tasks import QUEUE
 from .tasks import remotecontentimport
 
@@ -187,6 +188,31 @@ class EndlessKeyContentManifest(ContentManifest):
 
         return tasks
 
+    def get_contentthumbnail_tasks(self):
+        """Return a serializable object to create thumbnail contentimport tasks
+
+        For all the channels in this content manifest.
+        """
+        tasks = []
+
+        for channel_id in _get_channel_ids_for_all_content_manifests():
+            channel_metadata = self._get_channel_metadata(channel_id)
+            tasks.append(
+                {
+                    "task": "remotecontentimport",
+                    "params": {
+                        "channel_id": channel_id,
+                        "channel_name": channel_metadata.name,
+                        "node_ids": [],
+                        "exclude_node_ids": [],
+                        "all_thumbnails": True,
+                        "fail_on_error": True,
+                    },
+                }
+            )
+
+        return tasks
+
     def _get_node_ids_for_channel(self, channel_metadata, channel_id):
         """Get node IDs regardless of the version
 
@@ -222,6 +248,8 @@ class DownloadStage(IntEnum):
     IMPORTING_CONTENT = auto()
     APPLYING_EXTERNAL_TAGS = auto()
     IMPORTING_EXTRA_CHANNELS = auto()
+    FOREGROUND_COMPLETED = auto()
+    IMPORTING_ALL_THUMBNAILS = auto()
     COMPLETED = auto()
 
 
@@ -230,6 +258,12 @@ class DownloadError(Exception):
 
 
 class CollectionDownloadManager:
+    TASKS_MAPPING = {
+        "remotechannelimport": remotechannelimport,
+        "remotecontentimport": remotecontentimport,
+        "applyexternaltags": applyexternaltags,
+    }
+
     def __init__(self):
         self._set_empty_state()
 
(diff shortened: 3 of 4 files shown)


dbnicholson:
Since https://github.com/learningequality/kolibri/pull/10780 was merged, I tested this out on my local checkout. Good news - it works. Bad news - it significantly slows down the initial starter pack download. I tried this with the Scientist starter pack since it appeared to be the smallest and it was spending a lot of time downloading thumbnails. Presumably this is because it includes wikiHow, which has [88 MB of thumbnails](https://github.com/endlessm/kolibri-explore-plugin/issues/549#issuecomment-1568851162).

wjt:
> I tried this with the Scientist starter pack since it appeared to be the smallest and it was spending a lot of time downloading thumbnails. Presumably this is because it includes wikiHow, which has [88 MB of thumbnails](https://github.com/endlessm/kolibri-explore-plugin/issues/549#issuecomment-1568851162).

Might it work to download the thumbnails later as a separate background step that happens after the onboarding flow?

manuq:
This looks good. Moving it to draft to avoid merging it until Kolibri 0.16 alpha15 is released.

dbnicholson:
I changed it so that once the initial download completes, the remaining thumbnails are queued up for downloading in the background. It worked, but I can't say this is particularly robust since the background tasks aren't tracked and I'm not sure if Kolibri will restart them if they fail.

Here's a screenshot with TechBridge Girls showing 2 unavailable nodes after I ran kolibri with `KOLIBRI_NAVIGATE_UNAVAILABLE=true`:

image

manuq:
@dbnicholson the other problem I see is that the UI won't be refreshed with the thumbnails after the background download completes. That's another other reason to have a frontend component directing/monitoring the background job.

dbnicholson:
> @dbnicholson it is a bit hacky but I think it will work. As long as the for loop (the one after marking "COMPLETED") ends without Kolibri stopping. Once the tasks are scheduled Kolibri persists them and they are ran across Kolibri sessions.
> 
> The less-hacky way is: direct the collection download from the App root Vue component (kolibri_explore_plugin/assets/src/app.js). Today it is directed from the Download page (kolibri_explore_plugin/assets/src/views/DownloadPage.vue).

Agreed it is hacky. I probably need a bit more help here. I don't understand how directing it from the app component rather than the download view affects what's going on here. Could you explain?

Regardless, I think the idea would be that once the collection download completes then the frontend would call a separate backend API to initiate the background thumbnail download. Is that right? Would we want to track those tasks in the session state like the collection download?

manuq:
> > @dbnicholson it is a bit hacky but I think it will work. As long as the for loop (the one after marking "COMPLETED") ends without Kolibri stopping. Once the tasks are scheduled Kolibri persists them and they are ran across Kolibri sessions.
> > The less-hacky way is: direct the collection download from the App root Vue component (kolibri_explore_plugin/assets/src/app.js). Today it is directed from the Download page (kolibri_explore_plugin/assets/src/views/DownloadPage.vue).
> 
> Agreed it is hacky. I probably need a bit more help here. I don't understand how directing it from the app component rather than the download view affects what's going on here. Could you explain?

I wish I have a clear understanding too! My gut feeling is that we want to monitor the background task in order to refresh the nodes thumbnail property in the frontend state after the download is completed. And we want that to happen in any view, not just the download view. The component that is always present is the root app, the parent of any every component.

> Regardless, I think the idea would be that once the collection download completes then the frontend would call a separate backend API to initiate the background thumbnail download. Is that right? Would we want to track those tasks in the session state like the collection download?

Yes, maybe split the foreground collection stages from the background stages (extra metadata and thumbnails). My point is that this will need more thinking to happen in the background. Maybe start easy by adding this in fhe foreground first?

About persisting, we decided to go with session state for the download. Maybe it deserves a Django model to persist it in the database? I wouldn't change it for now.

dbnicholson:
> > > @dbnicholson it is a bit hacky but I think it will work. As long as the for loop (the one after marking "COMPLETED") ends without Kolibri stopping. Once the tasks are scheduled Kolibri persists them and they are ran across Kolibri sessions.
> > > The less-hacky way is: direct the collection download from the App root Vue component (kolibri_explore_plugin/assets/src/app.js). Today it is directed from the Download page (kolibri_explore_plugin/assets/src/views/DownloadPage.vue).
> > 
> > 
> > Agreed it is hacky. I probably need a bit more help here. I don't understand how directing it from the app component rather than the download view affects what's going on here. Could you explain?
> 
> I wish I have a clear understanding too! My gut feeling is that we want to monitor the background task in order to refresh the nodes thumbnail property in the frontend state after the download is completed. And we want that to happen in any view, not just the download view. The component that is always present is the root app, the parent of any every component.

OK, I think I understand. We more or less want something attached to the root app that tracks any outstanding download tasks (whether foreground or background) so that any view can see the status of those tasks if they need to.

> > Regardless, I think the idea would be that once the collection download completes then the frontend would call a separate backend API to initiate the background thumbnail download. Is that right? Would we want to track those tasks in the session state like the collection download?
> 
> Yes, maybe split the foreground collection stages from the background stages (extra metadata and thumbnails). My point is that this will need more thinking to happen in the background. Maybe start easy by adding this in fhe foreground first?

Note that @dylanmccall is doing something very similar in #579. I'm going to sync up with Dylan to see if we can converge on a way to handle these background downloads. If that doesn't seem like it's going to materialize quickly then I'll go back to the original commit that simply adds `all_thumbnails=True` to the content download. It really slows things down, though.

> About persisting, we decided to go with session state for the download. Maybe it deserves a Django model to persist it in the database? I wouldn't change it for now.

I think the session state is fine. Kolibri already has all the information we need as long as we keep the job IDs. A couple lists of job IDs stored in the session state seems fine to me.

manuq:
> Note that @dylanmccall is doing something very similar in #579

Oh that's great! Seems to be the way to go for these thumbnails and the extras metadata.

dbnicholson:
This is ready for review again. I attempted to use the process from #579 to queue the background tasks and it didn't work. That method depends on the frontend to keep calling the `ek-collections/update-download` API to query and enqueue the tasks. Obviously it can't do that if it wants to move on.

Instead, I simply queue them in the background as before but a little nicer. I added a separate queue for background tasks, made the plugin restart them when the app starts, and made the enqueing happen in an explicit stage so that we don't reach `COMPLETED` until they're enqueued.

The idea is that the frontend can query the background and foreground queues for completion and update the UI as needed. I started to play with that but my Vue is too weak. I'm hoping someone else (looking at you @manuq) can implement that part since I believe it's generally needed to monitor the content tasks like the Device plugin does.

manuq:
@dbnicholson sorry for missing this one today. @dylanmccall could you review? Otherwise I'll be back on Tuesday.

dylanmccall (approved):
I read through this and it looks good to me. Adding a separate queue for background tasks definitely seems like the way to go :+1:

PR #585, opened 2023-06-12
CI: Fix the missed step id for job's output

starnight:
The build_for_tag job of bump2release.yml will not execute until bump_version job sets the output tagname properly.

This fixes commit 1b8c05e6a04f ("CI: Call the main workflow after bump the version for release").

https://phabricator.endlessm.com/T34804


starnight:
Verified according to https://github.com/endlessm/kolibri-explore-plugin/issues/589#issuecomment-1588746867

Issue #569, opened 2023-06-06
Add filtering functionality for online/offline content to the channel view


erikos:
We have landed the capabilities to view content that is not on your device yet. This can be overwhelming if there is a lot of content. The user should be able to filter the view to further edit the view. Filters is something already used in the channel view, this would be another one. All the rest of the view and order of individual items should stay intact.

![Screenshot from 2023-06-05 08-33-51](https://github.com/endlessm/kolibri-explore-plugin/assets/3714785/1647a717-1286-4e00-b5e0-d8ff8662cdc2)
![Screenshot from 2023-06-05 08-33-42](https://github.com/endlessm/kolibri-explore-plugin/assets/3714785/c9c0fb6c-f529-483d-a25f-4183f04a98c5)

manuq:
Let's consider Search. [This is the line](https://github.com/endlessm/kolibri-explore-plugin/commit/3cc6bbdb41a28e5e290c694dc0c0c2a9b1718a83#diff-d930bb853e45e628a495564d70884234ac9c4da9ef2e2d946fe65d9269b189feR245) that adds `no_available_filtering` conditionally inside the `searchChannelsOnce()` method. Insted, that should be a new parameter. Then the `SearchPage.vue` view would be able to call the method with "filter available" true/false.

manuq:
The control itself could be a BootstrapVue switch style checkbox: https://bootstrap-vue.org/docs/components/form-checkbox#individual-checkbox-switch-style

jofilizola:
The control from Bootstrap is totally fine, I just used the material library for convenience (because of the icons)
![image](https://github.com/endlessm/kolibri-explore-plugin/assets/51095367/417b11e6-7984-4eac-8557-8d6ba073288d)

jprvita:
Moving back to //In Progress// as I believe this is still missing the switch on initial page when entering a channel. #591 only adds the filter to the search results.

erikos:
One major piece is missing here that would lead to inconsistency when landing the button to filter on the channel view top level: 
* a channel can have have sub categories - those work differently than the top level views  @GeorgesStavracas to fill in the details here)

manuq:
Moving to in-progress and assigning @GeorgesStavracas since he is working on this.

erikos:
Not on the to-do-for-now list anymore, back to the backlog.

PR #591, opened 2023-06-12
Filter unavailable content in Channel view

GeorgesStavracas:
There are two open questions about this PR:
 * How to handle translatable strings?
 * ~~The initial state of the check button must come from `plugin_data` somehow. How?~~
   * I've worked around this by adding a global getter to the KolibriApi class. It seems to work in practice, but please let me know if the approach is wrong.

Other than that, seems to work.

https://github.com/endlessm/kolibri-explore-plugin/issues/569

### DIFF
--- a/kolibri_explore_plugin/assets/src/kolibriApi.js
+++ b/kolibri_explore_plugin/assets/src/kolibriApi.js
@@ -105,7 +105,7 @@ class KolibriApi {
 
   searchContent(options) {
     let searchPromise;
-    const { keyword } = options;
+    const { keyword, showUnavailable } = options;
     if (!keyword) {
       searchPromise = Promise.resolve({
         page: 0,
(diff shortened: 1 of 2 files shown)


manuq:
> There are two open questions about this PR:
> 
> * How to handle translatable strings?

For the record we just sync'ed about this. There will be a workaround in https://github.com/endlessm/kolibri-explore-plugin/pull/571/commits and we'll sort it out later in a separate issue. 

> * ~The initial state of the check button must come from `plugin_data` somehow. How?~
>   
>   * I've worked around this by adding a global getter to the KolibriApi class. It seems to work in practice, but please let me know if the approach is wrong.

I think it will be useful to have a "give me all feature flags" API in the template at one point. For now, we are removing the NAVIGATE_UNAVAILABLE flag entirely. So the global getter won't be necessary.

jprvita:
> I think it will be useful to have a "give me all feature flags" API in the template at one point. For now, we are removing the NAVIGATE_UNAVAILABLE flag entirely. So the global getter won't be necessary.

Yes, but we can address that separately.

The branch LGTM for now, but I'm still taking it for a quick spin before merging.

jprvita:
I was able to test this now, merging!

erikos:
Deployed in Ladybird 6.9-339 to internal testing.

Issue #589, opened 2023-06-12
Explore plugin release fails with "GitHub Releases requires a tag"


manuq:
The last "Bump version to release" action succeded to upload the WHL file to PyPi. But failed to build the apps bundle https://github.com/endlessm/kolibri-explore-plugin/actions/runs/5243316309/jobs/9467897415 with error "GitHub Releases requires a tag"

starnight:
According to the [Usage of action/checkout@v3](https://github.com/actions/checkout/blob/v3.5.3/README.md#usage)
```
# The branch, tag or SHA to checkout. When checking out the repository that
# triggered a workflow, this defaults to the reference or SHA for that event.
# Otherwise, uses the default branch.
ref: ''
```
The **ref** is the **reference** or **SHA** when trigger the workflow, not like normal simple `git clone`.  So, the following **checkouts** are at the original reference, not the commit after bump version.

Sorry, I do not understand action/checkout enough.

starnight:
Because `github.ref` is not at a tag, s[oftprops/action-gh-release's optional **tag_name**](https://github.com/softprops/action-gh-release#-customizing) is not at a tag, too.  And, it shows "requires a tag" error.

manuq:
That's fantastic! Awesome work @starnight !!

PR #593, opened 2023-06-13
CI: Fix checkout branch for bump version from wrokflow

starnight:
The action/checkout@v3's ref [1] is the reference or SHA when trigger the workflow by default. So, the checkout action will not checkout at the tag until the "ref" is specified with a tag name.

This commit lists the scenarios of both workflow_call and non-workflow_call for action/checkout and softprops/action-gh-release.

[1]: https://github.com/actions/checkout/blob/v3.5.3/README.md#usage

Fixes: https://github.com/endlessm/kolibri-explore-plugin/issues/589

### DIFF
--- a/.github/workflows/bump2release.yml
+++ b/.github/workflows/bump2release.yml
@@ -37,7 +37,7 @@ jobs:
           git config --global user.email $GIT_USEREMAIL
 
       - name: Set up Python 3.7
-        uses: actions/setup-python@v2
+        uses: actions/setup-python@v4
         with:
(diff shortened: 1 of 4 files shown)


starnight:
The fix includes the commit ("CI: Add workflow dispatch feature into main workflow") adding workflow_dispatch feature.  So, if the overall bump2version flow is failed, but tagged, then maintainers still can trigger the main.yml on the new tag to do the failed release part.

erikos:
Build succeeded.

PR #601, opened 2023-06-13
Template: Workaround i18n

manuq:
The next commit will add the first translatable EK component. But the i18n from Kolibri isn't hooked with the template-ui. So add a dummy function that will return the English string to avoid the template from failing and to keep Discovery/Library using translations.

Issue: #568


jprvita (approved):

GeorgesStavracas (approved):
Opened https://github.com/endlessm/kolibri-explore-plugin/issues/604 to track proper i18n in template-ui. For now, let's move on with this!

PR #603, opened 2023-06-13
Bring back test page

manuq:
Issue #568


GeorgesStavracas (approved):
LGTM, and seems to work from a quick test!

Issue #599, opened 2023-06-13
Browse content: NAVIGATE_UNAVAILABLE is not set to true by default


erikos:
Tested with Ladybird 6.8-338

We need to set the option in order to browse content that is not on your device.

jprvita:
I believe #605 is enough to fulfill the requirements of this ticket, although there is more work related to this, coming up in #596.

erikos:
Deployed in Ladybird 6.9-339 to internal testing.

erikos:
Works as epected.

manuq:
> Works as epected.

OK, marking the issue as done.

Issue #596, opened 2023-06-13
Library: change behavior and labeling for filtering functionality for online/offline content of search results 


erikos:
Follow up of #572

Joana thinks the toggle for showing only downloaded content should be the other way around: it should be off by default, and when enabled should reduce the visible content. The label should be: “Only show downloaded items”

jprvita:
@jofilizola Have you decided on the final label for the button yet?

wjt:
The label we discussed was:

> (• ) Only show downloaded items

jprvita:
> The label we discussed was:
> 
> > (• ) Only show downloaded items

I'm just trying to understand if that decision is final, as up until late next week it was still TBD

wjt:
Yes, that was the string that was finalised today

wjt:
Update: the string may not be final :)

jprvita:
Ready for review

manuq:
merged!

erikos:
Deployed in Ladybird 6.10-342

erikos:
Verified in Ladybird 6.12-342

PR #605, opened 2023-06-13
Drop NAVIGATE_UNAVAILABLE plugin option

jprvita:
This commit drops the NAVIGATE_UNAVAILABLE option from kolibri-explore-plugin, and hardcodes the behavior to the same as if the option had been set to true.

This does not bring any functional change in comparison to running the plugin with KOLIBRI_NAVIGATE_UNAVAILABLE=1.

https://github.com/endlessm/kolibri-explore-plugin/issues/599


jprvita:
Note for reviewers: I know `NO_AVAILABLE_FILTERING` is not really needed anymore, but I'm keeping it around for now to make it easier to find the places in the code where the filtering happens. I may drop it as part of https://github.com/endlessm/kolibri-explore-plugin/issues/596.

manuq (approved):

PR #600, opened 2023-06-13
Bump kolibri tools

manuq:
Issue: #568

PR #608, opened 2023-06-14
Download individual resources in the background - Part one

manuq:
Issue: #568


packages/ek-components/src/components/EkContentCard.vue
```diff
@@ -3,14 +3,14 @@
     noBody
     class="my-2 rounded-lg"
     :class="{
-      'shadow-sm': !isHovered && node.available,
-      'shadow': isHovered && node.available,
-      'disabled': !node.available,
+      'shadow-sm': !isHovered && !isDisabled,
+      'shadow': isHovered && !isDisabled,
```
jprvita, line 7:
```suggestion
      'shadow-sm': !isHovered && isEnabled,
      'shadow': isHovered && isEnabled,
```
This makes it a bit easier to read (IDK if there is a way to avoid the `!` for `isHovered` as well).

jprvita (approved):
This all looks good to me, although I'm still learning how Vue.js works and I'm still unfamiliar with Kolibri's API, so you may want another set of eyes. I haven't tested it myself, but had a play-through of this code yesterday with Manuel and Georges, which is sufficient to me.

Left a small suggestion to improve readability, feel free to take it or leave it.

packages/ek-components/src/components/EkContentCard.vue
manuq: Good catch!

manuq:
> I have nothing else to add, other than JP's comment. It seems fine to me. Despite my best attempts, I could not test this branch, as it was failing to build with `Can't resolve 'kolibri.utils.syncTaskUtils'` but please don't let that block the merging of this PR.

Ah yes, for that please check https://github.com/endlessm/kolibri-explore-plugin/pull/600#issuecomment-1589550620

GeorgesStavracas:
Reinstalling everything from scratch and I could run this. Looks good! Let's land as soon as CI goes green

kolibri_explore_plugin/assets/src/modules/manageContent/handlers.js
```diff
@@ -0,0 +1,78 @@
...
+      resolve(constants.DownloadState.READY);
+    });
+  }
+  // Else, check how the task is going and update the task in the store:
+  return TaskResource.get(task.id).then(newTask => {
+    store.commit('manageContent/SET_CONTENT_DOWNLOAD_TASK', {
+      channelId,
+      contentId,
+      task: newTask,
+    });
```
dbnicholson, line 41: This looks nice and I get why vuex is used to manage the state globally. However, I'm a bit worried about task persistence. If you close the app or the browser tab, then the vuex store data goes away, right? I think we really want something that's going to load and store the tasks from disk.

But now that I look more closely I see that this module was basically taken from the upstream device plugin. I guess later we can add its parts around `managedTasks` that pull the tasks from the backend.

manuq: @dbnicholson you are totally right. The Vuex state is not persisted, it is only for the current run. And that's why I left the `SET_ALL_CONTENT_DOWNLOAD_TASKS` setter to set all at once at the beginning. Although it isn't used yet. 

The Device tab is also not persisting the tasks on disk. What it does is fetching all tasks from the task manager and refreshing the Vuex state: https://github.com/learningequality/kolibri/blob/ea1e0754add20f7705c7a5da9a9a8883a1eca47e/kolibri/plugins/device/assets/src/modules/manageContent/actions/taskActions.js#L40

The problem is that we need to know which task corresponds to which resource in order to recreate the state, but fetched tasks don't have enough info for doing it. The task has `channel_id` and `channel_name` in extra metadata, but it doesn't have the original `node_ids` parameter. I was thinking in passing node_id as the channel_name (we put "foo" now) as a hacky solution. What do you think?

dbnicholson: Yikes, the tasks API is really not helpful. The job returned by the API is [filtered](https://github.com/learningequality/kolibri/blob/ea1e0754add20f7705c7a5da9a9a8883a1eca47e/kolibri/core/tasks/api.py#L80) from a [Job](https://github.com/learningequality/kolibri/blob/ea1e0754add20f7705c7a5da9a9a8883a1eca47e/kolibri/core/tasks/job.py#L116), which has the original `args` and `kwargs`. Furthermore, it appears `extra_metadata` is not what's provided by the caller (I'm pretty sure we can drop our usage of it) but rather [set by the validator](https://kolibri-dev.readthedocs.io/en/develop/backend_architecture/tasks.html#enqueuing-tasks-via-the-post-api-tasks-tasks-api-endpoint).

So, I guess abusing `channel_name` would be fine in the short term, but I really think the tasks API views should return the `args` and `kwargs`. I can't see why they shouldn't exposed and I think someone should make a PR upstream for it.

Another option would be to persist the mapping from node ID download (or whatever is our useful identifier) to job ID in the browser's `localStorage`. It kinda stinks since it's another data source that needs to be kept in sync with the others, but I don't think it would be too hard to manage.

dbnicholson: It also wouldn't be hard to persist the `node_ids` and `exclude_node_ids` in `extra_metadata` in addition to [`kwargs`](https://github.com/learningequality/kolibri/blob/ea1e0754add20f7705c7a5da9a9a8883a1eca47e/kolibri/core/content/tasks.py#L82). That seems useful and fits within the current system.

dbnicholson: https://github.com/learningequality/kolibri/pull/10853 was merged, so after the next release you can parse out the node IDs to see what the task was for. What I'd actually like even better (and may propose upstream) is that you can specify `extra_metadata` and it's merged with what the validator puts in there. That would allow you to put your own label in there for better identification.

Issue #610, opened 2023-06-14
Decouple collections from app builds


manuq:
Fetch the collection JSON manifests from Github release. Currently we serve them as static files which are included in the wrapper apps (Android, Windows) at build time. This way the collections release will be decupled from software releases.

PRs:
- [ ] https://github.com/endlessm/kolibri-installer-android/pull/150
- [ ] https://github.com/endlessm/endless-key-app/pull/122
- [ ] https://github.com/endlessm/kolibri-explore-plugin/issues/610

manuq:
I also created PRs for the wrapper apps, so they stop bundling the collections.

erikos:
Let's move to the next iteration to not risk any issues before we get the Ladybug release out.

manuq:
> Let's move to the next iteration to not risk any issues before we get the Ladybug release out.

Sounds good! I'm also unassigning myself so others can take it from here. The PR already got good feedback from @dylanmccall @dbnicholson and @pwithnall .

dylanmccall:
To smooth the transition here, I am proposing two more PRs which only affect the build system:

- endlessm/kolibri-installer-android#152
- endlessm/endless-key-app#124

dbnicholson:
Last night I whipped up a PR to publish the endless-key-collections JSON to [GitHub Pages](https://github.com/endlessm/endless-key-collections/pull/29). I think we should go back to @manuq's PRs and use https://endlessm.github.io/endless-key-collections/ as the base URL.

manuq:
Blocking until after ISTE. Please unblock and review again after it.

manuq:
I'm deassigning myself from this one. Someone else could follow up the PR.

manuq:
We decided to don't go this way.

PR #611, opened 2023-06-14
Collection download: Fetch manifests from Github

manuq:
This way the software update is decoupled from content update.

Issue #610


kolibri_explore_plugin/collectionviews.py
```diff
@@ -565,12 +565,15 @@ def _build_remotechannelimport_task(channel_id):
...
+_collection_download_manager = None
 
 
-def _read_content_manifests():
+def _initiate():
     global _content_manifests
     global _content_manifests_by_grade_name
+    global _collection_download_manager
+
+    _collection_download_manager = CollectionDownloadManager()
```
dbnicholson, line 576: I'm not sure why the download manager initialization needs to be delayed. The only time the manifests come into play is when other code calls `_collection_download_manager.from_manifest()` or `_collection_download_manager.from_state()`.

kolibri_explore_plugin/collectionviews.py
```diff
@@ -27,6 +28,12 @@
 
 logger = logging.getLogger(__name__)
 
+COLLECTION_URL_TEMPLATE = (
+    "https://raw.githubusercontent.com/endlessm"
+    "/endless-key-collections/main/json"
+    "/{grade}-{name}.json"
+)
```
dbnicholson, line 35: This seems pretty dangerous to make this use the `main` HEAD commit at runtime. If someone commits a mistake to `main` or is in the process of refactoring the collections then all apps are broken immediately. I think if we want to move in this direction then we should make releases of endless-key-collections and use released artifacts.

kolibri_explore_plugin/collectionviews.py
manuq: Well, today the manifests are read and parsed at import time. For example when `kolibri manage migrate` is called and this plugin is enabled. Not ideal. Worse would it be to add an online fetch when running migrations.

kolibri_explore_plugin/collectionviews.py
manuq: Yeah. @pwithnall started creating releases already. There is a ZIP file with all the source. I wonder if the JSON files could be separate release artifacts so they can be requested directly. Otherwise this code should request the ZIP and unpack it.

manuq: Okey. The other improvement I'd do is allow changing the URL in an `options.ini` plugin setting. That way the same app build will be able to test different collections. It can be useful for content curation and testing (passing a mock collection with fewer content to avoid long download times).

dbnicholson: Another option which doesn't require releases and artifacts but is a little safer is to have a `stable` branch in endless-key-collections that we occasionally fast forward to a known good point on master. Then you can continue using the direct github links with more confidence that it won't randomly break.

Another thing to consider is using github pages since I expect that would scale a lot better than the raw links. We could arrange that the pages branch is only updated on release.

dbnicholson: The collection JSON is now published to https://endlessm.github.io/endless-key-collections per https://github.com/endlessm/endless-key-collections/pull/29, so I think you should change the base URL in this PR. When new tags are made in endless-key-collections, the published JSON will automatically be updated.

manuq:
Ready for review again. I have updated the URL according to https://github.com/endlessm/kolibri-explore-plugin/pull/611#discussion_r1235630239 and made it configurable. I think this will be useful to test unreleased collections and alternate collections (quickly iterate through the process with minimum content).

PR #615, opened 2023-06-14
Bump webpack dependencies

### LINKED ISSUE #614: Fix webpack errors
Local development has become a bit difficult when we migrated to Kolibri 0.16 alpha14, and even more when migrated to alpha15. This is because webpack is showing a overlay with multiple errors on each page reload. So we devs have to close 6 or 7 dialogs each time code changes.

There is no webpack configuration in kolibri-explore-plugin directly. Instead the configuration is picked from kolibri-tools. See `WebpackBundleHook` in https://kolibri-dev.readthedocs.io/en/develop/backend_architecture/plugins.html?highlight=plugin#example-implementation

We should make sure that the plugin has the same webpack version than Kolibri.

- [ ] https://github.com/endlessm/kolibri-explore-plugin/pull/615

manuq:
To be the same as current Kolibri

Fixes #614

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,6 +25,10 @@
     "packages/*"
   ],
   "devDependencies": {
+    "webpack": "^5.82.1",
+    "webpack-cli": "^5.1.1",
+    "webpack-dev-server": "^4.15.0",
+    "webpack-merge": "^5.8.0",
     "lerna": "^4.0.0",
     "sass": "^1.49.9"
   },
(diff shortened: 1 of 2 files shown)


manuq:
@GeorgesStavracas @jprvita again you'll have to run `yarn install` for testing this. And in master whenever is merged.

jprvita (approved):
Nice and thank you very much!

PR #616, opened 2023-06-14
Invert show/hide available content logic

jprvita:
Invert the logic so when the UI component is active the content that is
not downloaded on the device is hidden.
    
https://github.com/endlessm/kolibri-explore-plugin/issues/596


manuq (approved):
All good. The detailed commits are very appreciated!

PR #622, opened 2023-06-15
SectionsSearchRow: Hide download button

GeorgesStavracas:
It was agreed on, as it doesn't do what people
expect it to do.


jprvita (approved):
That's a creative way to neuter a line! LGTM

Issue #598, opened 2023-06-13
Discovery page: keeps on loading without showing the cards


erikos:
Tested with Ladybird 6.8-338 

After downloaded the initial pack and then displayed it once - restarting the app will result in a loading screen of the Discovery page without ever finishing. Switching to the Library tab and back displays the cards.

![Screenshot from 2023-06-13 15-17-38](https://github.com/endlessm/kolibri-explore-plugin/assets/3714785/deed64c6-f876-4c7a-9721-afca676a0f80)

- [ ] https://github.com/endlessm/kolibri-explore-plugin/pull/628

GeorgesStavracas:
I will try and figure this out, but this seems to be a complicated issue and I'm afraid there's a chance I'll fail to deliver this on time for this sprint. Just something to keep in mind.

vanessa-chang:
I'm seeing this issue today. (test package: v6.9-340)
1. After content completes the download, it shows a "Something went wrong!" message.
2. Press "try again" button does not help.
3. Close and reopen the app, it will then running into this issue.

erikos:
> I'm seeing this issue today. (test package: v6.9-340)
> 
> 1. After content completes the download, it shows a "Something went wrong!" message.
> 2. Press "try again" button does not help.
> 3. Close and reopen the app, it will then running into this issue.

Just to make sure. Steps 1 and 2 should be independent of this issue. This can be always reproduced independent if something went wrong in the download phase.

manuq:
It seems that disabling the browser cache is not working anymore in current Kolibri alpha. This is the same issue we are facing after the resources download in #568. Before, [these 2 lines](https://github.com/endlessm/kolibri-explore-plugin/blob/a0af9d6a559dfbba2b507fcb882501da308a2a26/kolibri_explore_plugin/assets/src/modules/topicsRoot/handlers.js#L194-L195) helped but that's no longer the case. See 61636fd6c for reference.

A workaround is to bring back the page reloading 4eb00dbd . Basically calling `location.reload()` in the frontend after the download completes. I'm afraid that workaround won't be a good fit for the individual resources download.

jprvita:
@GeorgesStavracas is taking some days off, so we agreed that I should work on this instead.

jprvita:
It looks like this is always reproducible in a development environment (in the browser) by simply opening a new tab and pointing it to the explore plugin address (i.e., 127.0.0.1:8000/explore).

jprvita:
[...]

PR #628, opened 2023-06-16
Startup: Don't fetch carousel for collections

manuq:
These hardcoded nodes are for EK Iguana (USB) which has a fixed presentation. But the dynamic collections don't have them, so they are provoking 404s and doing unneeded calculations on startup.

#598


jprvita (approved):
Great!

Issue #625, opened 2023-06-16
Update card after download of single resource


erikos:
Follow up of #568

At the moment when the download of the single resource has not finished the card is not updating. No thumbnail for the item is shown. 

- [ ] https://github.com/endlessm/kolibri-explore-plugin/pull/629
- [ ] https://github.com/endlessm/kolibri-explore-plugin/pull/634

erikos:
Potentially this is not an issue anymore when we have the thumbnails downloaded as part of the metadata. Not sure if there are any other consequences than the thumbnail not being displayed.

manuq:
@erikos thanks for splitting this into a separate task.

> Potentially this is not an issue anymore when we have the thumbnails downloaded as part of the metadata. Not sure if there are any other consequences than the thumbnail not being displayed.

Right. If we assume that thumbnails will be there, I can workaround by just setting the card to "available" in the frontend.

manuq:
I managed to do this reactively (without any reloading) in the Search page. Adding the PR as a subtask. Next is the custom presentation.



https://github.com/endlessm/kolibri-explore-plugin/assets/83944/33cb2182-c4f1-4eb4-b6f9-1e2423440bd7

manuq:
I'm still not sure why the nodes keep cached in the custom presentation, but I managed to workaround it in a separate PR.

So far I did Home, I still need to do this in 3 more views:
- packages/template-ui/src/views/ListSection.vue
- packages/template-ui/src/views/Section.vue
- packages/template-ui/src/components/FilterResult.vue

manuq:
Template partr ready, considering every case where we have downloadable cards in the custom presentation https://github.com/endlessm/kolibri-explore-plugin/pull/634

manuq:
I'm going to merge as-is because there were reviews already and we can improve in the future. Specially finding out what's the matter with kolibri cache.

vanessa-chang:
Verified passed in Ladybird 6.15-346.

Windows issue: #641

PR #629, opened 2023-06-16
Search: Refresh card after downloading

manuq:
Add a callback to the "node updated" signal and change the corresponding array. For that Vue.$set has to be used so the app reacts to the change.

Fixes part of #625

### DIFF
--- a/kolibri_explore_plugin/assets/src/views/SearchPage.vue
+++ b/kolibri_explore_plugin/assets/src/views/SearchPage.vue
@@ -42,6 +42,7 @@
               :nodes="nodes"
               :mediaQuality="mediaQuality"
               :cardColumns="cardColumns"
+              @nodeUpdated="onNodeUpdated"
             >
               <div>
                 <h4 class="text-muted">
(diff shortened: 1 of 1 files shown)


kolibri_explore_plugin/assets/src/views/SearchPage.vue
```diff
@@ -283,6 +285,23 @@
         this.query = words.join(' ');
         this.$router.push({ params: { query: this.query } });
       },
+      onNodeUpdated(nodeId) {
+        ContentNodeResource.fetchModel({ id: nodeId }).then(newNode => {
```
dbnicholson, line 289: Isn't this going to suffer the same caching issue you were just addressing in #634?

manuq: > Isn't this going to suffer the same caching issue you were just addressing in #634?

Nope. That is "the mistery of the day" to me. Why `ContentNodeResource.fetchModel` works well here and not in `kolibriApi`.

jprvita (approved):
I'm not familiar with this part of the code, but it looks OK to me, especially if it avoids the problem for now. It seems a good idea to have something that visually works this week, and later improve upon it.

(Approving without merging -- @manuq feel free to go ahead when you are comfortable)

PR #630, opened 2023-06-16
Import content cleanup

dbnicholson:
Here's a couple things I came across while working on #549. I decided to split them out instead of including them in that PR.


manuq (approved):
All very welcome improvements! Thanks for finding why the tasks weren't showing in the Device tab. So this means that we are actually using the task manager, just maybe not in the expected way through the frontend Resource objects.

dbnicholson:
> All very welcome improvements! Thanks for finding why the tasks weren't showing in the Device tab. So this means that we are actually using the task manager, just maybe not in the expected way through the frontend Resource objects.

I couldn't figure out how to get the Device tab up, so I'm not totally sure. However, I'd also read the Device code like you and came to the conclusion that it's tracking everything in the `content` queue (which you'd already wired up!).

PR #631, opened 2023-06-16
CI: Only run workflows on push to master or stable

dbnicholson:
Currently when you open a PR, the Build and Lint workflows run twice because they run on any push event in addition to the pull request event. Instead, limit the push events to the master or stable branch.


dbnicholson:
Notice how this PR has 4 workflows, but #584 has 6 workflows with the Build and Test workflows repeated.

PR #634, opened 2023-06-16
Update nodes template

manuq:
Fixes part of #625

### DIFF
--- a/kolibri_explore_plugin/assets/src/kolibriApi.js
+++ b/kolibri_explore_plugin/assets/src/kolibriApi.js
@@ -103,7 +103,10 @@ class KolibriApi {
     });
   }
 
-  getContentById(id) {
+  getContentById(id, ignoreCache = false) {
+    if (ignoreCache && id in ContentNodeResource.cache) {
+      delete ContentNodeResource.cache[id];
+    }
     return ContentNodeResource.fetchModel({ id: id });
   }
 
--- a/packages/template-ui/src/components/FilterResult.vue
+++ b/packages/template-ui/src/components/FilterResult.vue
@@ -13,6 +13,7 @@
         :mediaQuality="mediaQuality"
         :cardColumns="cardColumns"
         @loadMoreNodes="onLoadMoreNodes"
+        @nodeUpdated="onFilterNodeUpdated"
       />
     </div>
   </div>
@@ -21,9 +22,11 @@
 <script>
   import { mapState } from 'vuex';
   import { constants } from 'ek-components';
+  import updateNodeMixin from '@/mixins/updateNodeMixin';
 
   export default {
     name: 'FilterResult',
+    mixins: [updateNodeMixin],
     props: {
       node: {
         type: Object,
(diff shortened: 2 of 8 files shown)


kolibri_explore_plugin/assets/src/kolibriApi.js
```diff
@@ -103,7 +103,10 @@ class KolibriApi {
     });
   }
 
-  getContentById(id) {
+  getContentById(id, ignoreCache = false) {
+    if (ignoreCache && id in ContentNodeResource.cache) {
+      delete ContentNodeResource.cache[id];
+    }
```
dbnicholson, line 109: That's unfortunate. It seems like `ContentNodeResource.cache` is entirely separated from `ContentCacheKey`. It sorta seems like the entire `ContentNodeResource.cache` should be invalidated when `ContentCacheKey` changes, but I get the impression that JS only gets the `ContentCacheKey` on initial page load and isn't aware when it changes.

dbnicholson (approved):
Seems reasonable to me.

jprvita (approved):
I'm not familiar with this part of the code, but it looks OK to me, especially if it avoids the problem for now. It seems a good idea to have something that visually works this week, and later improve upon it.

(Approving without merging -- @manuq feel free to go ahead when you are comfortable)

kolibri_explore_plugin/assets/src/kolibriApi.js
manuq: @dbnicholson I think the flow in Kolibri is:
- Content is downloaded, then the [set_content_visibility()](https://github.com/learningequality/kolibri/blob/9d118277aa2c3f6a89e2daf459dddb9a96af7aa4/kolibri/core/content/utils/annotation.py#L707) annotation is called which calls [ContentCacheKey.update_cache_key()](https://github.com/learningequality/kolibri/blob/9d118277aa2c3f6a89e2daf459dddb9a96af7aa4/kolibri/core/content/utils/annotation.py#L701)
- The next API request will be a `GET /api/content/contentnode/` with a new cache key eg: `?contentCacheKey=1686918514.6423008`

manuq:
Ready for final review. I ended up moving the template-ui logic to a mixin.

kolibri_explore_plugin/assets/src/kolibriApi.js
dbnicholson: Right, but the problem is that you're never making the request to the backend because it's preferring it's in-memory cached representation. If you look at how [`ContentNodeResource`](https://github.com/learningequality/kolibri/blob/develop/kolibri/core/assets/src/api-resources/contentNode.js) works, it never looks at `contentCacheKey`. What [`contentCacheKey`](https://github.com/learningequality/kolibri/blob/develop/kolibri/core/content/api.py) is used for is to add `Cache-Control` headers to backend responses so that the frontend will use the cached responses. However, `Cache-Control` is handled at the browser request and cache level. If you never make the request because you have some custom in memory JS cache like this, then it won't do anything.

Issue #627, opened 2023-06-16
Content download speed regression in Kolibri 0.16


erikos:
Tracking upstream issue:

- https://github.com/learningequality/kolibri/issues/10848

erikos:
Joana reports that she has not been able to download a content pack once in the last three days. Bumping up priority to unbreak now.

wjt:
@dbnicholson @dylanmccall between you, can you figure out how we can work around this download speed regression for Ladybug? Particularly now that we download all channels' metadata during onboarding, and my reading of https://github.com/endlessm/kolibri-explore-plugin/issues/549 is that the intention is to also eagerly download 500MB of thumbnails, this is a real issue. (I'm just a guy but, is it straightforward to revert the change that switched to chunked downloads? Or bump up the chunk size?)

dbnicholson:
In #549 I'm just about done with background downloading and it worked fine in my testing. However, your point is completely valid.

The most straightforward thing to do would be to copy the file transfer code from 0.16.0a12 (last release before the range requests were landed), and then wire up our own download manager and task in the explore plugin. It should be pretty mechanical.

I updated my [download test script](https://gist.github.com/dbnicholson/c9ffc907546d8b866ad6210f6442a2e1) to compare 0.15.12, 0.16.0a12 and 0.16.0a15 and 0.16.0a12 matched the 0.15.12 performance:

```
Min 0.15 time 7.535127711016685 seconds
Min 0.16.0a12 time 7.166952276020311 seconds
Min 0.16.0a15 time 55.26884927297942 seconds
```

erikos:
https://github.com/endlessm/kolibri-explore-plugin/pull/638

erikos:
Perfect, has been merged and I kicked off a new release of the plugin.

erikos:
Ladybird 6.16-347

vanessa-chang:
Verified Failed:

It fails to proceed the download after you select a content pack.
Only a white screen is shown with no progress bar. 

Test packages: Ladybird 6.16-347 & Windows 6.16.0.0

manuq:
Yes, as Vanessa says, the application now shows a white screen right after Kolibri starts. In the logs I see a browser console error "get_download_status is not a function":
```
[...]

PR #638, opened 2023-06-16
Vendor old file download code

dbnicholson:
Since 0.16.0-alpha13, Kolibri's `FileDownload` class uses HTTP range requests to move towards resumable downloads. This significantly slows down network throughput for large downloads as it requires many serial requests to retrieve a single file. Upstream has been made aware and will presumably fix the regression.

In the meantime, this vendors the `FileDownload` code from 0.16.0-alpha12 and adds the necessary plumbing to use it. Comparing the `FileDownload` implementation with the inventor pack, I measured the following timings for the initial content pack download:

Upstream FileDownload:
Start 14:15:18,457
End 14:22:32,008
Total 7m14s

0.16.0-alpha12 FileDownload:
Start 14:09:19,618
End 14:12:01,681
Total 2m42s

#627

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,6 @@
+[flake8]
+extend-exclude =
+  build/,
+  static/,
+  dist/,
+  node_modules/
--- a/.flake8
+++ b/.flake8
@@ -3,4 +3,5 @@ extend-exclude =
   build/,
   static/,
   dist/,
-  node_modules/
+  node_modules/,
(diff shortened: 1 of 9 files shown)


jprvita (approved):
This looks sane to me.

dylanmccall (approved):
This looks good to me as well! Thank you for figuring it out, @dbnicholson. I hope we'll be able to revert it soon :b

PR #640, opened 2023-06-19
CI: Trigger kolibri-installer-android job with new kolibri-explore-plugin release

starnight:
Add trigger_kolibri-installer-android job to trigger "kolibri-installer-android" job on EOSF's Jenkins with the new kolibri-explore-plugin release. It will execute the new introduced trigger_jenkins.py using python-jenkins API to send a webhook.

- [x] https://github.com/endlessm/kolibri-installer-android/pull/153

https://phabricator.endlessm.com/T34805

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -145,3 +145,38 @@ jobs:
                 VERSION: "${{ env.VERSION }}",
               }
             })
+
+
+  trigger_kolibri-installer-android:
+    runs-on: ubuntu-latest
+    if: startsWith(github.ref, 'refs/tags/') || startsWith(inputs.tagname, 'v')
+    needs: [build_apps-bundle, build_kolibri-explore-plugin_whl]
+    env:
+      branch: ${{ github.ref }}
+
+    steps:
+      - name: Set branch if this is triggered by bump2release.yml
+        if: startsWith(inputs.tagname, 'v')
+        run: |
+          # The kolibri-explore-plugin release flow from bump2release.yml will
+          # append a new commit to master branch. Therefore, the github.ref is
+          # not at the latest commit. Checkout master instead.
+          echo "branch=${{ inputs.tagname }}" >> $GITHUB_ENV
+
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          ref: ${{ env.branch }}
+
+      - name: Install trigger_jenkins dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y python3-pip
+          python3 -m pip install requests python-jenkins
+
+      - name: Trigger building kolibri-installer-android
+        run: |
+          ./scripts/trigger_jenkins.py kolibri-installer-android UPLOAD=true
+        env:
+          JENKINS_USER: ${{ secrets.JENKINS_USERNAME }}
+          JENKINS_PASSWORD: ${{ secrets.JENKINS_PASSWORD }}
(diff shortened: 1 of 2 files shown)


starnight:
Pushed a modified commit to make lint happy.

starnight:
This is blocked on https://github.com/endlessm/kolibri-installer-android/pull/153

starnight:
Rebase upon master branch.

starnight:
Sync the parameter `UPLOAD` with https://github.com/endlessm/kolibri-installer-android/pull/153

.github/workflows/main.yml
```diff
@@ -145,3 +145,32 @@ jobs:
...
+          echo "branch=${{ inputs.tagname }}" >> $GITHUB_ENV
+
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          ref: ${{ env.branch }}
+
+      - name: Trigger building kolibri-installer-android
+        run: |
+          scripts/trigger_jenkins.py --build-url kolibri-installer-android UPLOAD=true
```
dbnicholson, line 173: I don't know how `PATH` is setup, but I suspect you want to run this as `./scripts/trigger_jenkins.py`.

Also, I doubt the runner is going to have all of `python3`, `requests` and `python-jenkins` installed, so running this script will just fall over. I think for such a simple case you could just add another step before this that installs the prereqs from Ubuntu:

```
run: |
  sudo apt update
  sudo apt install -y python3 python3-jenkins python3-requests
```

dbnicholson (changes requested):

dbnicholson:
I addressed the changes. I decided to use `pip` to install the dependencies. Probably there should be some kind of requirements file, but meh. Also, I dropped the `--build-url` option for `trigger_jenkins.py` as that will wait until the job is actually running. Since kolibri-installer-android is very slow, that means the action could block for a long time. I think as long as the job is queued, that's good enough.

.github/workflows/main.yml
```diff
@@ -145,3 +145,38 @@ jobs:
...
+    needs: [build_apps-bundle, build_kolibri-explore-plugin_whl]
+    env:
+      branch: ${{ github.ref }}
+
+    steps:
+      - name: Set branch if this is triggered by bump2release.yml
+        if: startsWith(inputs.tagname, 'v')
+        run: |
+          # The kolibri-explore-plugin release flow from bump2release.yml will
+          # append a new commit to master branch. Therefor, the github.ref is
```
wjt, line 162:
```suggestion
          # append a new commit to master branch. Therefore, the github.ref is
```

scripts/trigger_jenkins.py
```diff
@@ -0,0 +1,179 @@
...
+
+    # Poll the queue for a bit
+    waited = 0
+    while True:
+        data = jenkins.get_queue_item(queue_number)
+
+        if "executable" in data:
+            print(data["executable"]["url"])
+            break
+        elif data.get("cancelled", False) is True:
```
wjt, line 160:
```suggestion
        elif data.get("cancelled", False):
```

or is there some reason we specifically want to compare for identity with `True` rather than truthiness?

wjt (approved):
LGTM, a couple of minor suggestions

scripts/trigger_jenkins.py
dbnicholson: Good question! This is a copy of something I wrote [several years ago](https://github.com/endlessm/eos-build/blob/master/ci-integration/jenkins-job-trigger), so I'd have to ask myself :face_with_spiral_eyes:. My guess is that I wasn't sure of the actual schema at the time and was worried that you'd get an empty string.

starnight:
[Build for release / trigger_kolibri-installer-android](https://github.com/endlessm/kolibri-explore-plugin/actions/runs/5383904247/jobs/9771639271#logs) sends the web hook to EOS' Jenkins CI kolibri-installer-android job successfully.  The kolibri-installer-android job builds and packages, then scheduled kolibri-installer-android-upload job as expected.

Issue #592, opened 2023-06-12
Download additional channel metadata in the background


dylanmccall:
In #548 / #580, we added a system where kolibri-explore-plugin downloads metadata for channels from _all_ content collections, instead of just channels from the initial content collection. At the moment, this is an extra stage in that initial content import which blocks the UI until it is finished. Instead, we would like for this additional metadata download to happen in the background while the user is browsing content.

wjt:
There are some related PRs for this that we should tie up. But this is still not the case - the additional collections are fetched as part of the onboarding flow. It's not too slow though – we don't include massive channels like Khan Academy! – and this way we don't have to deal with inconsistent state.

dbnicholson:
I whipped together something in https://github.com/endlessm/kolibri-explore-plugin/tree/background-downloads. It works pretty nicely but needs a little polish and I need to figure out the right way to ensure incomplete tasks are restarted when the app restarts.

dbnicholson:
I'm going to make a new release so this can be tested out in the apps.

dbnicholson:
This has been released as kolibri-explore-plugin 6.39.0 with Android Manatee 6.39-393 and Windows 6.39.1.0.

@vanessa-chang could you test out the Android app from internal testing?

vanessa-chang:
I’ve tested the new beta 6.39-393:
- the content pack download speed is back to normal 
-  the contents are revealed after all downloads complete (~app. 10mins of app running time) & it’s show.n after the page is refreshed by switching the tab or restarting the app
-  the app shows a black screen when restarting the app during metadata download (still needs a further clarification)

I will report another ticket for the back screen issue once it’s clarified.

I prefer to have it fixed before we push to the production. Let me know if you have any suggestions. Thank you

dbnicholson:
> * the app shows a black screen when restarting the app during metadata download (still needs a further clarification)
> 
> I will report another ticket for the back screen issue once it’s clarified.

Please do. I have a feeling this is because the background downloads never actually stop when the app is closed. https://github.com/learningequality/kolibri/pull/11120 will help that a lot but hasn't been released yet.

vanessa-chang:
@dbnicholson thanks. I have open the ticket #769 for the issue. I currently report in the Kolibri-explore-plugin repo. However, as this issue cannot be reproduced in the Windows app, let me know if I should file in the other repo for a correction. Thanks.

PR #650, opened 2023-06-20
Download: Move extra channel metadata import to background

dbnicholson:
Like the unavailable content thumbnails, move the download of the extra channel metadata to the background instead of blocking the initial starter pack download on it.

I'm marking this as a draft for now since we need some frontend work so it can be notified when new channels appear.

Helps: #592

### DIFF
--- a/kolibri_explore_plugin/collectionviews.py
+++ b/kolibri_explore_plugin/collectionviews.py
@@ -50,8 +50,7 @@ COLLECTION_NAMES = ["0001"]
 
 PROGRESS_STEPS = {
     "importing": 0.1,
-    "downloading": 0.8,
-    "tagging": 0.9,
+    "downloading": 0.9,
     "completed": 1,
 }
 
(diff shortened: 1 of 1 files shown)


dbnicholson:
As discussed on Slack, this isn't going to work. The way this PR works is that the extra channel and thumbnail tasks are all created at once and put into a background queue. However, the thumbnail tasks can't be created until the channel metadata exists. There are 3 ways I can think to handle this:

1. Leave the extra channel download in the foreground. I.e., do nothing.
2. Download all the channels in the foreground immediately. This is the same as 1 except the channel metadata downloads aren't split into 2 stages.
3. Have the frontend keep driving the stages even after it's left the download page. This would kinda weird since every page would need to keep polling the download manager.
4. Have the backend drive the background stages by using a [`StorageHook`](https://github.com/learningequality/kolibri/blob/6a65dd1a45ba74595663ebad4ad062cf52dc5515/kolibri/core/tasks/hooks.py#L8) that's notified when tasks complete. That would allow it to do something similar to how the foreground tasks are handled. I.e., a `background_tasks_pending` list that needs to be exhausted before moving on to the next background stage. Or you could go all the way and use this hook to drive all the stages and then the frontend polling would purely return status without also trying to manage the state machine.

dbnicholson:
I'm closing this since it's not going to ever work this way.

PR #652, opened 2023-06-21
GitHub issue template: Apply tag

manuq:
Add custom presentation tag to channel custom presentation requests.


manuq:
This is untested but it goes with github documentation. I hope it handles the space in the label correctly.

Adding this tag because now we use github issues for tracking everything and the list is becoming large.

PR #653, opened 2023-06-21
footer: Add app install links for Play and Microsoft stores

pwithnall:
These will allow the user to install the Endless Key app, if that would be a better fit for them on their device than using key.endlessos.org in a browser.

The badges came from:
 - https://play.google.com/intl/en_us/badges/
 - https://apps.microsoft.com/store/app-badge

Getting them the same size was a complete pain because, for no good reason, Google have baked a transparent border into the image they serve to you. We have to use the image they serve because otherwise we’d have to carry 20 different localised versions.

Changing the layout to match the design
(https://github.com/endlessm/endless-key-content-private/issues/64#issuecomment-1592442043) required splitting the footer bar into two separate containers, to allow for grouping the links/buttons at the start and end of the footer row.

https://github.com/endlessm/endless-key-content-private/issues/64

### DIFF
--- a/kolibri_explore_plugin/assets/src/components/AboutFooter.vue
+++ b/kolibri_explore_plugin/assets/src/components/AboutFooter.vue
@@ -2,25 +2,65 @@
 
   <b-container class="mb-5">
 
-    <b-row class="d-flex justify-content-center" cols="3">
+    <b-row
+      :class="{
+        'd-flex': true,
+        'justify-content-between': showStoreButtons,
+        'justify-content-around': !showStoreButtons,
+      }"
+    >
 
-      <b-col class="col-auto d-flex justify-content-center">
-        <button
-          v-b-modal.about-modal
-          class="btn d-md-block d-none shadow-none"
-        >
-          {{ $tr('aboutLabel') }}
-        </button>
+      <b-col class="col-auto d-flex justify-content-start">
+        <b-container class="h-100" fillHeight>
+          <b-row class="d-flex h-100 justify-content-center">
+
+            <b-col class="col-auto d-flex justify-content-start">
+              <button
+                v-b-modal.about-modal
+                class="btn d-md-block d-none shadow-none"
+              >
+                {{ $tr('aboutLabel') }}
+              </button>
+            </b-col>
+
+            <b-col class="col-auto d-flex justify-content-end">
+              <button
+                v-b-modal.about-modal
+                class="btn d-md-block d-none shadow-none"
+                @click="$root.$emit('setAboutSection', 'privacy-policy-link')"
+              >
+                {{ $tr('privacyPolicyLabel') }}
+              </button>
+            </b-col>
+
+          </b-row>
+        </b-container>
       </b-col>
 
-      <b-col class="col-auto d-flex justify-content-center">
-        <button
-          v-b-modal.about-modal
-          class="btn d-md-block d-none shadow-none"
-          @click="$root.$emit('setAboutSection', 'privacy-policy-link')"
-        >
-          {{ $tr('privacyPolicyLabel') }}
-        </button>
+      <b-col v-if="showStoreButtons" class="col-auto d-flex justify-content-end">
+        <b-container class="h-100" fillHeight>
+          <b-row class="d-flex h-100 justify-content-center">
+
+            <b-col class="col-auto d-flex justify-content-start">
+              <a
+                :href="googleStoreUrl"
+                target="_blank"
+                class="google-play-button"
+              >
+                <img :src="googlePlayImage" :alt="$tr('getOnGooglePlay')">
+              </a>
+            </b-col>
+
+            <b-col class="col-auto d-flex justify-content-end">
+              <script type="module" src="https://get.microsoft.com/badge/ms-store-badge.bundled.js"></script>
+              <ms-store-badge
+                :productid="windowsApplicationId"
+                cid="ek-online"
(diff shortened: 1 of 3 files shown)


kolibri_explore_plugin/assets/src/components/AboutFooter.vue
```diff
@@ -2,25 +2,61 @@
...
 
-      <b-col class="col-auto d-flex justify-content-center">
-        <button
-          v-b-modal.about-modal
-          class="btn d-md-block d-none shadow-none"
-          @click="$root.$emit('setAboutSection', 'privacy-policy-link')"
-        >
-          {{ $tr('privacyPolicyLabel') }}
-        </button>
+      <b-col class="col-auto d-flex justify-content-end">
```
manuq, line 34: According to the requirements https://github.com/endlessm/endless-key-content-private/issues/64 this is only needed for the online instance. So this `<b-col>` should be added conditionally with a `v-if` https://vuejs.org/guide/essentials/conditional.html#v-if .

Currently there is no "is online instance" option in kolibri-explore-plugin. The only exception we do for online instance is the HIDE_DISCOVERY_TAB option. So either:

* add a new IS_ONLINE_INSTANCE option and derive "hide discovery tab" + "display Play/Microsoft buttons" from it, or
* add a new DISPLAY_STORE_BUTTONS option, to be set in the online instance only.

manuq (changes requested):

kolibri_explore_plugin/assets/src/components/AboutFooter.vue
pwithnall: Oops, I forgot about that requirement while diving into the vue rabbit hole

I’ve pushed a fixup which implements a third option: it uses the [app IDs configured on the PWA plugin](https://github.com/learningequality/kolibri/pull/10883) to determine whether to show the badges.

The rest of the changes in the fixup are a load of vue/bootstrap hackery to get things centred properly regardless of whether the badges are shown. There’s probably a better way of doing this.

Badges:
buttons


No badges:
no buttons

kolibri_explore_plugin/kolibri_plugin.py
```diff
@@ -56,6 +56,14 @@ def plugin_data(self):
             ],
             "hideDiscoveryTab": conf.OPTIONS["Explore"]["HIDE_DISCOVERY_TAB"],
             "feedbackUrl": conf.OPTIONS["Explore"]["FEEDBACK_URL"],
+            # Use the app IDs from the PWA plugin, which advertises them as
+            # related applications.
+            "androidApplicationId": conf.OPTIONS["Pwa"][
+                "ANDROID_APPLICATION_ID"
+            ],
+            "windowsApplicationId": conf.OPTIONS["Pwa"][
+                "WINDOWS_APPLICATION_ID"
```
manuq, line 65: Ah, this is the first time that a setting from another plugin is referenced in the kolibri-explore-plugin. This is what happens when the PWA plugin is not installed. Example from my local environment but it will happen in Android/Windows apps too:

Captura de pantalla de 2023-06-21 16-59-16

manuq (changes requested):

manuq:
@pwithnall I have pushed a fixup that safeguards the KeyError in https://github.com/endlessm/kolibri-explore-plugin/pull/653#discussion_r1237619220 . But all in all, I'm not sure about this kind of cross-plugin reference. Any thoughts @jprvita @dylanmccall @dbnicholson ? Maybe I'm overthinking and this is OK as we manage both plugins.

dbnicholson:
> @pwithnall I have pushed a fixup that safeguards the KeyError in [#653 (comment)](https://github.com/endlessm/kolibri-explore-plugin/pull/653#discussion_r1237619220) . But all in all, I'm not sure about this kind of cross-plugin reference. Any thoughts @jprvita @dylanmccall @dbnicholson ? Maybe I'm overthinking and this is OK as we manage both plugins.

It's a bit unusual, but I don't think it's that bad. Ultimately the explore plugin is the one that's creating the presentation, so it does need to be aware of other plugins. I think it would be a little cleaner if the pwa plugin was the one that exported the options to `plugin_data`, though. Perhaps that could be added in https://github.com/learningequality/kolibri/pull/10883.

pwithnall:
> I think it would be a little cleaner if the pwa plugin was the one that exported the options to `plugin_data`, though. Perhaps that could be added in [learningequality/kolibri#10883](https://github.com/learningequality/kolibri/pull/10883).

I’m probably missing something, but I don’t think that’s possible.

Looking at [the implementation](https://github.com/endlessm/kolibri/blob/release-v0.14.x/kolibri/core/webpack/hooks.py#L298), `plugin_data` is indexed by the generating bundle ID on the webpack compilation side, and [automatically deindexed](https://github.com/endlessm/kolibri/blob/release-v0.14.x/kolibri/core/assets/src/utils/plugin_data_src.js) by the client-side module name. So if it were to be exported in`def plugin_data()` by the PWA plugin, the keys would end up in `global['kolibriPluginDataGlobal']['kolibri.plugins.pwa.$something']`, whereas the kolibri-explore-plugin client code would be looking for them in `global['kolibriPluginDataGlobal']['kolibri_explore_plugin.app']`. I don’t think there’s a neat way to index back up the dict.

manuq:
@pwithnall sounds good then. Feel free to squash and merge. Thanks for the feedback @dbnicholson!

PR #654, opened 2023-06-21
Fix loading

### LINKED ISSUE #639: "Something Went wrong" error is always shown after download completes
"Something went wrong" error is always shown after download completes.

This issue was seen in v6.9-340 and the latest package v6.14.345

Screenshot 2023-06-14 5 29 41 PM

I'm seeing this issue today. (test package: v6.9-340)
1. After content completes the download, it shows a "Something went wrong!" message.
2. Press "try again" button does not help.
3. Close and reopen the app, it will then running into this issue.

It fails twice with the explorer pack download for me today. (failure rate: 2/2)
It maybe worth to give it a try to see if you can reproduce the issue. I will try downloading the other content pack too.

_Originally posted by @vanessa-chang in https://github.com/endlessm/kolibri-explore-plugin/issues/598#issuecomment-1590846581_

manuq:
Fixes #639

### DIFF
--- a/kolibri_explore_plugin/assets/src/modules/topicsRoot/handlers.js
+++ b/kolibri_explore_plugin/assets/src/modules/topicsRoot/handlers.js
@@ -185,7 +185,6 @@ export function decideDownload(store) {
             store.commit('CORE_SET_PAGE_LOADING', false);
             store.commit('CORE_SET_ERROR', null);
           }
-          location.reload();
         });
       }
     }
--- a/kolibri_explore_plugin/assets/src/views/ContentUnavailablePage.vue
+++ b/kolibri_explore_plugin/assets/src/views/ContentUnavailablePage.vue
@@ -40,7 +40,7 @@
     computed: {
       homePageLink() {
         return {
-          name: PageNames.TOPICS_ROOT,
+          name: PageNames.ROOT,
         };
       },
     },
--- a/kolibri_explore_plugin/assets/src/views/DiscoveryPageContenPacks.vue
+++ b/kolibri_explore_plugin/assets/src/views/DiscoveryPageContenPacks.vue
@@ -116,6 +116,11 @@
         }
       },
     },
+    mounted() {
+      if (!this.loading) {
+        return this.fetchHighlighted();
+      }
+    },
     methods: {
       fetchHighlighted() {
         this.loadingNodes = true;
--- a/kolibri_explore_plugin/assets/src/views/DownloadPage.vue
+++ b/kolibri_explore_plugin/assets/src/views/DownloadPage.vue
@@ -36,7 +36,7 @@
               </b-button>
             </template>
             <template v-else-if="isCompleted">
-              <b-button pill variant="primary" :to="homePageLink">
+              <b-button pill variant="primary" @click="onConfirm">
                 {{ $tr('confirmLabel') }}
               </b-button>
             </template>
@@ -54,7 +54,6 @@
   import client from 'kolibri.client';
   import urls from 'kolibri.urls';
   import { mapState } from 'vuex';
-  import { PageNames } from '../constants';
   import { getCollectionInfo } from '../modules/coreExplore/utils';
 
   const UPDATE_DELAY = 1500;
@@ -89,11 +88,6 @@
       },
       gotDownloadError() {
         return this.errorMessage != null;
-      },
-      homePageLink() {
-        return {
-          name: PageNames.TOPICS_ROOT,
-        };
       },
       titleLabel() {
         if (this.isDownloading) {
@@ -145,6 +139,9 @@
         this.errorMessage = null;
         this.retry = true;
         return this.setUpdateInterval();
+      },
+      onConfirm() {
+        this.$router.go(0);
       },
       updateLoop() {
         return this.updateDownload()


jprvita (approved):

GeorgesStavracas:
I was going to comment the following:

I understood commits 1, 2, and 4.

I don't understand commit 3, but the new confirmation method seems to work.

But I guess I was too late :)

dbnicholson:
I'm not sure the "reload on confirm" commit really is a workaround. With the old code, I can see it making a request to `/api/content/channel`, which means it's not immediately returning the in memory cached data. In other words, `force=true` is doing what it's supposed to do and making the actual request to the backend.

However, the `contentCacheKey` query parameter is unchanged and I believe the backend is itself returning cached data instead of doing the actual database query. The reason it's unchanged is because `contentCacheKey` is defined statically in a `<script>` node in the HTML (that's the only way for the backend to actively send data to the frontend, anyways). So, without actually reloading the page, the HTML won't be regenerated and you'll therefore use the same `contentCacheKey` for every request.

That said, my understanding of the various kolibri caching layers may not be accurate.

jprvita:
Shall we open a follow-up ticket to look more deeply into the points above?

dbnicholson:
> Shall we open a follow-up ticket to look more deeply into the points above?

I think we should since I suspect we'll keep having an issue as the content is dynamically updated. This was really bugging me, so I did some testing with debugging. The issue is definitely that since the request is made with the same `contentCacheKey` query parameter, the backend responds with cached data. When the request is made with the original `contentCacheKey` from when the page was originally loaded, you get an empty list of channels. I copied the request out from the chrome devtools with the convenient "Copy as cURL" then tweaked the `contentCacheKey` parameter to see the difference:

```
$ curl -sI 'http://127.0.0.1:8080/api/content/channel/?contentCacheKey=1687449006.893226'   -H 'Accept: application/json, text/plain, */*'   -H 'Accept-Language: en-US,en;q=0.9'   -H 'Cache-Control: no-cache'   -H 'Connection: keep-alive'   -H 'Cookie: visitor_id=b8ed88e0adca4b45af936cb3b9fdb65b; kolibri_csrftoken=hSMfjcxMjGqdWIFHX23dkHCKRVuro7TsCrgelHOGXsykHoZVB6wNIMKQ93741bE6; kolibri=op4hywf4mmsciokxzt3ag07ueoqlz12d'   -H 'Pragma: no-cache'   -H 'Referer: http://127.0.0.1:8080/en/explore/'   -H 'Sec-Fetch-Dest: empty'   -H 'Sec-Fetch-Mode: cors'   -H 'Sec-Fetch-Site: same-origin'   -H 'User-Agent: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/114.0.0.0 Safari/537.36'   -H 'X-CSRFToken: hSMfjcxMjGqdWIFHX23dkHCKRVuro7TsCrgelHOGXsykHoZVB6wNIMKQ93741bE6'   -H 'X-Requested-With: XMLHttpRequest'   -H 'sec-ch-ua: "Not.A/Brand";v="8", "Chromium";v="114", "Google Chrome";v="114"'   -H 'sec-ch-ua-mobile: ?0'   -H 'sec-ch-ua-platform: "Linux"'   --compressed | grep -i content-length
Content-Length: 2
$ curl -sI 'http://127.0.0.1:8080/api/content/channel/?contentCacheKey=1687449006.893227'   -H 'Accept: application/json, text/plain, */*'   -H 'Accept-Language: en-US,en;q=0.9'   -H 'Cache-Control: no-cache'   -H 'Connection: keep-alive'   -H 'Cookie: visitor_id=b8ed88e0adca4b45af936cb3b9fdb65b; kolibri_csrftoken=hSMfjcxMjGqdWIFHX23dkHCKRVuro7TsCrgelHOGXsykHoZVB6wNIMKQ93741bE6; kolibri=op4hywf4mmsciokxzt3ag07ueoqlz12d'   -H 'Pragma: no-cache'   -H 'Referer: http://127.0.0.1:8080/en/explore/'   -H 'Sec-Fetch-Dest: empty'   -H 'Sec-Fetch-Mode: cors'   -H 'Sec-Fetch-Site: same-origin'   -H 'User-Agent: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/114.0.0.0 Safari/537.36'   -H 'X-CSRFToken: hSMfjcxMjGqdWIFHX23dkHCKRVuro7TsCrgelHOGXsykHoZVB6wNIMKQ93741bE6'   -H 'X-Requested-With: XMLHttpRequest'   -H 'sec-ch-ua: "Not.A/Brand";v="8", "Chromium";v="114", "Google Chrome";v="114"'   -H 'sec-ch-ua-mobile: ?0'   -H 'sec-ch-ua-platform: "Linux"'   --compressed | grep -i content-length
Content-Length: 991887
$ curl -sI 'http://127.0.0.1:8080/api/content/channel/'   -H 'Accept: application/json, text/plain, */*'   -H 'Accept-Language: en-US,en;q=0.9'   -H 'Cache-Control: no-cache'   -H 'Connection: keep-alive'   -H 'Cookie: visitor_id=b8ed88e0adca4b45af936cb3b9fdb65b; kolibri_csrftoken=hSMfjcxMjGqdWIFHX23dkHCKRVuro7TsCrgelHOGXsykHoZVB6wNIMKQ93741bE6; kolibri=op4hywf4mmsciokxzt3ag07ueoqlz12d'   -H 'Pragma: no-cache'   -H 'Referer: http://127.0.0.1:8080/en/explore/'   -H 'Sec-Fetch-Dest: empty'   -H 'Sec-Fetch-Mode: cors'   -H 'Sec-Fetch-Site: same-origin'   -H 'User-Agent: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/114.0.0.0 Safari/537.36'   -H 'X-CSRFToken: hSMfjcxMjGqdWIFHX23dkHCKRVuro7TsCrgelHOGXsykHoZVB6wNIMKQ93741bE6'   -H 'X-Requested-With: XMLHttpRequest'   -H 'sec-ch-ua: "Not.A/Brand";v="8", "Chromium";v="114", "Google Chrome";v="114"'   -H 'sec-ch-ua-mobile: ?0'   -H 'sec-ch-ua-platform: "Linux"'   --compressed | grep -i content-length
Content-Length: 991887
```

Notice the first one has a `Content-Length` of 2 corresponding to a body of `[]`. In the second request I bumped the last digit of the `contentCacheKey` up 1 and then I large response as noted `Content-Length`. On the 3rd request I removed the `contentCacheKey` query parameter entirely and also got the expected response. To ensure I really understood what was going on, I was running kolibri with `--debug-database` so it would log the actual database queries. In the first query I could see that no database query was made, which implied that it was returning a response cached by the backend.

I don't really understand how the backend cache with `contentCacheKey` works (it seems like it only does some response header mangling), but I think a fix is that when `force` is `true` in the frontend, then the request is made without the `contentCacheKey` query parameter. Although, that really conflates the frontend and backend caches. Not to mention the browser/webview cache. `-ETOOMANYCACHES`

PR #655, opened 2023-06-21
CI: Fix version substitution in endless-key-app trigger

dbnicholson:
The node script passed to the `github_script` action needs to have the `VERSION` value derived from the `env` context rather than expecting it to be resolved like a shell variable.

https://phabricator.endlessm.com/T34806

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -142,6 +142,6 @@ jobs:
               repo: "endless-key-app",
               event_type: "kolibri-explore-plugin-release",
               client_payload: {
-                VERSION: "$VERSION",
+                VERSION: "${{ env.VERSION }}",
               }
             })


wjt (approved):

.github/workflows/main.yml
```diff
@@ -142,6 +142,6 @@ jobs:
               repo: "endless-key-app",
               event_type: "kolibri-explore-plugin-release",
               client_payload: {
-                VERSION: "$VERSION",
+                VERSION: "${{ env.VERSION }}",
```
wjt, line 145: I would guess without proof that this would work:

```suggestion
                VERSION: env.VERSION,
```

wjt: Or maybe not. Merging what you have!

dbnicholson: I'm not really sure. I think that would imply that the substitution would happen within the action in node. I don't know if `env` actually corresponds to environment variables in that context. I think directly substituting from the context variable with the expression syntax is more likely to work, though.

starnight: @dbnicholson you got it!  Thanks!

Issue #637, opened 2023-06-16
Channel override: PhET Interactive Simulations (es) 


jofilizola:
Channel ID: Pending...

Styling:
- [ ] Custom primary color?: `#136D91`
- [ ] Custom secondary color? `#136D91`
- [ ] Custom header color? `#613863` 
- [ ] Custom thumbnail? https://www.dropbox.com/sh/t24n9xqjm0jciab/AAC8o7Vg-uwo5Zw1i1TuZgvXa?dl=0
- [ ] Custom icon? https://www.dropbox.com/sh/t24n9xqjm0jciab/AAC8o7Vg-uwo5Zw1i1TuZgvXa?dl=0

Options:
- [ ] Has carousel section? NO
- [ ] Has dark header? YES
- [ ] Display logo in header? YES

![Uploading image.png…]()

hydrolet:
PhET (es) channel token is: gobaz-fakin

I'm not sure how to find the Channel ID but here is the URL when viewing in Kolibri Studio. I'm assuming it is one of the long numbers at the end? 

https://studio.learningequality.org/en/channels/8fa678af1dd05329bf3218c549b84996/#/d3ed15b53a6042a5affb2db979491d9b

jofilizola:
@hydrolet thanks! Comparing the URLs from studio with the online instace channel ID bit, I think it's  this bit: 
https://studio.learningequality.org/en/channels/ **8fa678af1dd05329bf3218c549b84996**/ #/d3ed15b53a6042a5affb2db979491d9b

jofilizola:
@manuq the English and Spanish versions can share the presentation

manuq:
@jofilizola from now on when reusing a presentation you can remove all the placeholder text (except ID) and say "like that other channel".

jofilizola:
The English Channel as I can see on the online instance has the wrong background color, since they are reusing the presentation I imagine both have the same problem.

 Custom header color should be #613863 for: 

- [ ]  197934f144305350b5820c7c4dd8e194
- [ ]  8fa678af1dd05329bf3218c549b84996

![Uploading Screenshot 2023-07-21 at 11.02.39.png…]()

erikos:
@jofilizola that is 
![Screenshot from 2023-07-24 09-45-45](https://github.com/endlessm/kolibri-explore-plugin/assets/3714785/03641dfe-2c4a-4fd4-9187-bbec48ee4451)
[...]

PR #664, opened 2023-06-24
Fix custom presentation for PhET

erikos:
Add custom icon and follow the design for the presentation.

follow up for #637


kolibri_explore_plugin/assets/src/customApps.js
```diff
@@ -295,6 +295,7 @@ export const ChannelCardIcons = {
   '08897e003ea9489eb3d86fc94ba08c21': 'support-ukraine.png',
   '608fd926be60462aba2feccba6c050c9': 'harper-collins.png',
   '6da97a60cbd44e918137827bd5425efb': 'code-org.png',
+  '8fa678af1dd05329bf3218c549b84996': 'phet-interactive-simulations.png', // phet spanish
```
jprvita, line 298: Don't we want the same customization for the Part English as well?

jprvita (approved):
One small question about the consistency between the English and Spanish versions of Phet, but LGTM otherwise.

kolibri_explore_plugin/assets/src/customApps.js
erikos: Yes, for some reason I could not figure out the English channel has the right channel icon. I could make it explicit, I guess?

jprvita: Ok, IIUC we do NOT need the same customization for Phet English. In that case, this LGTM.

jprvita (approved):

PR #668, opened 2023-06-25
Add title attribute for clamped text

dylanmccall:
For text that may be ellipsized, include a "title" attribute with the full text so it appears as a tooltip in most web browsers.


manuq:
@dylanmccall @jprvita maybe we should move away of this vue-clamp dependency. I opened https://github.com/endlessm/kolibri-explore-plugin/issues/676 as follow-up.

PR #673, opened 2023-06-25
PackSelection: Reinstate white background

GeorgesStavracas:
PackSelection has an EkSlidableGrid, not an EkChannelCard, so removing the 'hasWhiteBackground' property was a mistake.

Reinstate that.


GeorgesStavracas:
GitHub forbids me to approve this, but a Canadian bird was happy with this fix!

PR #674, opened 2023-06-25
Remove extra space in Support link

dylanmccall:


dylanmccall:
Completely random, but this probably is a good starting point to learn some context around whitespace with (our version of) VueJS. Might be we should check on our build options? I don't see a reason why we _wouldn't_ want to strip whitespace.

GeorgesStavracas (approved):

manuq:
Nice, this whitespace bugged me a lot as well. One day I tried to fix it exactly like you did but the code formatting was silly reverting it. Great to know about the upstream issue.

PR #675, opened 2023-06-26
Add labels for missing media types

### LINKED ISSUE #560: Fix Play button for HP5 content
Seen in playthrough yesterday. The H5P content doesn't match any media type so the label is empty. They should look like HTML ones. Label: "interact".

Image

manuq:
Adapt to mediatypes missing from Kolibri like 'h5p'.

Resolves #560

### DIFF
--- a/packages/ek-components/src/components/EkPlayButton.vue
+++ b/packages/ek-components/src/components/EkPlayButton.vue
@@ -77,6 +77,9 @@ export default {
           return 'ExerciseIcon';
         case 'html5':
         case 'zim':
+        case 'slideshow':
+        case 'h5p':
+        case 'quiz':
           return 'Html5Icon';
         case 'bundle':
           return 'BundleIcon';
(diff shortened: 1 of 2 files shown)


manuq:
This is the query I did in Kolibri's `kolibri manage shell` to obtain all the media types:
``` python
In [1]: from kolibri.core.content.models import ContentNode

In [2]: k = next(f for f in ContentNode._meta.fields if f.attname == 'kind')

In [3]: k.choices
Out[3]: 
(('topic', 'Topic'),
 ('video', 'Video'),
 ('audio', 'Audio'),
 ('exercise', 'Exercise'),
 ('document', 'Document'),
 ('html5', 'HTML5 App'),
 ('slideshow', 'Slideshow'),
 ('h5p', 'H5P'),
 ('zim', 'Zim'),
 ('quiz', 'Quiz'))

```

manuq:
> LGTM, confirmed by running the same kolibri command. I wonder if eventually it would be worth adding fallback code in the play button components, so that if no string or icon is available, it returns a generic icon and label. But this is a separate thought!

Good point, let me add that now that I'm near this code.

manuq:
@GeorgesStavracas ready for review again. I also used the same icon for the new media types as the other "interact" types use.

GeorgesStavracas (approved):
Thanks! LGTM, please feel free to merge whenever you're happy with it

PR #677, opened 2023-06-26
Improve spacing of Featured Channels section

GeorgesStavracas:
These margins and paddings are making Joana sad. Nobody likes sad Joana, so let's fix these margins and paddings.

| Before | After |
|-----|-----|
| image | image |

https://github.com/endlessm/kolibri-explore-plugin/issues/679

### DIFF
--- a/kolibri_explore_plugin/assets/src/views/DiscoveryPageContenPacks.vue
+++ b/kolibri_explore_plugin/assets/src/views/DiscoveryPageContenPacks.vue
@@ -14,17 +14,20 @@
               {{ $tr('channelLabel') }}
             </h5>
           </b-container>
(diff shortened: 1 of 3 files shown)


manuq:
This is how it looks like with some actual featured channels. If Joana is happy with this then feel free to merge!
Captura de pantalla de 2023-06-26 13-29-59

jprvita (approved):
To my eyes this fulfills the request that we discussed with Joana yesterday, and the implementation makes sense too (plus I see Manuel also approved).

But I agree we should take advantage of being in person and show the screenshot to Joana before merging.

jprvita:
@GeorgesStavracas do we have a ticket number for this work?

PR #678, opened 2023-06-26
Discovery page: Allow unavailable featured channels

manuq:
The featured channels in the collection may be unavailable because the collection is not importing any content from them. Still they should appear in the Featured row.

### DIFF
--- a/kolibri_explore_plugin/assets/src/views/DiscoveryPageContenPacks.vue
+++ b/kolibri_explore_plugin/assets/src/views/DiscoveryPageContenPacks.vue
@@ -137,6 +137,7 @@
 
         const featuredPromise = ContentNodeExtrasResource.fetchByExternalTag('featured-channel', {
           only_root_nodes: true,
+          no_available_filtering: true,
         }).then(({ data }) => {
           const nodes = data.map(n => {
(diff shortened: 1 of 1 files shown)


manuq:
While testing #677 I realize that this regressed, probably when we removed the "NAVIGATE_UNAVAILABLE" feature flag.

jprvita (approved):
The change looks correct to me, however IDK how to trigger the original problem, so I could not test the fix.

I think I need to have a channel that is featured but have no content for it, how can I do this?

manuq:
> The change looks correct to me, however IDK how to trigger the original problem, so I could not test the fix.
> 
> I think I need to have a channel that is featured but have no content for it, how can I do this?

Yes, right. For example you can import a channel that you don't have with:
```
kolibri manage importchannel network $CHANNEL_ID
```

Then tag the root node (which has same ID as the channel) with:
```
kolibri manage applyexternaltags --tags featured-channel $CHANNEL_ID
```

jprvita:
Thanks! I was able to test this after following your instructions plus disabling cache in the inspector (and @dylanmccall's teachings that the "disable cache" setting in the inspector only works while the inspector is open!).

PR #680, opened 2023-06-26
New clamp

### LINKED ISSUE #676: Revisit line clamp
We have a dependency [vue-clamp](https://vue-clamp.vercel.app/) for it. Modern CSS already has it: https://caniuse.com/?search=line-clamp but `vue-clamp` doesn't fallback to this CSS at all. Instead it has a resize observer and does calculations that may be affecting performance.

Kolibri is also doing this with CSS in a TextTruncatorCss component that we should be using directly or borrowing: https://github.com/learningequality/kolibri/blob/develop/kolibri/core/assets/src/views/TextTruncatorCss.vue

manuq:
Resolves #676

### DIFF
--- a/packages/ek-components/package.json
+++ b/packages/ek-components/package.json
@@ -16,7 +16,6 @@
     "core-js": "^3.6.5",
     "lodash": "^4.17.21",
     "vue": "^2.6.11",
-    "vue-clamp": "^0.3.1",
     "vue-material-design-icons": "^4.12.1"
   },
   "devDependencies": {
--- a/packages/ek-components/src/components/EkCardBody.vue
+++ b/packages/ek-components/src/components/EkCardBody.vue
@@ -18,26 +18,20 @@
         class="align-items-center mb-1 title"
         :class="{ 'text-muted': !node.available }"
       >
-        <VClamp
-          autoresize
+        <EkClamp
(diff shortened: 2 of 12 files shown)


jprvita (approved):
I don't completely follow all the logic in EkClamp, but the replacement look correct and it seems to be working as expected.

manuq:
> I don't completely follow all the logic in EkClamp, but the replacement look correct and it seems to be working as expected.

It basically adds different CSS rules depending on conditions:
- If maxLine is 1
- Else, if CSS `line-clamp` rule is supported
- Else, a fallback for Internet Explorer and older versions of other browsers

jprvita:
> > I don't completely follow all the logic in EkClamp, but the replacement look correct and it seems to be working as expected.
> 
> It basically adds different CSS rules depending on conditions:
> 
> * If maxLine is 1
> * Else, if CSS `line-clamp` rule is supported
> * Else, a fallback for Internet Explorer and older versions of other browsers

Thanks! I also did some side-by-side review with @dylanmccall

manuq:
> Thanks! I also did some side-by-side review with @dylanmccall

Cool, I forgot to mention that i have added samples to the showcase "Test" page of template-ui to verify it didn't regress anything.

PR #682, opened 2023-06-26
Channel header: Safeguard empty description

manuq:
Issue #676


manuq:
Follow-up of the new EkClamp. I have found one case where the topic description was undefined (Channel STEAM topic Chemistry).

manuq:
In fact we have other cases. It's better to make the text attribute not required.

jprvita (approved):

jprvita (approved):

PR #685, opened 2023-06-27
New carousel cards

manuq:


GeorgesStavracas:
I've found this situation when testing this branch:

Captura de tela de 2023-06-27 16-34-24

Joana hasn't seen this yet, but I suspect the desired behavior in this case is to left align the grid instead of center it.

GeorgesStavracas (commented):
The patch series is fantastic, it was super easy to review. Thank you for that.

I had to learn what `<b-container fluid>`  means, and AFAIU one would pass `fluid` to make the container have 100% of the width.

The only thing that I found was the centralized grid I described above. Now that I'm testing it more, I wonder if it should scroll when scrolling horizontally with touchpads too.

manuq:
@GeorgesStavracas thanks for testing! You are absolutely right about the bad centering. I realize that we don't need a carousel component at all if there is no pagination (less-or-equal than 4 cards in a big screen). So I appended a commit that uses a simple (responsive) grid in that case.

packages/ek-components/src/main.js
```diff
@@ -62,6 +64,8 @@ const components = {
   EkSearchBar,
   EkSlidableCardGrid,
   EkSlidableGrid,
+  EkSlidableGridNew,
+  EkSlidableCardGridNew,
```
jprvita, line 68: Can we choose a better name for these components? `EkSlidableGrid` vs `EkSlidableGridNew` has zero meaning for someone who was not around when they were created. I tried to think of some suggestions and the best I could come up with was `EkInertialSlidableGrid`, but I'm not sure how good that is.

jprvita (changes requested):
I did not have time to test this yet, but read through the code. It LGTM, with a small nitpick + bikeshedding opportunity on the name of the new component.

manuq:
> I did not have time to test this yet, but read through the code. It LGTM, with a small nitpick + bikeshedding opportunity on the name of the new component.

The new one is meant to replace the current one. I don't see any reason to keep the old component once we are happy with the new implementation. My idea was to temporarely keep it as a separate component (suffixed "new") for comparing them side-by-side.

jprvita:
> > I did not have time to test this yet, but read through the code. It LGTM, with a small nitpick + bikeshedding opportunity on the name of the new component.
> 
> The new one is meant to replace the current one. I don't see any reason to keep the old component once we are happy with the new implementation. My idea was to temporarely keep it as a separate component (suffixed "new") for comparing them side-by-side.

I thought we would want to keep both components. If we are replacing the old one immediately, the `new` suffix works, but if we are taking more than a couple of days I would still use a more meaningful name, just in case we end up switching priorities and leaving this potential for confusion behind for longer than we wanted (aka tech debt).

Issue #684, opened 2023-06-27
Add feedback to content node Download buttons


dylanmccall:
The download button for individual content nodes should show easily visible feedback when it is hovered over and clicked. To do this, we would like to add a label indicating its status, a hover state, and (possibly) a fun animation on click.

manuq:
Reviewed: https://github.com/endlessm/kolibri-explore-plugin/pull/689#pullrequestreview-1507335444

PR #689, opened 2023-06-29
Add labels and hover feedback for content download buttons

dylanmccall:
Closes https://github.com/endlessm/kolibri-explore-plugin/issues/684

### DIFF
--- a/packages/ek-components/src/components/EkCardButtons.vue
+++ b/packages/ek-components/src/components/EkCardButtons.vue
@@ -15,6 +15,9 @@
       :class="downloadButtonClasses"
       @click="$emit('downloadButtonClicked')"
     >
+      <span v-if="downloadText" class="download-button-text">
+        {{ downloadText }}
+      </span>
       <DownloadIcon
         v-if="isDownloadReady"
         :title="$tr('downloadLabel')"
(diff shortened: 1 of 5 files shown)


packages/ek-components/src/components/EkPlayButton.vue
```diff
@@ -114,7 +114,7 @@ export default {
 }
 
 button:disabled {
-  opacity: 33%;
+  opacity: 0.5;
```
manuq, line 117: :+1:

packages/ek-components/src/components/EkCardButtons.vue
```diff
@@ -127,6 +139,14 @@ export default {
 <style lang="scss" scoped>
 @import '../styles.scss';
 
+.card-buttons {
+  &::after {
+    content: '';
+    display: block;
+    clear: both;
+  }
+}
```
manuq, line 148: Nit: this change has nothing to do with the commit message.

manuq (approved):
All looks good to me! Only a nit observed.

packages/ek-components/src/components/EkCardButtons.vue
dylanmccall: Oops, that should be in its own _issue_, even: Joana and I noticed the margins were inconsistent when we were looking at this. Good thing to keep track of properly :) I'll move that change elsewhere for orderliness and then merge this. Thank you for the review!

dylanmccall: I split this into https://github.com/endlessm/kolibri-explore-plugin/issues/690.

PR #156, opened 2021-07-02
Discovery page: Add theme debugging in the development info modal

manuq:
https://phabricator.endlessm.com/T32141

PR #693, opened 2023-07-05
EkPlayButton: Don’t use nullish coalescing operator

pwithnall:
My ChromeOS test machine has Android WebView 68, which doesn’t support this operator. That causes a syntax error when loading the kolibri-explore-plugin, and hence the app never loads.

Nowhere else in the code is the nullish coalescing operator used, so let’s get rid of this instance for now.

It’s apparently supported by WebView 80 onwards
(https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing#browser_compatibility). I don’t know if we want to depend on that yet.

Helps: #156

### DIFF
--- a/packages/ek-components/src/components/EkPlayButton.vue
+++ b/packages/ek-components/src/components/EkPlayButton.vue
@@ -100,7 +100,7 @@ export default {
       if (this.label) {
         return this.label;
       }
-      return MediaTypeVerbs[this.kind] ?? MediaTypeVerbs['video'];
+      return (MediaTypeVerbs[this.kind] !== null && MediaTypeVerbs[this.kind] !== undefined) ? MediaTypeVerbs[this.kind] : MediaTypeVerbs['video'];
     },
   },
 };


pwithnall:
One reason to merge this rather than saying I should update my WebView is that I *did* run all the system updates I could find before testing, and that still didn’t update the WebView. Specifically searching for WebView in the Play Store on the machine shows an ‘Update’ button, so I could potentially update it. I won’t do that for now, so that I can test the fix if this lands.

If doing all the system and app updates shown to me by Android didn’t update the WebView, I expect that can happen to other users, and hence we might want to be conservative about depending on newer WebView versions.

dylanmccall:
I'm okay with the quick fix, but also it's very strange that this is happening. This whole thing runs babel using a `browserslist` from here…

https://github.com/learningequality/kolibri/blob/develop/packages/browserslist-config-kolibri/index.js

… which _should_ be telling the appropriate pieces to please support `Android > 4.4.3` and `ChromeAndroid >= 49`. So Babel should be transforming the nullish coalescing operator to something else automatically. I wonder why it isn't?

dylanmccall:
Oh, but I see the template-ui packages do their own thing here: https://github.com/endlessm/kolibri-explore-plugin/tree/master/packages/template-ui. They don't appear to use `browserslist-config-kolibri` and instead have a more conventional [`> 1%, last 2 versions, not dead`](https://github.com/endlessm/kolibri-explore-plugin/blob/master/packages/template-ui/.browserslistrc). Is it possible this issue is specifically with channel apps, or is it affecting the Discover page itself as well?

dylanmccall (approved):

pwithnall:
> Is it possible this issue is specifically with channel apps, or is it affecting the Discover page itself as well?

I don't know: I could only reproduce it on the EK app on Android, and one of the symptoms was that it wouldn't successfully download a content pack, so I never got to test further into the app than initial download.

pwithnall:
I tested the apk with this fix landed, and it [hit another syntax error](https://github.com/endlessm/kolibri-installer-android/pull/157#issuecomment-1625306295), so it looks like we need [a more comprehensive fix](https://github.com/endlessm/kolibri-installer-android/issues/159).

Issue #694, opened 2023-07-05
Fix alignment of pills in Channel view


GeorgesStavracas:
The filter pills in the Channel view have their inner texts vertically misaligned. This is driving Joana mad, and nobody likes when our only designer is going crazy on alignment. Let's fix that.

![image](https://github.com/endlessm/kolibri-explore-plugin/assets/3518204/1bc99cf4-a859-49c5-95b9-d287bbc4cd3d)

erikos:
QA in Ladybird v6.30.0 (Vanessa and Simon) - pass from our side.

vanessa-chang:
@jofilizola Joana, what do you think about the new fix?  :)

wjt:
It also looks good to me [here](https://key.endlessos.org/en/explore/#/topics/1520f018610256549c98ca0140cceebe). I assume Joana will speak up if she disagrees :)

PR #695, opened 2023-07-05
MainSections: Center pill text

GeorgesStavracas:
This is mostly a workaround to how Bootstrap styles pill-like buttons by default. It only applies to the pills at channel views, which are the most visible ones.

https://github.com/endlessm/kolibri-explore-plugin/issues/694

---

| Comparison |
|---|
| Before: |
| image |
| After |
| image |


jprvita (approved):
LGTM. It took me a minute to understand where the `.rounded-pill` CSS class got applied to those buttons, but I found the answer in https://bootstrap-vue.org/docs/components/button#pill-style.

Issue #687, opened 2023-06-27
Content caching issues


dbnicholson:
This is a follow up to #654. In that PR a [workaround](https://github.com/endlessm/kolibri-explore-plugin/pull/654/commits/22dcee453f0993fb91b505afe903676a15e0bf5d) was added to ensure that the plugin didn't get stale content data. I spent some time looking at how caching works in Kolibri and wanted to track those issues here.

dbnicholson:
From the frontend perspective, there are 3 caches for content data:

1. An in memory cache on the JS resource [model](https://github.com/learningequality/kolibri/blob/develop/kolibri/core/assets/src/api-resource.js#L11) or [collection](https://github.com/learningequality/kolibri/blob/develop/kolibri/core/assets/src/api-resource.js#L261).
2. The browser or webview cache. When the backend responds with an appropriate `Cache-Control` header (as happens for content API responses), the browser will reuse the cached response as long as it hasn't expired.
3. A file backed [Django cache](https://docs.djangoproject.com/en/1.11/topics/cache/#the-per-site-cache) managed by the backend. This is opt-in per response view, but the content API [uses it](https://github.com/learningequality/kolibri/blob/develop/kolibri/core/content/api.py#L90).

Only after all 3 caches are bypassed does an actual database query take place.

The content API has a particularly unique setup. There is a backend [content cache key](https://github.com/learningequality/kolibri/blob/develop/kolibri/core/device/models.py#L215) that gets synchronized to the frontend in HTML `<head>` as `plugin_data.contentCacheKey` when a page is loaded. The frontend channel and content node resources [append the value as a query parameter](https://github.com/learningequality/kolibri/blob/develop/kolibri/core/assets/src/api-resources/utils.js#L6) when making API requests.

The purpose of this query parameter is to alter the URL so that the request bypasses the browser and backend caches since the caches are keyed by URL. This is particularly important for the content API since the backend responds with a `Cache-Control` setting `max-age` to 1 year. If the URL never changed, then the browser would continue to reuse its cached response essentially forever.

dbnicholson:
[...]

PR #696, opened 2023-07-05
Disable content caching

dbnicholson:
Add custom content API endpoints that don't use caching in the browser or server to ensure that fresh content is always returned. The content cache key as `ETag` is preserved so that the server can still return early when the browser includes the current content cache key in an `If-None-Match` header. I'm sure I'm "doing it wrong" in JS.

The last commit fixes a minor issue where upstream's channel import tasks don't update the content cache key if the channel doesn't have any nodes. See https://github.com/learningequality/kolibri/pull/10899.

I'm marking this as a draft since the commits need a bit of polish and probably more testing, but in some light testing it seems to fix the stale cached content data issues we've been hitting.

Fixes: #687

### DIFF
--- a/kolibri_explore_plugin/api_urls.py
+++ b/kolibri_explore_plugin/api_urls.py
@@ -11,11 +11,19 @@ from .collectionviews import resume_download
 from .collectionviews import start_download
 from .collectionviews import update_download
 from .viewsets import ContentNodeExtrasViewset
+from .viewsets import ExploreChannelMetadataViewSet
+from .viewsets import ExploreContentNodeViewset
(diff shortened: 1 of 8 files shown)


dbnicholson:
I tested this again and I think it's useful. I was trying to add actual tests, but that was turning into a whole thing since there's no existing setup here. I dropped the last commit since my upstream fix was [merged](https://github.com/learningequality/kolibri/pull/10899).

I'd like to get this in and let it simmer a bit before proposing it upstream.

dylanmccall (approved):
This all makes sense to me. Let's merge it! :)

PR #701, opened 2023-07-13
Pipfile: Update Pipfile.lock for dependent packages management

starnight:
The [CI](https://github.com/endlessm/kolibri-explore-plugin/actions/runs/5532800096/jobs/10095397076) is failed, because bumpversion finds the modified Pipfile.lock.  The modification comes from bootstrap scripts' [`pipenv install --dev`](https://github.com/endlessm/kolibri-explore-plugin/blob/04b5ceb8721a0eae315d613f589858a20019899f/scripts/bootstrap.sh#L5).

According to [Pipfile & Pipfile.lock](https://pipenv.pypa.io/en/latest/pipfile/), both Pipfile and Pipfile.lock should be in version control to secure dependent packages' version.

### DIFF
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "1e6e0a830ff90f06e4b32cb5d621564a0744065db62567a5410ccbab2afedf9a"
+            "sha256": "357b4e79b2a9fb0ac09b570d7513ebd1bcfc18492cf320343ab5bc227c43a99b"
         },
         "pipfile-spec": 6,
         "requires": {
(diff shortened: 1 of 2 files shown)


starnight:
This follows https://github.com/endlessm/kolibri-explore-plugin/pull/701
The Pipefile.lock should be updated along with the new Pipefile.

wjt:
Typo: "Pipefile" should be "Pipfile"

pwithnall:
This LGTM, I think it was caused because I forgot to update the lockfile last time I edited the Pipfile (probably due to rebasing that branch and hiding the sausage making).

I wonder if we should be calling `pipenv` with `--ignore-pipfile` in CI, though, so it always uses (and doesn’t regenerate) the lockfile. CI should be testing the changes we make in a PR, not those plus the changes coming from an arbitrary set of updated dependencies.

I’ll put together a PR for that.

pwithnall:
> I’ll put together a PR for that.

⇒ https://github.com/endlessm/kolibri-explore-plugin/pull/702

PR #702, opened 2023-07-13
ci: Add --ci argument to bootstrap.sh and use it from CI jobs

pwithnall:
Passing `--ci` will cause `--ignore-pipfile` to be passed to `pipenv install`, which will cause it to only use the `Pipfile.lock` for installing dependencies, and not install the latest versions which match `Pipfile`, or to update the `Pipfile.lock`. This means that the dependency versions used in CI are properly pinned, and hence we’re not testing them as well as whatever changes are in the associated PR at the same time.

`bootstrap.sh` unfortunately can’t be used from `node.yml` because it benefits from using `actions/setup-node` to cache the node installation.

Prompted by https://github.com/endlessm/kolibri-explore-plugin/pull/701

### DIFF
--- a/.github/workflows/bump2release.yml
+++ b/.github/workflows/bump2release.yml
@@ -52,7 +52,7 @@ jobs:
 
       - name: Install dependencies in pipenv
         run: |
-          sh ./scripts/bootstrap.sh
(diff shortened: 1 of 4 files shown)


starnight (approved):
Thanks!

Issue #703, opened 2023-07-14
Android & Windows apps do not start since build system rework


erikos:
![Screenshot 2023-07-14 16 26 04](https://github.com/endlessm/kolibri-explore-plugin/assets/3714785/06e6dae8-9182-4b42-809d-aee118727f9b)

Version should be Ladybird 6.27-366 - though the Play console shows:
![Screenshot from 2023-07-14 16-33-13](https://github.com/endlessm/kolibri-explore-plugin/assets/3714785/8adcc4c9-e571-4e7a-8f92-9f6e84ab7d24)

dylanmccall:
It looks like this regressed in https://github.com/endlessm/kolibri-explore-plugin/pull/662: our builds no longer include `kolibri_explore_plugin/build/kolibri_explore_plugin.app_stats.json`.

dylanmccall:
Oh oops, except we probably don't want to actually close this yet until we've verified it :)

erikos:
Kicked of a new release of the explore plugin https://github.com/endlessm/kolibri-explore-plugin/actions/runs/5561682701

wjt:
The app now sits forever at the Kolibri bird animation.

I connected to the server from a regular web browser and looked at the web console:

```
GET http://192.168.4.43:42819/static/kolibri_explore_plugin.app/kolibri_explore_plugin.app6.28.0.css net::ERR_ABORTED 404 (Not Found)
GET http://192.168.4.43:42819/static/kolibri_explore_plugin.app/kolibri_explore_plugin.app-6.28.0.js net::ERR_ABORTED 404 (Not Found)
```

I think this is likely to be the problem.

wjt:
(Those filenames look suspicious to me.)

wjt:
I see the same inspecting the app's “Kolibri” webview via `chrome://inspect#devices`.

wjt:
I think the problem is that the wheel used to include the compiled Javascript & CSS (but no source) and now it includes the source (but no compiled assets).

This would explain why the compiled JS and CSS is failing to load :-)

Before:
![Screenshot from 2023-07-18 09-42-24](https://github.com/endlessm/kolibri-explore-plugin/assets/86760/e15b9b98-2337-49db-b896-70e6519812a5)

After (notice that the `kolibri_explore_plugin.app` directory is completely missing):

[...]

PR #706, opened 2023-07-18
build: Include all the yarn-built app files in the dist wheel

pwithnall:
These were mistakenly omitted from commit
bf95c0021f55262c6196c052fcba78e9e945f567.

Fixes: #703


pwithnall:
I think this is necessary, but not sufficient, to fix #703.

dbnicholson (approved):
This is correct and I believe all that's needed.

Issue #728, opened 2023-08-03
Download content failed after update Kolibri to v0.16.0-beta0


starnight:
The download of Artist package is stuck at 10%
![10percent](https://github.com/endlessm/kolibri-explore-plugin/assets/2100946/bc16b09e-61ad-402e-8d5a-a2455c1c2cbd)
Seems the error: `AttributeError: 'FileDownload' object has no attribute 'run'`
```
[...]

PR #729, opened 2023-08-03
Use Kolibri's FileDownloader

starnight:
We tried to fix download speed issue with [Vendor old file download code](https://github.com/endlessm/kolibri-explore-plugin/pull/638).

However, Kolibri made some big change for `FileDownloader` since the commit ("[Simplify Transfer subclasses now that they don't need to expose iterators.](https://github.com/learningequality/kolibri/commit/c2909072710d18cd22f350806e0782e63acf6139)") included by [v0.16.0-alpha19(+)](https://github.com/learningequality/kolibri/releases/tag/v0.16.0-alpha19).  And, our fix conflicts to the change and leads issue [Download content failed after update Kolibri to v0.16.0-beta0](https://github.com/endlessm/kolibri-explore-plugin/issues/728).

I think this is the time using Kolibri's `FileDownloader` again.

Fixes: https://github.com/endlessm/kolibri-explore-plugin/issues/728


dbnicholson (approved):
It would be nice to verify that downloads are actually fixed, but I think this is for the best and any further regressions should probably be pursued upstream.

starnight:
Tested download speed of Artist package with Endless Key (Windows) 6.31.2.0 in Taipei office
* Start: 2023-08-07 13:41:22
* End: 2023-08-07 13:46:32

=> 5 minutes and 10 seconds -> Looks like around the speed as before.

Issue #730, opened 2023-08-04
Channel override change: TED Ed


jofilizola:
Channel ID: 2091ca47ff544c96b4ae02b3a92346e1
- [ ] Should not display as bundle
We should keep all other overrides.

**Following up this task:**

-  https://github.com/endlessm/endless-key-content-private/issues/86
We removed the quizzes of the TED Ed channel, since they were broken. With no quizzes there is no reason for the content to show as bundle, as the videos become hidden under a folder.

erikos:
The channel ID has not changed, so the changes from Heather in the Kolibri channel have been pulled in already. If you look at the channel presentation now, the following is visible:
* the carousel only has one item
* in the detail view the video card is centered


![Image](https://github.com/endlessm/kolibri-explore-plugin/assets/3714785/b47f2ce6-3dee-4913-aebb-5d63f33e4707)
![Image](https://github.com/endlessm/kolibri-explore-plugin/assets/3714785/b0dae825-5d45-47c3-936c-80ae6d5bb232)
![Image](https://github.com/endlessm/kolibri-explore-plugin/assets/3714785/f348c03b-6721-4d9c-a61a-8089a24583b3)

This should be because we are using the [bundle option](https://github.com/endlessm/kolibri-explore-plugin/blob/master/packages/template-ui/channel-overrides/ted-ed-lessons/options.json#L3) still in the config. Maybe we should also remove the carousel option for videos.

erikos:
Available in https://github.com/endlessm/kolibri-explore-plugin/releases/tag/v6.35.0 and in internal testing in the Android app.

erikos:
Looks good now.

PR #749, opened 2023-08-16
Channel override: adjust representation for the removal of bundles

erikos:
We simplified the TED-Ed channel to not have bundles anymore, just the videos. Remove the bundle specific custom presentation.

https://github.com/endlessm/kolibri-explore-plugin/issues/730


GeorgesStavracas (commented):
Seems enough, but wouldn't it be necessary to remove the `style.scss` file as well?

erikos:
> Seems enough, but wouldn't it be necessary to remove the `style.scss` file as well?

The https://github.com/endlessm/kolibri-explore-plugin/blob/master/packages/template-ui/channel-overrides/ted-ed-lessons/styles.scss still has some settings we want to keep.

Issue #742, opened 2023-08-09
Channel override: SAMHSA


jofilizola:
Channel ID: a8e6591f1afa426d859318a0a29d1237

Styling:
- [ ] Custom thumbnail? https://www.dropbox.com/scl/fo/iwk64bj9nd6r450w0h3ks/h?rlkey=mxai6bg4bvy0nj530n5ri24kx&dl=0
- [ ] Custom icon? https://www.dropbox.com/scl/fo/iwk64bj9nd6r450w0h3ks/h?rlkey=mxai6bg4bvy0nj530n5ri24kx&dl=0

erikos:
Follow up just pushed https://github.com/endlessm/kolibri-explore-plugin/commit/ec4c9dbff7056e1e5cc9df99548da83dcc391171

PR #752, opened 2023-08-16
Custom channel representation: add SAMSHA channel

erikos:
Only add a thumbnail and an icon for the SAMSHA channel.

https://github.com/endlessm/kolibri-explore-plugin/issues/742


kolibri_explore_plugin/assets/src/customApps.js
```diff
@@ -218,6 +218,7 @@ export const ChannelCardThumbnails = {
   '057f871caa405ec29d62ba0523c193d7': 'music.jpg',
   '914fee213ee146de869016c287116b23': 'novels.jpg',
   '59bb2e5a3d2e5e3b85b87d9ab4daa2f3': 'reading.jpg',
+  a8e6591f1afa426d859318a0a29d1237: 'samhsa.png',
```
GeorgesStavracas, line 221: I think the hash needs to be quoted, `'a8e6591f1afa426d859318a0a29d1237'`

kolibri_explore_plugin/assets/src/customApps.js
```diff
@@ -281,6 +282,7 @@ export const ChannelCardIcons = {
   '057f871caa405ec29d62ba0523c193d7': 'music.png',
   '914fee213ee146de869016c287116b23': 'novels.png',
   '59bb2e5a3d2e5e3b85b87d9ab4daa2f3': 'reading.png',
+  a8e6591f1afa426d859318a0a29d1237: 'samhsa.png',
```
GeorgesStavracas, line 285: I think the hash needs to be quoted, `'a8e6591f1afa426d859318a0a29d1237'`, here too

GeorgesStavracas (changes requested):

jprvita:
> I think the hash needs to be quoted

Would it be possible to check that in the lint phase of CI?

erikos:
> > Would it be possible to check that in the lint phase of CI?
> 
> Might be worthy, yes

Check the other occasions in the same file - ids starting with a character do not have to be quoted. I am sure Manuel has a nice story to go with, but I do not remember the reason.

PR #753, opened 2023-08-16
Fix bootstrap arguments passed to pipenv

dbnicholson:
If you don't specify any options to bootstrap.sh, `pipenv_install_args` will retain its default of an empty string and pipenv will be passed an empty quoted argument. That's wrong and some versions of pipenv fail expecting the argument to be a package requirement.

Instead, initialize the variable as an array and append to it as needed. Bash expands an empty array to nothing using the `"${var[@]}"` notation.

### DIFF
--- a/scripts/bootstrap.sh
+++ b/scripts/bootstrap.sh
@@ -3,14 +3,14 @@
 # Setup Project
 
 # Argument parsing
-pipenv_install_args=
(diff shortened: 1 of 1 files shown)


dylanmccall (approved):
Oh yeah, that makes sense :)

PR #755, opened 2023-08-17
Custom channel presentation: add carousel for SAMHSA channel

erikos:


erikos:
This is a follow up of https://github.com/endlessm/kolibri-explore-plugin/issues/742 based on a check-in with Joana and Heather.

GeorgesStavracas (approved):

Issue #756, opened 2023-08-17
Replace Welcome Screen Assets


jofilizola:
The assets of the welcome screen still have that old typo. These new ones are the same size and aspect ration, but no text: also localization-proof if we don't get to finish the refresh of this page.
![image 99](https://github.com/endlessm/kolibri-explore-plugin/assets/51095367/52ef68f2-3fff-4740-9780-923bfb78f124)
![image 100](https://github.com/endlessm/kolibri-explore-plugin/assets/51095367/870cefde-5a27-413b-a007-c3156db4064e)

jofilizola:
@jprvita

dylanmccall:
Hi @jofilizola! Just to be sure, is it intentional that these new assets are zoomed out a bit more than the originals? The old ones were cropped differently, like https://github.com/endlessm/kolibri-explore-plugin/blob/2bc544ad2fd733b74230bf2e635bc96320492232/packages/welcome-screen/src/assets/welcome-bg-bottom-left.jpg.

I added a comparison in https://github.com/endlessm/kolibri-explore-plugin/pull/757#issuecomment-1683158456.

jofilizola:
Ha! I think I exported the full assets and Manuel cropped them. Here they are, cropped the same.
![bottom-left-cropped](https://github.com/endlessm/kolibri-explore-plugin/assets/51095367/fb5b137b-bef5-4a01-986e-2a3fd8879df6)
![top-right-cropped](https://github.com/endlessm/kolibri-explore-plugin/assets/51095367/6ea94d7c-b53f-46d3-9be4-056edaa296d4)

jprvita:
Here how it looks with the new assets:
![Screenshot from 2023-08-18 11-07-35](https://github.com/endlessm/kolibri-explore-plugin/assets/73894/5a1ef750-b2eb-4ff4-81bd-abd766e6b732)

And the original ones, for comparison:
![Screenshot from 2023-08-18 11-15-13](https://github.com/endlessm/kolibri-explore-plugin/assets/73894/0bc0851f-424a-46a1-8872-03bae90045e4)

If you look very closely you can tell the new assets are a tiny bit more "zoomed out" (ex., the new "read aloud" tile has a smaller piece cut-off), but I don't think it matters or have any negative effect, so I'm marking the PR as ready for review + merging again.

jofilizola:
Looks good to me! Thank you, thank you!

jprvita:
Merged!

[...]

PR #757, opened 2023-08-17
welcome-screen: Update background assets

jprvita:
New assets provided by design, which should work well accross different languages.

https://github.com/endlessm/kolibri-explore-plugin/issues/756


jprvita:
I think this is all that is needed here. I'm not sure how to test it though, as I don't have a access to a Chromebook right now, and this welcome screen is skipped in the dev environment.

dylanmccall:
One convenient way to test this is with `lerna run serve --scope welcome-screen` from the project's base directory, which runs the thing at http://localhost:8080. I went ahead and did that, advancing to the welcome page by running `WelcomeApp.showWelcome() ` in the Javascript console, and this all looks good to me :)

For reference, here is the old one:

Screen Shot 2023-08-17 at 17 36 10

And here is the new one:

Screen Shot 2023-08-17 at 17 34 10

dylanmccall:
As you can see, these assets are a smidge different in shape than the originals. Let's double check in #756 if that is intentional, but I'll mark this approved for now since it's doing what it says :)

dylanmccall (approved):

jofilizola:
I've exported cropped assets! https://github.com/endlessm/kolibri-explore-plugin/issues/756#issuecomment-1683828013

jprvita:
> One convenient way to test this is with `lerna run serve --scope welcome-screen` from the project's base directory, which runs the thing at http://localhost:8080. I went ahead and did that, advancing to the welcome page by running `WelcomeApp.showWelcome() ` in the Javascript console, and this all looks good to me :)

Thanks! I had gotten close by running `yarn serve` from `packages/welcome-screen`, but was missing the `WelcomeApp.showWelcome()` bit. I'm guessing `lerna` is a different tool to do roughly the same thing.

Converting back to draft while I update with the new assets.

GeorgesStavracas (approved):

jprvita:
Updated with newer assets, see https://github.com/endlessm/kolibri-explore-plugin/issues/756#issuecomment-1683995730 for how it looks and more info.

PR #557, opened 2023-05-31
Channels presentation: Add overrides for HarperCollins

jprvita:
Fixes: https://github.com/endlessm/kolibri-explore-plugin/issues/555

PR #758, opened 2023-08-18
Custom channel presentation: add also empty styles file.

erikos:
Otherwise the override bundle gets not created apparently. See harper-collins, blockly-games and oceanx as other examples. Mentioned here https://github.com/endlessm/kolibri-explore-plugin/pull/557#issuecomment-1571882036

Follow up of https://github.com/endlessm/kolibri-explore-plugin/issues/742


pwithnall:
> See Wikipedia as another example.

Unless I’m looking in the wrong place, it looks like the [Wikipedia style file](https://github.com/endlessm/kolibri-explore-plugin/blob/master/packages/template-ui/channel-overrides/wikipedia/styles.scss) is not (and has never been) empty.

erikos:
> > See Wikipedia as another example.
> 
> Unless I’m looking in the wrong place, it looks like the [Wikipedia style file](https://github.com/endlessm/kolibri-explore-plugin/blob/master/packages/template-ui/channel-overrides/wikipedia/styles.scss) is not (and has never been) empty.
> 
> blockly-games and oceanx appear to have empty `styles.scss` files. So I guess this could land if the commit message is tweaked to mention one of them instead.

Thanks for taking a look. I mixed up things - while double checking I found that Harper Collins is also empty and here is a reference comment https://github.com/endlessm/kolibri-explore-plugin/pull/557#issuecomment-1571882036 - I will tweak the commit message.

pwithnall:
I updated the commit message

PR #762, opened 2023-08-22
Download extra channels in the background

dbnicholson:
This introduces a database model to keep track of the background download tasks to run after the initial download completes. That allows serializing the tasks to run one at a time with a `StorageHook` as the needed tasks are tracked in a persistent location. With that in place, we can defer downloading the extra channels until after the initial download completes. Together these changes have 2 major benefits: the initial download finishes faster and the background downloads cause less load.

Fixes: #592


kolibri_explore_plugin/jobs.py
```diff
@@ -0,0 +1,108 @@
...
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
```
pwithnall, line 15: Nitpick: Could benefit from an SPDX license identifier.

kolibri_explore_plugin/jobs.py
```diff
@@ -105,13 +107,20 @@ def storage_update_hook(job, orm_job, state=None, **kwargs):
         # We only care about state transitions here.
         return
 
+    try:
+        bg_task = BackgroundTask.objects.get(job_id=job.job_id)
+    except BackgroundTask.DoesNotExist:
+        # Not one of our tasks.
+        return
+
+    # Syncronize the state if needed.
```
pwithnall, line 157: Typo ‘synchronize’

kolibri_explore_plugin/models.py
```diff
@@ -1,7 +1,13 @@
+import json
```
pwithnall, line 1: Ideally this file could do with a copyright/licensing header (as a separate commit), but that’s not a blocker for merging this PR because it’s a pre-existing issue.

kolibri_explore_plugin/jobs.py
```diff
@@ -94,13 +96,52 @@ def get_remotecontentimport_task(
...
     task_func = import_path_to_callable(task)
     job_data, _ = task_func.validate_job_data(user, params)
     return job_storage.enqueue_job(job_data, queue=queue, priority=priority)
 
 
+def enqueue_next_background_task():
+    """Locate the next background task and enqueue it for running"""
+
+    # If there's already a job in progress, do nothing. We only want a
+    # single background task running.
```
pwithnall, line 121: Perhaps link to this issue, and include a brief rationale like “… only want a single background task running in order to not slow down the user interface and to leave room in the worker pool for user-triggered downloads”.

pwithnall (commented):
LGTM! I have a small collection of thoroughly insignificant comments.

kolibri_explore_plugin/jobs.py
dbnicholson: :+1:

kolibri_explore_plugin/models.py
dbnicholson: Good call. I think I'll add a prep commit that adds SPDX license identifiers to all the python files.

kolibri_explore_plugin/jobs.py
dbnicholson: That's a good idea. It's not really clear why you'd want to do that.

Issue #743, opened 2023-08-14
Backend test suite


dbnicholson:
Currently whenever any changes are made to the backend code, manual testing is needed to validate it. That slows down development and makes it more likely that regressions will be introduced. There should be a test suite for the python backend to automatically validate changes.

Since Kolibri is a Django application, familiarity with [Django testing](https://docs.djangoproject.com/en/1.11/topics/testing/) is required. However, since Kolibri initializes Django, the Django testing infrastructure can't be used directly without adding a custom test runner. Furthermore, the Django testing tools are based on [unittest](https://docs.python.org/3/library/unittest.html), which I personally find to be unsatisfactory. Kolibri itself uses [pytest-django](https://pytest-django.readthedocs.io/en/latest/), which bridges Django to [pytest](https://docs.pytest.org/en/stable/). Even then a bit of custom setup code is needed to get Kolibri and Django to initialize in a way that resembles Kolibri at runtime.

I worked on this a few weeks ago and have some minimal tests that can serve as a starting point.

pwithnall:
#763 looked excellent, so I merged it. Are you intending for more tests to be written imminently as part of this issue (in which case I should move it back to In Progress), or is #763 it for now (in which case I should move the issue to Verify)?

dbnicholson:
Thanks! I'm going to try to get the collection downloader tests landed since I'm close and that we be a big win. If it seems like that's taking too long, I'll split that out to a separate task and close this one.

dbnicholson:
I was working on several other tasks and the collection downloader tests got pushed down. I opened #778 to hopefully fix in the next cycle.

PR #763, opened 2023-08-23
Python backend testing

dbnicholson:
This contains the infrastructure and a couple initial tests for testing the python backend with `pytest`. The added tests aren't especially useful, but they're included as a starting point. A bit more infrastructure will be needed to test the collection downloader and views. I have some of this locally, but it's messy. I'm hoping to get in this initial portion first.

Helps: #743


pwithnall (approved):
:rocket:

Issue #760, opened 2023-08-18
UI Localization


dbnicholson:
We want the frontend components to be localized to Spanish. For now we'll use in house translation, but it might be nice to wire up Transifex.

dbnicholson:
That all makes sense to me, thanks for looking into it.

pwithnall:
PR ready for some initial glancing at here: https://github.com/endlessm/kolibri-explore-plugin/pull/766, but only if you’re bored. It’s not ready for proper review yet.

A companion issue now exists with the full set of strings for translation, so we can get started on translations: https://github.com/endlessm/kolibri-explore-plugin/issues/765

I’ve filed an issue against kolibri-tools about `pyproject.toml` support, because currently the tooling only supports `setup.cfg`. https://github.com/learningequality/kolibri/issues/11146

---

For my notes, here are some of the other Kolibri plugin projects which I’ve looked at for inspiration:
 * https://github.com/learningequality/kolibri-exercise-perseus-plugin/blob/develop/package.json
 * https://github.com/learningequality/cookiecutter-kolibri-plugin
 * https://github.com/learningequality/kolibri-instant-schools-plugin/blob/develop/kolibri_instant_schools_plugin/instant_schools_settings.py
 * https://github.com/learningequality/studio/blob/d76efcf03a8c2fe6d43c67afcfd186f7b5ae667f/package.json#L12
 * https://github.com/learningequality/studio/blob/d76efcf03a8c2fe6d43c67afcfd186f7b5ae667f/package.json#L38

pwithnall:
In #765, Joana found that there are a number of untranslatable strings in the `welcome-screen` package. I’ve filed #770 about that.

pwithnall:
I put together a rough (rubbish) Spanish translation today in order to test the frontend translation functions, and have a result!

![image](https://github.com/endlessm/kolibri-explore-plugin/assets/53172/5171506f-bb0e-4774-9b94-0f5c5a31681d)

That didn’t actually need any frontend changes to start using the JSON files containing the translations. As long as the JSON file is named correctly, Kolibri’s webpack bundle hook will pick it up via its well-known name and inject it into the frontend.

I’ve added a fixup commit to explain that in the docs.

As can be seen from the screenshot, there are still some English strings on the main page. That is separate from #765 as this isn’t the welcome screen. I’ll look into making sure they’re translated tomorrow, and then this issue should be ready for review and merging.

pwithnall:
[...]

PR #766, opened 2023-08-24
build: Add i18n machinery

pwithnall:
This adds an initial version of internationalisation support. It doesn’t
support Transifex (translations have to be committed manually) for now.

See `docs/i18n.md` for a guide to what commands to use and what the
files do.

Fixes: https://github.com/endlessm/kolibri-explore-plugin/issues/760


pwithnall:
This is a draft. The machinery seems to work (it doesn’t error locally for me, although it remains to be seen whether it’ll work on CI) and all the right files end up in the `.whl` file.

I haven’t tested it at runtime yet though.

package.json
```diff
@@ -2,14 +2,18 @@
...
-    "build": "yarn build:version && yarn build:packages && yarn deploy:welcome && yarn build:plugin && yarn build:info",
+    "build": "yarn build:version && yarn build:packages && yarn deploy:welcome && yarn build:plugin && yarn build:info && yarn build:i18n",
     "build:plugin": "./scripts/set_override.py default && kolibri-tools build prod --plugins kolibri_explore_plugin --transpile",
     "build:packages": "yarn build:libs",
     "build:libs": "lerna run build --ignore template-ui",
     "build:apps": "lerna run build --scope template-ui",
     "build:info": "./scripts/build_info.sh > kolibri_explore_plugin/static/build-info.json",
     "build:version": "python -m setuptools_scm",
+    "build:i18n": "yarn run i18n-extract && yarn run i18n-django-compilemessages",
     "build-dist": "./scripts/clean.sh && yarn clean && yarn build && python -m build",
```
dbnicholson, line 17: Am I wrong or does `i18n-download-translations` need to be run by one of these build targets to make sure the compiled messages are included in the package? Or is the idea that the compiled messages are committed and therefore we want to release that instead of regenerating them at the release time?

dbnicholson (commented):
Looks useful so far!

package.json
pwithnall: Following what Kolibri do, the compiled messages (`.mo`, `kolibri_explore_plugin/assets/src/{vue-,}intl-locale-data.js`) are committed, so you’re right that the idea is that we release those rather than regenerating them.

pwithnall:
Pushed some fixups for the comments above, plus improvements to the docs to reflect my new understanding of how the translated messages are registered in the frontend (which I got wrong before).

See [today’s progress report](https://github.com/endlessm/kolibri-explore-plugin/issues/760#issuecomment-1696047070).

pwithnall:
Actually, re-marking as a draft because there are unmerged fixup commits in the branch. They give the diffs since the previous review. It’s all ready for review, but hopefully this will stop it being accidentally merged before those are squashed.

dbnicholson (approved):
This looks great! I much appreciate the details of how it works. Feel free to squash and merge.

Issue #712, opened 2023-07-24
PhET simulations (eng/es) have white text on the background color


vanessa-chang:
The current header background is set to default. The text is displayed in white text and it's hard to read.

![PhET](https://github.com/endlessm/kolibri-explore-plugin/assets/31236978/66cff842-eaac-4f28-b04c-e702a6020e59)

Both channels should be fixed. The header color should be #613863. The white text color works in that case of the purple bg color.

- 197934f144305350b5820c7c4dd8e194
- 8fa678af1dd05329bf3218c549b84996

jprvita:
This is how it looks with the header set to `#613863`.

![Screenshot from 2023-08-24 16-51-33](https://github.com/endlessm/kolibri-explore-plugin/assets/73894/47477dc1-9a41-406f-9ad5-2fd556da6daa)

vanessa-chang:
Verified pass in Chromebook Manatee 6.40-394 & Windows 6.40.1.0

PR #767, opened 2023-08-24
PhET Simulations: Adjust header color

jprvita:
https://github.com/endlessm/kolibri-explore-plugin/issues/712


jprvita:
Screenshot of how it looks with this change in https://github.com/endlessm/kolibri-explore-plugin/issues/712#issuecomment-1692560824.

GeorgesStavracas (approved):

Issue #772, opened 2023-08-29
Add translation calls to template-ui


pwithnall:
Currently the template-ui package doesn’t fully support localisation/translation. It has a few translatable strings, but mostly hard-codes English strings in its code/templates.

PR #774, opened 2023-08-29
template-ui: Make various UI strings translatable

pwithnall:
Previously they were hard-coded as English. A few strings were translatable, but most weren’t.

Fixes: #772

### DIFF
--- a/packages/template-ui/src/components/ChannelFooter.vue
+++ b/packages/template-ui/src/components/ChannelFooter.vue
@@ -13,7 +13,7 @@
       <template #right>
         <div v-if="isEndlessApp" class="endless-seal text-right">
           <div class="mb-1">
-            <span class="font-weight-bold">Best of the Web</span> curated by
+            {{ $tr('bestOfTheWeb') }}
           </div>
           <img :src="logo" aria-hidden="true">
         </div>
@@ -32,6 +32,14 @@ export default {
     ...mapState(['channel', 'isEndlessApp']),
     logo() {
       return EndlessLogo;
+    },
+  },
+  $trs: {
+    bestOfTheWeb: {
+      // FIXME: This isn’t translatable to languages where the logo can’t be a suffix,
+      // or RTL languages.
+      message: '<span class="font-weight-bold">Best of the Web</span> curated by ',
+      context: 'Footer curation message. The Endless Key logo is displayed just after the end of the sentence',
     },
   },
 };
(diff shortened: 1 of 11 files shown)


packages/template-ui/src/views/Search.vue
```diff
@@ -146,6 +145,28 @@ export default {
...
+    noContentFoundGuidance: {
+      message: 'You can try a different search, maybe use fewer words, or try one of the topic suggestions below.',
+      context: 'Guidance given to the user when no search results are found',
+    },
+    exploreTopics: {
+      message: 'Explore topics',
+      context: 'Heading for a section of content grid in the search results',
+    },
+    resultsCount: {
+      message: '{num_results} Results',
```
dylanmccall, line 166:
```suggestion
      message: '{num_results, number, integer} {num_results, plural, one {Result} other {Results}}',
```

We can be a bit fancier now this is going through the translation stuff :)

packages/template-ui/src/components/ChannelFooter.vue
```diff
@@ -34,6 +34,14 @@ export default {
       return EndlessLogo;
     },
   },
+  $trs: {
+    bestOfTheWeb: {
+      // FIXME: This isn’t translatable to languages where the logo can’t be a suffix,
+      // or RTL languages.
+      message: '<span class="font-weight-bold">Best of the Web</span> curated by ',
```
dylanmccall, line 41:
```suggestion
      message: '<strong>Best of the Web</strong> curated by ',
```

While we're at it, could we use `<strong>` here instead of a `<span …>`? (And it'll probably be styled the same, but double check that). It'll be less trouble for translating.

packages/template-ui/src/components/ChannelFooter.vue
```diff
@@ -34,6 +34,14 @@ export default {
       return EndlessLogo;
     },
   },
+  $trs: {
+    bestOfTheWeb: {
+      // FIXME: This isn’t translatable to languages where the logo can’t be a suffix,
+      // or RTL languages.
```
manuq, line 40: :+1:

packages/template-ui/src/views/Test.vue
```diff
@@ -1,4 +1,5 @@
 <template>
+  <!-- Deliberately not localised because it’s only test content -->
```
manuq, line 2: :+1:

manuq (approved):

Issue #761, opened 2023-08-22
Add download buttons to online Instance more prominently


jofilizola:
# Context

from Rodrigo:

"The California Academy of Sciences has agreed to help us promote the app. They will be posting something on their website and their newsletter. While reviewing the information, I noticed the recommend the online version instead of the app which is fine. I'll be asking her to change that however, in some cases, we won't be able to control that and while chatting with Luke about this, we noticed that the buttons to download the app in the online version are at the bottom and not at the top and there's no CTA to promote app download.
Could the inclusion of these logos and a textual guide to encourage visitors to download the app be a big lift? "

# Specs and assets

> [!NOTE] 
> UPDATED SPECS, according to what discussed with @jprvita on Weekly meeting and Slack.

- Add button that triggers the automatic animated scroll to the bottom, where the store badges are.
- [Figma prototype](https://www.figma.com/proto/Mwgjbi2CHUoTMdkE9aGQfP/EK-Manatee?page-id=501%3A5042&type=design&node-id=525-44171&viewport=-20371%2C242%2C0.79&t=1dmDrnT5CKpNEOkv-1&scaling=min-zoom&starting-point-node-id=525%3A44171&show-proto-sidebar=1&mode=design)
- [▶ Online Instance - EK Manatee - 24 August 2023 - Watch Video](https://www.loom.com/share/0701dbd88d8b46c488223a5f158063eb)
- [Figma board](https://www.figma.com/file/Mwgjbi2CHUoTMdkE9aGQfP/EK-Manatee?type=design&node-id=525%3A42050&mode=dev)

![Image](https://github.com/endlessm/kolibri-explore-plugin/assets/51095367/58d44717-bd8a-4069-9c10-3984e6b4e992)

> [!NOTE] 
> OLD SPECS
- ~~Add smaller versions of app stores buttons on the top of the page~~
- ~~Add accompanying copy~~
- ~~Add the "source" parameters to the links so we know where people are coming from~~
- ~~Keep the bottom buttons~~
- ~~[Figma board](https://www.figma.com/file/Mwgjbi2CHUoTMdkE9aGQfP/EK-Manatee?type=design&node-id=511%3A29570&mode=design&t=MqKOqLdOgV5xto8Q-1)~~
- ~~Store badges: https://www.dropbox.com/scl/fo/nymgatj4ha6hqf0k9cmco/h?rlkey=fy9phhpkaqnez9pqatv9ujadv&dl=0~~

[...]

PR #779, opened 2023-08-31
NavBar: Add download button

jprvita:
This commit adds a download button to the navigation bar that scrolls the viewport to the bottom of the page when clicked, where the app store badges can be found.

https://github.com/endlessm/kolibri-explore-plugin/issues/761

### DIFF
--- a/kolibri_explore_plugin/assets/src/components/DiscoveryNavBar.vue
+++ b/kolibri_explore_plugin/assets/src/components/DiscoveryNavBar.vue
@@ -26,6 +26,23 @@
       </b-nav-text>
     </b-button-group>
     <b-navbar-nav>
+      <b-button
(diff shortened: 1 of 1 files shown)


GeorgesStavracas (approved):

GeorgesStavracas:
I don't fully understand the consequences of using `d-none` and `d-inline` as they were used here, but I couldn't *prove* they wouldn't work. So let's go with it.

jprvita:
Thanks for reviewing!

> I don't fully understand the consequences of using `d-none` and `d-inline` as they were used here, but I couldn't _prove_ they wouldn't work. So let's go with it.

This is a good reference about how that works: https://vuetifyjs.com/en/styles/display/#visibility

PR #781, opened 2023-09-04
Explain how to workaround version in editable mode

manuq:
Add a README note as reminder.

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,6 +92,10 @@ pip install --editable .
 **Note:** you still need to enable the plugin and run migrations as in
 Usage above!
 
+**Note:** since the version number is now generated at build time, for
+the project to work in editable mode you need to workaround this by placing
+something like `__version__ = 'dev'` in `kolibri_explore_plugin/_version.py`.
+
 Then serve the plugin in watch mode:
 
 ```


manuq:
I needed this in my development setup.

pwithnall (changes requested):
Why is this needed? What’s the failure mode if you don’t do this?

aiui this shouldn’t be needed because there’s an exception handler in `kolibri_explore_plugin/__init__.py` which sets a default version if `_version.py` is missing. Do we need something else like that somewhere else?

manuq:
> Why is this needed? What’s the failure mode if you don’t do this?
> 
> aiui this shouldn’t be needed because there’s an exception handler in `kolibri_explore_plugin/__init__.py` which sets a default version if `_version.py` is missing. Do we need something else like that somewhere else?

I got this when running the migrations with `kolibri manage migrate`. Maybe the kolibri migration doesn't use the `__init__.py` as entry point?

```
INFO     2023-09-04 15:12:47,957 Invoking command migrate
WARNING  2023-09-04 15:12:48,095 Tried to import module urls from kolibri_explore_plugin but an error was raised
ERROR    2023-09-04 15:12:48,095 No module named 'kolibri_explore_plugin._version'
Traceback (most recent call last):
  File "/sysroot/home/manuq/eos/hack/gitrepos/kolibri-16-alpha/kolibri/plugins/__init__.py", line 256, in _return_module
    return import_module(models_module_name)
  File "/usr/lib/python3.9/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1030, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1007, in _find_and_load
  File "<frozen importlib._bootstrap>", line 986, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 680, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 790, in exec_module
  File "<frozen importlib._bootstrap>", line 228, in _call_with_frames_removed
  File "/sysroot/home/manuq/eos/hack/gitrepos/kolibri-explore-plugin/kolibri_explore_plugin/urls.py", line 7, in <module>
    from ._version import __version__ as VERSION
ModuleNotFoundError: No module named 'kolibri_explore_plugin._version'
```

dbnicholson:
Can this be fixed with the following change?

```
diff --git a/kolibri_explore_plugin/urls.py b/kolibri_explore_plugin/urls.py
index e6adaee..2bf4be7 100644
--- a/kolibri_explore_plugin/urls.py
+++ b/kolibri_explore_plugin/urls.py
@@ -4,7 +4,7 @@ import os
 
 from django.conf.urls import url
 
-from ._version import __version__ as VERSION
+from . import __version__ as VERSION
 from .views import AppFileView
 from .views import AppMetadataView
 from .views import ExploreView
```

manuq:
> Can this be fixed with the following change?

Ah yes, that works! I'll update the PR, thanks.

manuq:
Superceeded by #791

Issue #773, opened 2023-08-29
Add translation calls to kolibri_explore_plugin


pwithnall:
Currently the kolibri_explore_plugin package itself doesn’t fully support localisation/translation. About half of its strings are translated, but the other half are currently hard-coded in English.

PR #782, opened 2023-09-05
kolibri_explore_plugin: Make various UI strings translatable

pwithnall:

Fixes: #773

### DIFF
--- a/kolibri_explore_plugin/assets/src/components/AboutModal.vue
+++ b/kolibri_explore_plugin/assets/src/components/AboutModal.vue
@@ -29,38 +29,26 @@
         <b-col id="sections-column" class="sections-column" md="9" sm="12">
           <div>
             <h3 id="about-section" class="pt-3 text-primary">
-              Endless Key — learning and discovery resources, no internet required.
+              {{ $tr('aboutHeading') }}
             </h3>
             <p>
-              The Endless Key unlocks the potential for student
-              curiosity and choice through independent
-              learning. Curated to be interest-based and inclusive,
-              our collection of content encourages students to
-              discover knowledge and skills that will support them now
-              and in their futures. The Endless Key is designed to
-              complement the broadband-for-all roll-out in the United
-              States and provide offline students with equitable
-              access to engaging content. Students can access Endless
-              Key content by downloading the Endless Key app
-              (available for ChromeOS and Windows). After the initial
-              download, no internet is required to interact with the
-              content.
+              {{ $tr('aboutBody') }}
             </p>
             <hr>
             <template v-if="buildInfo">
-              <h5>Release information</h5>
+              <h5>{{ $tr('releaseInformationHeading') }}</h5>
               <b-list-group class="bg-light p-3 rounded-lg">
                 <b-list-group-item class="bg-light rounded-0">
-                  Release: {{ buildInfo.version_name }} {{ buildInfo.last_release }}
+                  {{ $tr('releaseKey') }} {{ buildInfo.version_name }} {{ buildInfo.last_release }}
                 </b-list-group-item>
                 <b-list-group-item class="bg-light rounded-0">
-                  Date: {{ buildInfo.date }}
+                  {{ $tr('dateKey') }} {{ buildInfo.date }}
                 </b-list-group-item>
                 <b-list-group-item class="bg-light rounded-0">
-                  Commit: {{ buildInfo.commit }}
+                  {{ $tr('commitKey') }} {{ buildInfo.commit }}
                 </b-list-group-item>
                 <b-list-group-item class="bg-light rounded-0">
-                  Kolibri: {{ kolibriVersion }}
+                  {{ $tr('kolibriKey') }} {{ kolibriVersion }}
                 </b-list-group-item>
               </b-list-group>
               <hr>
@@ -68,15 +56,10 @@
           </div>
           <div>
             <h3 id="support-section" class="pt-3 text-primary">
-              Support
+              {{ $tr('supportHeading') }}
             </h3>
-            <p>
-              You can find more information on how to use your Endless Key on
-              <strong>The Quick Start Guide</strong> document. For more
-              information, when you have an internet connection, you can access
-              Endless Support Site at
-              <a href="https://support.endlessos.org/en/endless-key" target="_blank">https://support.endlessos.org/en/endless-key</a>.
-            </p>
+            <!-- eslint-disable-next-line vue/no-v-html -->
+            <p v-html="$tr('supportBody')"></p>
             <hr>
           </div>
           <div>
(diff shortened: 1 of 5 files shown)


pwithnall:
This is based on #777, so that needs to be reviewed and merged before this one. Only the top commit in this PR is actually specific to `kolibri_explore_plugin` core.

pwithnall:
Rebasing this now that #777 has landed.

pwithnall:
Pushing to test CI. It is failing to build locally for me but I can’t work out why (and so is `master`), so I think it might be an environment problem for me.

manuq:
@pwithnall running the linter locally succeeds on my end. The CI says `error kolibri-explore-plugin@: The engine "node" is incompatible with this module. Expected version "16.x". Got "18.17.1"` . Since when the CI is running node v18? I have v16 here and that's what the `scripts/bootstrap.sh` ensures too.

manuq:
@dbnicholson maybe the changes in a06ac87382095f882a89e37794b4303a6b8a4caf (which are very well intentioned) regressed the node version in the CI.

pwithnall:
Locally I’m getting the following (and am also getting it on `master`), and I haven’t been able to make headway on understanding it yet:
```
Kolibri Build: Building 'kolibri_explore_plugin.app' bundle...
Kolibri Build: Building 'kolibri_explore_plugin.side_nav' bundle...
Kolibri Build: Completed 'kolibri_explore_plugin.side_nav' bundle in 2.685s!
[BABEL] Note: The code generator has deoptimised the styling of /home/philip/Projects/kolibri-explore-plugin/node_modules/bootstrap-vue/esm/icons/icons.js as it exceeds the max of 500KB.
[BABEL] Note: The code generator has deoptimised the styling of /home/philip/Projects/kolibri-explore-plugin/node_modules/lodash/lodash.js as it exceeds the max of 500KB.
Kolibri Build: Failed to compile 'kolibri_explore_plugin.app' bundle!
kolibri_explore_plugin.app6.40.1.dev12+g0f97ae2.css from Css Minimizer plugin
TypeError: Cannot read properties of undefined (reading 'Universal')

Kolibri Build: Build complete in 34.315 seconds
Kolibri Build: kolibri_explore_plugin.app:
  ERROR in kolibri_explore_plugin.app6.40.1.dev12+g0f97ae2.css
  kolibri_explore_plugin.app6.40.1.dev12+g0f97ae2.css from Css Minimizer plugin
  TypeError: Cannot read properties of undefined (reading 'Universal')
      at Object.<anonymous> (/home/philip/Projects/kolibri/packages/kolibri-tools/node_modules/css-select/lib/sort.js:6:30)
      at Module._compile (node:internal/modules/cjs/loader:1103:14)
      at Object.Module._extensions..js (node:internal/modules/cjs/loader:1155:10)
      at Module.load (node:internal/modules/cjs/loader:981:32)
      at Function.Module._load (node:internal/modules/cjs/loader:822:12)
      at Module.require (node:internal/modules/cjs/loader:1005:19)
      at require (node:internal/modules/cjs/helpers:102:18)
      at Object.<anonymous> (/home/philip/Projects/kolibri/packages/kolibri-tools/node_modules/css-select/lib/compile.js:32:30)
      at Module._compile (node:internal/modules/cjs/loader:1103:14)
      at Object.Module._extensions..js (node:internal/modules/cjs/loader:1155:10)

  ERROR in kolibri_explore_plugin.app6.40.1.dev12+g0f97ae2.rtl.css
  kolibri_explore_plugin.app6.40.1.dev12+g0f97ae2.rtl.css from Css Minimizer plugin
  TypeError: Cannot read properties of undefined (reading 'Universal')
      at Object.<anonymous> (/home/philip/Projects/kolibri/packages/kolibri-tools/node_modules/css-select/lib/sort.js:6:30)
      at Module._compile (node:internal/modules/cjs/loader:1103:14)
      at Object.Module._extensions..js (node:internal/modules/cjs/loader:1155:10)
      at Module.load (node:internal/modules/cjs/loader:981:32)
      at Function.Module._load (node:internal/modules/cjs/loader:822:12)
      at Module.require (node:internal/modules/cjs/loader:1005:19)
      at require (node:internal/modules/cjs/helpers:102:18)
      at Object.<anonymous> (/home/philip/Projects/kolibri/packages/kolibri-tools/node_modules/css-select/lib/compile.js:32:30)
      at Module._compile (node:internal/modules/cjs/loader:1103:14)
      at Object.Module._extensions..js (node:internal/modules/cjs/loader:1155:10)

  kolibri_explore_plugin.app compiled with 2 errors
error Command failed with exit code 1.

```

dbnicholson:
I might be wrong, but I think this is https://github.com/ekalinin/nodeenv/issues/324, which is ultimately https://github.com/nodejs/build/issues/1993.

Well, a more "proper" way to fix it would be to add back the `setup_node` action and add a switch to `bootstrap.sh` that explicitly skips running `nodeenv`. I don't think that the current way where errors from `bootstrap.sh` are silently ignored are a good idea. I can work on that in a bit, but I have something to do this morning.

pwithnall:
> Well, a more "proper" way to fix it would be to add back the `setup_node` action and add a switch to `bootstrap.sh` that explicitly skips running `nodeenv`.

I’ll look at doing that now, since I’m getting absolutely nowhere with this ‘Css Minimizer’ failure.

pwithnall:
OK, I think this is ready for review now. The only change since last time anyone saw it was that I’ve had to use [`v-html`](https://v2.vuejs.org/v2/api/#v-html) to insert the `supportBody` string, as it contains inline markup which would otherwise be escaped. The use of `v-html` could be an injection vector if we’re not careful to make sure it is only ever passed developer-controlled data, so it could do with some close review.

kolibri_explore_plugin/assets/src/components/AboutModal.vue
```diff
@@ -371,6 +433,173 @@
...
+        message: 'Kolibri:',
+        context: 'Table key for software Kolibri version',
+      },
+      supportHeading: {
+        message: 'Support',
+        context: 'Heading for technical support section',
+      },
+      supportBody: {
+        message:
+          'You can find more information on how to use your Endless Key on <strong>The Quick Start Guide</strong> document. For more information, when you have an internet connection, you can access the Endless Support Site at <a href="https://support.endlessos.org/en/endless-key" target="_blank">https://support.endlessos.org/en/endless-key</a>.',
```
manuq, line 471: I think it's fine to have the v-html. I considered using a name like `"...you can access the Endless Support Site at { siteURL}."` and have that URL as vue data. Maybe it's better to de-hardcode the URL from the translators message, but doesn't avoid the use of v-html.

manuq (approved):

kolibri_explore_plugin/assets/src/components/AboutModal.vue
manuq: It would avoid the problem of a malicious translator sending users to a different URL, but what are the chances?

pwithnall: Yeah, I think the use of `v-html` is fine as things are, but it’s something we want to be careful not to break in future.

pwithnall: This is also something I think we could iterate on in future if it turns out that having the markup in the translatable string causes problems.

In GNOME, this kind of setup occasionally causes problems where translators literally translate `<strong>` to (for example) `<fuerte>`. But that is quite rare.

pwithnall:
Merging as per approval

Issue #604, opened 2023-06-13
Implement proper i18n in template-ui


GeorgesStavracas:
While implementing downloadable content, it was noticed that the template components under `template-ui` do not support translations. A workaround has been proposed, and is used for now, but strings still aren't localized as the workaround merely uses the default string (always in English).

erikos:
Localization works now in 7.1 plugin upwards. The Spanish localization has been verified.

PR #784, opened 2023-09-06
template-ui: Call i18nSetup() to set up i18n for Vue

pwithnall:

Fixes: #604

### DIFF
--- a/packages/template-ui/src/main.js
+++ b/packages/template-ui/src/main.js
@@ -1,5 +1,6 @@
 import Vue from 'vue';
 import { BootstrapVue, IconsPlugin } from 'bootstrap-vue';
+import { i18nSetup } from 'kolibri.utils.i18n';
 import App from '@/App.vue';
 
 import router from '@/router';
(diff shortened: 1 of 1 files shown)


pwithnall:
This tidies up the stub `$tr` function which was previously added as a workaround in template-ui.

It calls `i18nSetup()` because the rest of the code here is calling Vue setup functions.

packages/template-ui/src/main.js
```diff
@@ -1,5 +1,6 @@
 import Vue from 'vue';
 import { BootstrapVue, IconsPlugin } from 'bootstrap-vue';
+import { i18nSetup } from 'kolibri.utils.i18n';
```
manuq, line 3: This fails as it requirer `kolibri` as dependency of the template-ui. I guess you are testing this on top of the other PR?

Here is my alternative proposal #790 . This is what I was meaning about passing the translations through the custom presentation API.

pwithnall: > This fails as it requires `kolibri` as dependency of the template-ui. I guess you are testing this on top of the other PR?

I don’t think I was testing it on top of the other PR, otherwise the branch would be based on that one. The CI seems happy with it as well.

> Here is my alternative proposal #790 . This is what I was meaning about passing the translations through the custom presentation API.

That looks fine to me, though I don’t really know what I’m talking about. Are you sure we don’t need the other things that `i18nSetup()` does, though? It sets up language and font settings, `Intl` polyfills, and imports locale data.

dbnicholson:
I decided to take #790 instead with the hopes that a `kolibri` dependency within template-ui (which affects every channel app zip) can be avoided. Maybe we'll have to bite the bullet on that when actually wiring up the i18n infrastructure in template-ui. Closing this for now.

Issue #780, opened 2023-09-01
Feature channel row is not shown 


vanessa-chang:
Tested with the latest beta version: Manatee 6.40-394 & Windows 6.40.1.0

The feature channel row is not shown unless you switch to the Library tab then switch back to Discovery page.
But the missing channel icons is fixed.

Chromebook log:
[feature_channel.txt](https://github.com/endlessm/kolibri-explore-plugin/files/12462302/feature_channel.txt)

Windows log:
[kolibri.txt](https://github.com/endlessm/kolibri-explore-plugin/files/12462305/kolibri.txt)

_Originally posted by @vanessa-chang in https://github.com/endlessm/kolibri-explore-plugin/issues/642#issuecomment-1697040570_

dbnicholson:
I didn't see this initially, but I do now. I'll see if I can figure it out.

dbnicholson:
This is definitely fallout from #642 due to b9b60d228ab7e11a7f6f7e222b6a0780dcdc5965. What I wanted to do there was use the root node object from `topicsRoot.rootNodes` because `rootNodes` has the channel metadata's inlined thumbnail set as the `thumbnail` property. Without that, you'd have to search through the channels again to find it.

I think the problem may be that that the use of `rootNodes` in `DiscoveryPageContenPacks` is not synchronized with `rootNodes` being populated. What I see in the console is:

```
TypeError: Cannot read properties of undefined (reading 'id')
    at DiscoveryPageContenPacks.vue:48:1
    at Proxy.renderList (vue.runtime.esm.js:2646:1)
    at fn (DiscoveryPageContenPacks.vue:44:1)
    at normalized (vue.runtime.esm.js:2605:1)
    at Proxy.renderSlot (vue.runtime.esm.js:2701:1)
    at fn (EkSlidableGrid.vue:85:1)
    at normalized (vue.runtime.esm.js:2605:1)
    at normalizeSlot (normalize-slot.js:51:1)
    at VueComponent.normalizeSlot (normalize-slot.js:24:1)
    at Proxy.render (carousel-slide.js:70:1)
```

Which is the change in b9b60d228ab7e11a7f6f7e222b6a0780dcdc5965 trying to find the equivalent node in `rootNodes`. So, I think the possible fixes are:

* Block until `rootNodes` is set with `topicsRoot/SET_STATE`.
* Revert b9b60d228ab7e11a7f6f7e222b6a0780dcdc5965 and add in the missing channel data manually.

[...]

PR #785, opened 2023-09-06
Really fix featured channel icons on discovery page

dbnicholson:
The previous fix to use `rootNodes` didn't work since that's populated asynchronously. Rather than try to block on that, this just repeats the channel lookup and mutates the node to have the channel's thumbnail fallbacks.

Fixes: #780


manuq:
Looks good!

PR #789, opened 2023-09-06
Build welcome-screen.zip for release

dbnicholson:
The welcome-screen.zip file is intended to be included as a release artifact, but it wasn't actually being built. Add the necessary build step and make `action-gh-release` fail when a pattern doesn't match any files. Without this it just prints a warning:

```
🤔 Pattern 'packages/welcome-screen/welcome-screen.zip' does not match any files.
```


starnight (approved):
LGTM!  Thanks!

PR #744, opened 2023-08-15
build: Add missing babel config for root package

pwithnall:
This was missing, meaning the default babel config was being used when
building the whole of kolibri-explore-plugin.

This meant that transpilation and ES5 polyfills were not being applied.

Signed-off-by: Philip Withnall <philip@tecnocode.co.uk>

Fixes: https://github.com/endlessm/kolibri-installer-android/issues/159

PR #790, opened 2023-09-07
template-ui: Hook i18n

manuq:
Pass the message ID through the API provided by the plugin.

On the plugin side, use kolibri i18n utility to create a translator for the given namespace and return the translated message.

Helps #604

### DIFF
--- a/kolibri_explore_plugin/assets/src/kolibriApi.js
+++ b/kolibri_explore_plugin/assets/src/kolibriApi.js
@@ -5,6 +5,7 @@ import { ContentNodeKinds } from 'kolibri.coreVue.vuex.constants';
 
 import router from 'kolibri.coreVue.router';
 import store from 'kolibri.coreVue.vuex.store';
+import { createTranslator } from 'kolibri.utils.i18n';
 import { utils } from 'ek-components';
 import { ChannelResource, ContentNodeResource } from './apiResources';
 import { showTopicsContentInLightbox } from './modules/topicsTree/handlers';
@@ -21,6 +22,9 @@ const DEFAULT_HIDE_UNAVAILABLE = false;
 class KolibriApi {
   constructor(channelId) {
     this.channelId = channelId;
+
+    // Cache of template-ui translators:
+    this.translators = {};
   }
 
   themeRenderer() {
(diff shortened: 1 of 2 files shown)


kolibri_explore_plugin/assets/src/kolibriApi.js
```diff
@@ -21,6 +22,9 @@ const DEFAULT_HIDE_UNAVAILABLE = false;
 class KolibriApi {
   constructor(channelId) {
     this.channelId = channelId;
+
+    // Cache of template-ui translators:
+    this.translators = {};
```
manuq, line 27: Also this should be a static class field. But the current linter is not happy about it. Has the linter regressed?

kolibri_explore_plugin/assets/src/kolibriApi.js
```diff
@@ -234,6 +238,22 @@ class KolibriApi {
...
     return DEFAULT_HIDE_UNAVAILABLE;
   }
+
+  translate(nameSpace, defaultMessages, messageId, args) {
+    // Create the translator and add it to the cache only if needed:
+    let translator;
+    // FIXME this would be more readable by using the nullish coalescing
+    // assignment (??=) operator, but current linter is not happy:
+    // this.translators[nameSpace] ??= createTranslator(nameSpace, defaultMessages);
+    // const translator = this.translators[nameSpace];
```
manuq, line 248: Has the linter regressed? I remember using es6 operators like this in the past.

dbnicholson: I don't remember changing the linter at all, but we did run into a problem with that operator in https://github.com/endlessm/kolibri-explore-plugin/pull/693.

pwithnall: I changed the babel configuration in https://github.com/endlessm/kolibri-explore-plugin/pull/744 and verified that it accepts nullish coalescing operators now. But I didn’t explicitly test the linter; perhaps it could be subject to the same problem? Does it load the babel config?

manuq: https://mastodon.social/@gknauss/111046912763765216

dbnicholson:
I rebased this on master to get the tests against the recent dependency changes. That said, I tried this locally with the nullish coalescing operator part uncommented and got the same linter errors. I don't know why that's happening.

dbnicholson (approved):
Merging this as it seems like the right thing to do to avoid adding a `kolibri` dependency in each channel app.

PR #791, opened 2023-09-07
Import VERSION properly in urls

manuq:
Otherwise running `kolibri manage migrate` fails.


manuq:
Auto-merging as requested.

PR #793, opened 2023-09-08
Migrate Welcome

### LINKED ISSUE #748: Move welcome-screen to the core plugin code
With #732, the welcome screen no longer deals with choosing between using external (USB) storage or internal storage. This means there is no need to run it before Kolibri starts. Instead, let's move the welcome screen code to the plugin's core Vue code - this should relate nicely to [DownloadPage.vue](https://github.com/endlessm/kolibri-explore-plugin/blob/master/kolibri_explore_plugin/assets/src/views/DownloadPage.vue) - and remove the welcome-screen package entirely.

This will require some additional changes in the Android and Windows app to support the new behaviour.

manuq:
Fixes #748


pwithnall:
What’s the loading-screen package for? I’m probably misunderstanding the plan in #748 entirely, but I thought the big win here was to remove the welcome-screen package. Renaming it looks to me like it would keep us with the same build system and i18n pain we have with welcome-screen, no?

manuq:
> What’s the loading-screen package for? I’m probably misunderstanding the plan in #748 entirely, but I thought the big win here was to remove the welcome-screen package. Renaming it looks to me like it would keep us with the same build system and i18n pain we have with welcome-screen, no?

Initially the loading screen was for the Endless Key physical USB for Windows, which had a very long startup time (minutes). A loading page in HTML would show up while Kolibri was loading. Also it had retry/error pages because Kolibri on Windows was a bit unreliable and startup could timeout.

Then after Android was introduced the requirement was to choose between: use data from a USB drive or download from network. Because of limitations in Kolibri we needed to add UI before Kolibri actually starts. Also for Android we needed screens to grant system permissions, ask user to plug a USB, etc. Thus the loading screen was extended and renamed to "welcome screen". Javascript evaluation/injection was used to communicate the Android app with the welcome page. And environment variables were used for passing the selected user choices to Kolibri.

The loading screen is also used on Android before Kolibri starts and during system pause/resume. The Android system can kill the Kolibri process and our resume handler takes care of restarting Kolibri while presenting the loading page. Implementation wise the app flips between 2 webviews, one containing the loading webapp and one containing the Kolibri frontend.

Also the flatpak wrapper needs a loading screen and I'm wondering why not reuse this one https://github.com/endlessm/endless-key-flatpak/pull/44

So yeah I guess the loading screen would need internationalization/localization for the sole 2 messages "welcome to endless key" and "powered by kolibri". Unless we change the design specs to a wordless UI.

A potential performance improvement could be replacing the loading webapp by a bunch of static HTML pages, one per language, as Kolibri does.

dbnicholson (approved):
I mostly skimmed, but this all looks good to me.

packages/loading-screen/src/main.js
```diff
@@ -4,17 +4,10 @@ import { BootstrapVue, IconsPlugin } from 'bootstrap-vue';
 import EkComponents from 'ek-components';
```
dbnicholson, line 4: Is `EkComponents` even needed after this? It looks to me like the only thing used from `EkComponents` is `loading-animation.gif`. It doesn't seem like any of the actual components from `EkComponents` are being used. Which means that the `kolibi.utils.i18n` issue can be sidestepped for now.

dbnicholson: If all that's needed is the gif, then maybe it could be duplicated and the ek-components dependency could be dropped (which would also make the standalone loading-screen smaller). Or you could move the gif into loading-screen and make loading-screen a dependency of the frontend.

manuq: Ah true! If only that GIF asset is left then let's copy it and remove the dependency. I will append a new commiit.

dylanmccall:
Looks good to me, too! :+1:

PR #794, opened 2023-09-09
Only use LZMA compression for channel template zips

dbnicholson:
Having `bundle_zip` default to LZMA compression meant it also applied to `welcome-screen.zip`. That's problematic since `unzip` can't handle that compression scheme. It's also unnecessary as `welcome-screen.zip` is only intended for releases. All of our apps will expand it into some location in their package, so the compression used is irrelevant.

Instead, make the default zlib deflate compression and explicitly request LZMA compression just for the channel templates. Note that the outer `apps-bundle.zip` is actually created with no compression, which is fine since it's entirely for distributing the already compressed channel templates.

See https://github.com/endlessm/endless-key-flatpak/issues/26.

### DIFF
--- a/scripts/_common.py
+++ b/scripts/_common.py
@@ -37,7 +37,7 @@ HIGHLIGHTED_CONTENT_PATH = os.path.join(
 BUNDLE_EXCLUDE_GLOBS = ["*.map"]
 
 
-def bundle_zip(zip_name, compression=zipfile.ZIP_LZMA):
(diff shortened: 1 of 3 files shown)


manuq (approved):
Looks great. It fixes the welcome ZIP for the flatpak issue. At the same time It avoids compressing twice the custom presentations uploaded to github release.

dbnicholson:
I'm going to merge and tag this so we can consume it in the flatpak before we have any more major overhauls from the welcome-screen and internationalization work.

Issue #802, opened 2023-09-12
Android app crashes in migration process


erikos:
The current Android build on internal testing fails to start in the migration phase. Version: Manatee 664-397

[Short_log_Manatee_644-397.log](https://github.com/endlessm/kolibri-explore-plugin/files/12589788/Short_log_Manatee_644-397.log)

dbnicholson:
I [cherry picked this](https://github.com/endlessm/kolibri-explore-plugin/commit/00dff35ae8924a184e487cdf6a3c99944fc904aa) to the `stable` branch and will be releasing 6.46.0 with it.

erikos:
With 6.46.0 the app does not crash in the migration process anymore. Thanks Dan. The app starts fine and gets me to the Discovery page.

PR #803, opened 2023-09-12
Fix BackgroundTask migration

dbnicholson:
There were a few problems here:

* `Job`s have no `fun` attribute. It should be `func`.
* Since a forward code migration was provided but no reverse code migration, Django assumes the the migration is not reversible.
* `job_storage` was not being imported from where it's defined. That would make it difficult to mock in tests.
* The JSON encoded parameters weren't sorted, which would make them impossible to test.

A test is added to verify that the `RunPython` migration code actually works.

Fixes: #802


dylanmccall (approved):
Okay, this looks fine to me :+1:

Issue #804, opened 2023-09-12
Download page confirm button assumes the landing page is in router history


dylanmccall:
If a user navigates _directly_ to the download page and downloads a set of content, the "Show me" button at the end of the process will appear to do nothing. This causes endlessm/endless-key-flatpak#40.

dylanmccall:
As far as I can tell, this happens because, instead of directly navigating to the root page, we are attempting to navigate backwards in the router's history: https://github.com/endlessm/kolibri-explore-plugin/blame/v6.46.0/kolibri_explore_plugin/assets/src/views/DownloadPage.vue#L143-L145.

dbnicholson:
I merged the PR to master, but I'm going to backport it to stable since I'm not sure when master will be ready to make releases from.

dbnicholson:
[Backported](14aaf51479a3446d4757253ec8d11dd29d6003e2) to stable. I'm waiting for the CI to complete successfully then I'll tag a release.

dbnicholson:
Released in kolibri-explore-plugin 6.47.0. This can be verified in https://github.com/endlessm/endless-key-flatpak/issues/40.

vanessa-chang:
Verified passed as the Show me button works

PR #805, opened 2023-09-12
DownloadPage: Reload using the location API

dylanmccall:
Previously, the page was reloaded using a particular effect of the Vue router's go() method, which behaved unexpectedly in some versions of WebKit.

Fixes: https://github.com/endlessm/kolibri-explore-plugin/issues/804

### DIFF
--- a/kolibri_explore_plugin/assets/src/views/DownloadPage.vue
+++ b/kolibri_explore_plugin/assets/src/views/DownloadPage.vue
@@ -141,7 +141,10 @@
         return this.setUpdateInterval();
       },
       onConfirm() {
-        this.$router.go(0);
+        // Reload the page so the app fetches new data from the backend. Upon
+        // reloading, it will redirect to the landing page.
+        // <https://github.com/endlessm/kolibri-explore-plugin/issues/687>
+        window.location.reload();
       },
       updateLoop() {
         return this.updateDownload()


dbnicholson:
This was originally changed in 22dcee453f0993fb91b505afe903676a15e0bf5d. The reason that was done was to actually reload the page because the caching didn't work. I think you either need to clear the JS cached collections (that should work now!) or switch this to `push` so it actually loads the page. I'd vote for `push` but the other reason it might be trying to go back to the beginning of the history is so that the Android back button will exit the app.

manuq:
> This was originally changed in [22dcee4](https://github.com/endlessm/kolibri-explore-plugin/commit/22dcee453f0993fb91b505afe903676a15e0bf5d). The reason that was done was to actually reload the page because the caching didn't work. I think you either need to clear the JS cached collections (that should work now!) or switch this to `push` so it actually loads the page. I'd vote for `push` but the other reason it might be trying to go back to the beginning of the history is so that the Android back button will exit the app.

Yep, this was done as a workaround for caching issues. If clearing the JS cached collections works in current Kolibri then we should do that instead and revert https://github.com/endlessm/kolibri-explore-plugin/commit/22dcee453f0993fb91b505afe903676a15e0bf5d

dbnicholson:
> Yep, this was done as a workaround for caching issues. If clearing the JS cached collections works in current Kolibri then we should do that instead and revert [22dcee4](https://github.com/endlessm/kolibri-explore-plugin/commit/22dcee453f0993fb91b505afe903676a15e0bf5d)

It should work. The point of a60ba3660513f9a68b0a3a652ea626358929c60b is that the backend no longer sets any `Cache-Control` headers that would cause the browser (or Django) to reuse an existing response. Now when the frontend does a fetch with `force: true`, it will result in a request to the backend that actually does something.

Looking over this stuff again, an alternative would be a brute force `clearCache()` for each of `ChannelResource` and `ContentNodeResource` (and maybe `ExternalContentTagResource` and `ContentNodeExtrasResource` to be complete) at the end of the initial download since we know that there have been lots of resource changes. This will ensure that any models or collections in those resources will be fetched from the backend. It would be unfortunate (for example), if the `ChannelResource` collection was refreshed by a forced fetch, but some other resource model or collection was still using stale cached entries.

dylanmccall:
Ah, I see! I can see I was misinterpreting what `router.go(0)` / [`window.history.go(0)`](https://developer.mozilla.org/en-US/docs/Web/API/History/go) actually does. (I think there's another similar API that behaves the way I assumed this one does, jumping to a point in the history stack). But also what the heck Javascript? Sigh. I guess that bad idea predates [pushState](https://developer.mozilla.org/en-US/docs/Web/API/History/pushState).

Anyway, I guess the real problem here is for some reason the webview isn't reloading the page. Looks like a WebKit thing, actually? Poking around in GNOME Web 44, I'm getting the same behaviour when I cal `window.history.go(0)` from a Javascript console anywhere in Kolibri, including at https://key.endlessos.org/. Although it's funny: this works as expected in other contexts, such as at http://example.com/. Must be something about _our_ history object. (Navigating backward and forward in the history stack is fine).

That's weird. I'll see if it reproduces in a newer WebKit as well. But maybe I can change this to call [`window.location.reload()`](https://developer.mozilla.org/en-US/docs/Web/API/Location/reload) instead? It's evilly bypassing Vue, but that's kind of what we're trying to achieve anyway.

dbnicholson:
I think you should just use `push` and not worry about the history or caching right now.

kolibri_explore_plugin/assets/src/views/DownloadPage.vue
```diff
@@ -141,7 +141,10 @@
         return this.setUpdateInterval();
       },
       onConfirm() {
-        this.$router.go(0);
+        // Reload the page so the app fetches new data from the backend. Upon
+        // reloading, it will redirect to the landing page.
+        // <https://github.com/endlessm/kolibri-explore-plugin/issues/687>
+        window.location.reload();
```
dbnicholson, line 147: This is fine as is, but why reload this page, get a redirect and load another page? Why not just load the desired page directly?

dylanmccall:
> I think you should just use `push` and not worry about the history or caching right now.

I might not be following here - is that caching stuff … fixed? My goal here is to keep this behaving as close to how it currently is and carry on, but if it's safe to simply revert 22dcee453f0993fb91b505afe903676a15e0bf5d (which I kind of did originally before I understood the context) then, yes, I'm happy with that :) If the caching stuff is still iffy, then I don't think using `push()` would quite work since that'll just be (effectively) a hash change; still the same state in our Vuex stores.

kolibri_explore_plugin/assets/src/views/DownloadPage.vue
dylanmccall: That redirect happens inside the [download route handler](https://github.com/endlessm/kolibri-explore-plugin/blob/v6.46.0/kolibri_explore_plugin/assets/src/routes/index.js#L20-L25) in the Vue code. It does a couple of async things in [`decideDownload()`](https://github.com/endlessm/kolibri-explore-plugin/blob/v6.46.0/kolibri_explore_plugin/assets/src/modules/topicsRoot/handlers.js#L183-L188) to figure out there is not actually a download happening, but it's all pretty trivial stuff happening in the client - it even does `router.replace()`, so the download page page gets dropped from the history stack.

dbnicholson:
> > I think you should just use `push` and not worry about the history or caching right now.
> 
> I might not be following here - is that caching stuff … fixed? My goal here is to keep this behaving as close to how it currently is and carry on, but if it's safe to simply revert [22dcee4](https://github.com/endlessm/kolibri-explore-plugin/commit/22dcee453f0993fb91b505afe903676a15e0bf5d) (which I kind of did originally before I understood the context) then, yes, that's excellent and I'm happy with that :) If the caching stuff is still iffy, then I don't think using `push()` would quite work since that'll just be (effectively) a hash change; still the same state in our Vuex stores.

I was under the impression that `push` actually loads a page, but it appears that it just adds an entry to the browser history. My fault, this is good enough then.

What's alluded to in 22dcee4 is using `ChannelResource.fetchCollection({ force: true })`, which would bypass the in memory JS cache. However, nothing in the code is actually trying to do that. Reloading the page has the same effect since clearly any in memory JS caching would be gone. My suggestion to fix these kinds of things in the future would be to add a helper that simply calls `ChannelResource.clearCache()` and `ContentNodeResource.clearCache()` when we know that the content has been updated. `*Resource.fetch*({force: true})` is a bit nicer since it would only operate on a particular model or collection, but I feel like it's hard to know when to do that in common code.

Anyways, this is fine as is. :+1:

dbnicholson (approved):
Let's get this going.

PR #812, opened 2023-09-14
Add Latin American Spanish

manuq:
By copying es_ES to es_419.


dbnicholson (changes requested):
I think you also want to update `kolibri_explore_plugin/locale/language_info.json`. I'm not sure the version in our plugin is actually used, but I think it's a good idea to keep it updated.

manuq:
> I think you also want to update `kolibri_explore_plugin/locale/language_info.json`. I'm not sure the version in our plugin is actually used, but I think it's a good idea to keep it updated.

Ah good catch. PR updated.

PR #813, opened 2023-09-14
786 spanish pack

### LINKED ISSUE #786: Select Spanish content pack based on locale
A user with a Spanish System should skip the download options and be directed straight to the Spanish pack. The Spanish starter pack was added in https://github.com/endlessm/endless-key-content-private/issues/90.

This has some overlap with https://github.com/endlessm/kolibri-explore-plugin/issues/748. Currently the starter pack is set externally as an option, but after that the starter pack will determined by the plugin. I think in the short term the way to handle this is to let the `INITIAL_CONTENT_PACK` option default be empty. When the plugin is loaded, look up the system locale. If it's Spanish, set it to `spanish`. Otherwise, fall back to the existing default of `explorer`.

manuq:
Fixes #786

### DIFF
--- a/kolibri_explore_plugin/assets/src/modules/topicsRoot/handlers.js
+++ b/kolibri_explore_plugin/assets/src/modules/topicsRoot/handlers.js
@@ -5,6 +5,7 @@ import router from 'kolibri.coreVue.router';
 import plugin_data from 'plugin_data';
 import { ContentNodeSearchResource } from 'kolibri.resources';
 import { getContentNodeThumbnail } from 'kolibri.utils.contentNode';
+import { currentLanguage, languageIdToCode } from 'kolibri.utils.i18n';
 
 import { ChannelResource, ContentNodeResource } from '../../apiResources';
 import { CarouselItemsLength, SEARCH_MAX_RESULTS, PageNames } from '../../constants';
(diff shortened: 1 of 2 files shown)


manuq:
Note: only the last commit is relevant. This is on top of https://github.com/endlessm/kolibri-explore-plugin/pull/793

jprvita (approved):
Looking at the last commit only, these changes seem to do the right thing.

kolibri_explore_plugin/assets/src/modules/topicsRoot/handlers.js
```diff
@@ -172,6 +173,21 @@ export function decideWelcome(store) {
   );
 }
 
+export function decidePackSelection(store) {
+  store.commit('CORE_SET_PAGE_LOADING', true);
+
+  if (currentLanguage.startsWith('es')) {
```
dbnicholson, line 179: I think this should use `languageIdToCode` rather than this type of matching since it would inadvertently match a language code like `esk` (even though there isn't one).

manuq: Ouch! Very good point @dbnicholson

manuq: > I think this should use `languageIdToCode` rather than this type of matching since it would inadvertently match a language code like `esk` (even though there isn't one).

I have changed the conditional to use `languageIdToCode` and rebased with master.

pwithnall (approved):

PR #814, opened 2023-09-14
Update to kolibri-tools 0.16.0-dev.4

dbnicholson:
This has several fixes that we upstreamed and hopefully fixes some strange linting issues in the frontend.


dbnicholson:
Wait, maybe not just yet. I ran `yarn run lint-frontend` and it gave a pile of errors that I don't understand:

```
Kolibri Linter: 
Kolibri Linter: Linting errors for kolibri_explore_plugin/assets/src/views/DiscoveryPageEkIguana.vue
Kolibri Linter: TypeError: Class extends value undefined is not a constructor or null
```

dbnicholson:
> Wait, maybe not just yet. I ran `yarn run lint-frontend` and it gave a pile of errors that I don't understand:
> 
> ```
> Kolibri Linter: 
> Kolibri Linter: Linting errors for kolibri_explore_plugin/assets/src/views/DiscoveryPageEkIguana.vue
> Kolibri Linter: TypeError: Class extends value undefined is not a constructor or null
> ```

The problems go away if I borrow the update to vue-cli 5/webpack 5 from #777.

pwithnall:
> > Wait, maybe not just yet. I ran `yarn run lint-frontend` and it gave a pile of errors that I don't understand:
> > ```
> > Kolibri Linter: 
> > Kolibri Linter: Linting errors for kolibri_explore_plugin/assets/src/views/DiscoveryPageEkIguana.vue
> > Kolibri Linter: TypeError: Class extends value undefined is not a constructor or null
> > ```
> 
> The problems go away if I borrow the update to vue-cli 5/webpack 5 from #777.

Yeah, I think if you’re doing anything to do with this you’ll want the full stack of changed dependencies from #777.

dbnicholson:
I split out those commits from #777 and added the change to build with `--no-module`. I tested this locally and I think it's good on its own. It will cause a small conflict with #793 since it operates on the `welcome-screen` package. I'm happy to rebase if that lands soon.

scripts/bootstrap.sh
```diff
@@ -27,5 +27,4 @@ done
 pipenv install --dev --deploy "${pipenv_install_args[@]}"
 pipenv run nodeenv -p --node=16.14.0
 pipenv run npm install -g yarn@1.22.19
-pipenv run yarn global add @vue/cli@v4.5.15
```
pwithnall, line 30: This command appears in a couple of workflows in `.github/workflows/`. Do we want to drop it from there too?

Also, `scripts/build_template_overrides.py` seems to call `vue-cli-service`. Should it be modified to call it through `yarn`, or are we relying on the fact that `build_template_overrides.py` itself is run by yarn, so should have the right environment to find `vue-cli-service` anyway?

pwithnall (approved):
One comment/question but this looks eminently mergeable.

scripts/bootstrap.sh
dbnicholson: I'm not hitting the CI errors locally, so this was a blind attempt to try to fix it. That said, I would like to clean this up appropriately so that CI runs the same setup developers have. Good catch on calling `vue-cli-service` from inside that script. Although it's probably being run from yarn 99% of the time, I think I'll make sure it's called via yarn.

pwithnall: >  That said, I would like to clean this up appropriately

Yeah, it’s always nagged me that `setup-bootstrap.sh` exists but isn’t used by the CI workflows. Either it should exist and be used everywhere, or not exist. If that’s part of what you’re referring to? If not, I’ve just had a good rant anyway.

dbnicholson: Exactly. I think it arrived there organically, but yeah, it bothered me, too.

dbnicholson:
It seems to be working now. I think the sass problems magically went away because loading-screen got completely decoupled from ek-components in #793. Still waiting for the apps-bundle build to complete (my god is it slow to webpack 60 apps), but I think this should be good to go.

pwithnall (approved):
LGTM

PR #816, opened 2023-09-15
Extract all backend locale messages

dbnicholson:
Extract and compile messages for all locales supported by the backend. This makes more sense for Endless where we need all the PO files to be up to date so that translations can be made in them. This does generate files for all the locales Kolibri supports, however. We have no intention of supporting the majority of them, but it's simpler than having a custom Django settings file that overrides `LOCALE_PATHS` to our supported subset.


manuq (approved):
Looks like it's doing the job for the single message. Merging.

pwithnall (commented):
LGTM

Issue #817, opened 2023-09-15
ek-components createTranslator usage broken


dbnicholson:
It looks like #777 was good enough for building but not good enough for runtime. I tried running the dev server and it just spins on the loader with this in the console:

```
Uncaught TypeError: Cannot read properties of undefined (reading 'createTranslator')
    at ./packages/ek-components/src/constants.js (constants.js:1:1)
    at __webpack_require__ (bootstrap:19:1)
    at ./node_modules/kolibri-tools/node_modules/vue-loader/lib/index.js??vue-loader-options!./packages/ek-components/src/components/EkCard.vue?vue&type=script&lang=js& (EkCard.vue:40:1)
    at __webpack_require__ (bootstrap:19:1)
    at ./packages/ek-components/src/components/EkCard.vue?vue&type=script&lang=js& (EkButtonsBar.vue:1:1)
    at __webpack_require__ (bootstrap:19:1)
    at ./packages/ek-components/src/components/EkCard.vue (EkButtonsBar.vue:95:1)
    at __webpack_require__ (bootstrap:19:1)
    at ./packages/ek-components/src/main.js (constants.js:211:1)
    at __webpack_require__ (bootstrap:19:1)
```

That makes sense because `window.kolibri` is set dynamically by the frontend. In the console, `window.kolibri` is currently `undefined`. Somehow this will need to be delayed until it can be resolved. I don't know how to do that.

CC @pwithnall @manuq

pwithnall:
Uff, this is my fault. I rushed too much on Friday and didn’t test the final iteration of my branch.

Possible solutions which come to mind:
 * Add some synchronisation between `main.js` and when it imports `constants.js`, but that’s hard (impossible?) because it re-exports `constants.js`.
 * Move `ekComponentsConstantsStrings` inside a function so that `window.kolibri.createTranslator` doesn’t have to be dereferenced until the first translated string is actually needed.

dbnicholson:
> Uff, this is my fault. I rushed too much on Friday and didn’t test the final iteration of my branch.

Well, I was certainly pushing to get something in, so don't take all the blame. It would help if we even a minimal smoke test that ran.

manuq:
[...]

PR #818, opened 2023-09-18
ek-components: Delay calling createTranslator() until actually needed

pwithnall:
`createTranslator()` in ek-components now relies on `window.kolibri` being available, and that’s only initialised as part of the app setup, which is after all the imports are processed.

This causes the `createTranslator()` calls to fail at import time as `window.kolibri` is undefined.

Fix that by wrapping them in an internal function, which is only called when translated strings are actually needed. Rely on `createTranslator()` to cache the translator between calls.

Fixes: #817


pwithnall:
#819 is also needed for the content apps to work

manuq (approved):
The fix makes sense and I have verified by running the branch!

PR #820, opened 2023-09-18
ci: Use setup-node to cache node.js environment in CI

pwithnall:
This is a partial revert of a06ac87382095f882a89e37794b4303a6b8a4caf. It turns out that the caching that setup-node does is too important to lose: without it, dodgy/slow downloads from node.js dependency servers cause spurious CI failures for us.


pwithnall:
See https://github.com/endlessm/kolibri-explore-plugin/pull/782#issuecomment-1723508031 for some discussion about the cause/symptoms of this.

scripts/bootstrap.sh
```diff
@@ -4,13 +4,17 @@
...
 
 while :; do
   case $1 in
     --ci)
       # Prevent the Pipfile.lock from being updated automatically. Just use
       # the pinned dependencies.
       pipenv_install_args+=( "--ignore-pipfile" )
+      # And assume that node has already been set up, as the CI system can
+      # cache that.
+      skip_nodeenv=1
```
dbnicholson, line 17: Nitpicky, but if you set this to `true`, then below it can be handled as `$skip_nodeenv ||` without the string comparison.

dbnicholson (changes requested):
That looks good, but you can add a commit that moves the `-e` in the shebang line to an explicit `set -e` command so that the script will reliably fail no matter how it's executed?

pwithnall:
> That looks good, but you can add a commit that moves the `-e` in the shebang line to an explicit `set -e` command so that the script will reliably fail no matter how it's executed?

oops, I completely forgot about that part of your comment in the other issue. Fixed!

dbnicholson (approved):
Looks good. When it gets past the critical part on the CI I'll merge it.

PR #821, opened 2023-09-18
Support release candidates

dbnicholson:
This adds some support for creating release candidates with `bumpversion`. Since `pip` considers a package with an `rc<N>` suffix to be a pre-release, it won't be installed without specifying the version or using `--pre`. This should allow us to get some testing on a Narwhal RC without making it generally available.

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -2,3 +2,7 @@
 current_version = 6.44.0
 commit = True
 tag = True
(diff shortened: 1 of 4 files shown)


starnight (approved):
The RC is an good idea.  But ya, the versioning issue, `RC` should be older.  However, the same version string with `RC` will be greater than without `RC`.  Human intervention is a kind of trade off.

PR #822, opened 2023-09-18
Save wheel and tarball as artifacts

dbnicholson:
We always build the wheel and tarball for every build. Rather than throwing them away, store them as artifacts. After that, rearrange the release into a separate job that downloads the built artifacts and includes them as release artifacts.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -54,36 +54,33 @@ jobs:
       - name: Build libs
         run: pipenv run yarn build:libs
 
+      - name: Upload loading-screen.zip
+        uses: actions/upload-artifact@v3
+        with:
+          name: loading-screen.zip
+          path: packages/loading-screen/loading-screen.zip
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -128,17 +125,75 @@ jobs:
...
+      - name: Get wheel filename
+        id: whl-filename
+        run: echo "whl-filename=$(ls dist | grep '\.whl$')" >> $GITHUB_OUTPUT
+
+      - name: Upload wheel
+        uses: actions/upload-artifact@v3
+        with:
+          name: ${{ steps.whl-filename.outputs.whl-filename }}
+          path: dist/${{ steps.whl-filename.outputs.whl-filename }}
+          if-no-files-found: error
```
starnight, line 137: The whl helps non-released version test.

starnight (approved):
LGTM!

Issue #770, opened 2023-08-28
Add translation calls to welcome-screen


pwithnall:
Currently the welcome-screen package doesn’t support localisation/translation. It hard-codes English strings in at least:
 * `LoadingInitial.vue`
 * `Welcome.vue`
 * `PackSelection.vue`
 * `PackReady.vue`

Those need to be changed to use a `$trs` array.

manuq:
This won't be needed since we are moving the welcome screens to the plugin. However, the de-hardcoding in this PR will be useful.

dbnicholson:
@manuq gave a spin with the translations from #845 and it worked from a browser. Further testing will be done for #765, I believe.

PR #824, opened 2023-09-19
app: Mark some more strings as translatable

pwithnall:
These came from the welcome-screen, and can now be translated as part of the main app.

Helps: #770


pwithnall:
I’ve split this out from #775 as these translation changes should be able to land fairly uncontroversially.

manuq (approved):
Excellent.

PR #825, opened 2023-09-19
More CI fixes

dbnicholson:
Since 124542c7dd3e2ff93e1f472fe3db641beb12627c, the python wheel file is created and saved as an artifact for every build. Unfortunately, the version created by `setuptools_scm` was wrong because `actions/checkout` does a shallow clone by default. While here, fix up a couple other minor issues in the workflows.


.github/workflows/main.yml
```diff
@@ -14,18 +14,15 @@ on:
...
-      - name: Checkout from none-workflow_call
-        uses: actions/checkout@v3
-        if: ${{ !startsWith(inputs.tagname, 'v') }}
-
-      - name: Checkout from workflow_call
+      - name: Checkout
         uses: actions/checkout@v3
-        if: startsWith(inputs.tagname, 'v')
         with:
+          # If the tagname is empty, checkout falls back to github.ref.
```
manuq, line 25: :+1:

manuq (approved):
Great good amount of improvements on top of the fix.

PR #826, opened 2023-09-19
Change KolibriApi translators property to static

manuq:


dbnicholson:
Quite possibly this is https://stackoverflow.com/questions/60046847/eslint-does-not-allow-static-class-properties? I can't tell what version of eslint we're actually using, but we're inheriting from the kolibri-tools `.eslintrc.js` and that has `ecmaVersion: 2020`. See https://eslint.org/blog/2021/10/eslint-v8.0.0-released/.

manuq:
Eslint fails because class properties aren't considered until eslint v8. Really Javascript?!

PR #827, opened 2023-09-19
Change KolibriApi translators property to global

manuq:
This fixes issues when KolibriApi class is used without instantiation, for which translators is undefined.


dbnicholson (approved):
It's not pretty, but it makes sense.

PR #829, opened 2023-09-19
Content modal: Prevent right click

### LINKED ISSUE #828: Prevent right click in content viewer
By design request, the browser right-click menu should not be available in the content viewer.

manuq:
As per design request.

Fixes #828


manuq:
Note that master won't show any content because of errors until https://github.com/endlessm/kolibri-explore-plugin/issues/817#issuecomment-1725936890 is fixed.

GeorgesStavracas (approved):

PR #832, opened 2023-09-20
Import Spanish translation from spreadsheet

manuq:


dbnicholson (approved):
Looks good. It's a bit unfortunate that the `es_ES` locale still has the original translation. I kinda wish Kolibri just had a generic `es` translation, but maybe they're different enough. What do I know?

manuq:
> Looks good. It's a bit unfortunate that the `es_ES` locale still has the original translation. I kinda wish Kolibri just had a generic `es` translation, but maybe they're different enough. What do I know?

Oh I forgot to `git add` the `es_ES` file. I'll open a new PR.

PR #835, opened 2023-09-20
Include MO files in distribution

dbnicholson:
a8f1ab8f64be9e58db4a88ef8ebdccc513039d92 removed the MO files from git, but sadly that also removes them from the distribution.


pwithnall:
image

Issue #765, opened 2023-08-24
Initial Spanish translation


pwithnall:
As part of #760, we want an initial translation done for kolibri-explore-plugin.

pwithnall:
Here are the English versions of the strings. They come in three files because of how kolibri-explore-plugin is split up between frontend and backend, and different frontend modules. `django.txt` is actually a [PO file](https://www.gnu.org/software/gettext/manual/html_node/PO-Files.html) but I had to rename it to get GitHub to accept it.

[django.txt](https://github.com/endlessm/kolibri-explore-plugin/files/12432344/django.txt)
[kolibri_explore_plugin-messages.csv](https://github.com/endlessm/kolibri-explore-plugin/files/12432345/kolibri_explore_plugin-messages.csv)
[welcome-screen-messages.csv](https://github.com/endlessm/kolibri-explore-plugin/files/12432346/welcome-screen-messages.csv)

pwithnall:
@jofilizola, the English versions of the strings for the kolibri-explore-plugin are above for you to take a look at.

jofilizola:
@pwithnall I am organizing the translation file with screenshots, for people translating it might not be very obvious where things sit, since they are not so familiar with all of the UI element. There are a few things that I need help with, all marked in the document:
- Error messages that I can't force to appear, would you be able to get me screenshots? YELLOW
- Strings that are in the UI with no identifiers listed RED
-  Strings that I could not find where the in the UI they live PURPLE (are these `fade.downloadCompletedLabel` related to the individual content download, card UI?)
- Strings that come straight from Kolibri as they are, the quizzes components are marked in gray, can we use the ones from Kolibri?

Here's [the document](https://docs.google.com/document/d/16spohQ1cssR3bkaAL0Qp71AGoK8m2CPV3x8plRIhSY8/edit), if you see something wrong there, please let me know!

Thanks!!

pwithnall:
Left some comments and some screenshots (where I can easily produce them) in the document.

pwithnall:
Here’s an updated version of the message files, including a lot more messages from other parts of the UI which were previously untagged for translation.

This should now be pretty much the full set of strings to translate.

[LC_MESSAGES.zip](https://github.com/endlessm/kolibri-explore-plugin/files/12597820/LC_MESSAGES.zip)

jofilizola:
[...]

PR #839, opened 2023-09-21
i18n: Update EN JSON files

manuq:
Helps #765.


pwithnall:
Presumably this is just the result of running `yarn run i18n-download-translations`?

manuq:
I guess we can merge this sole commit first, then continue updating the Spanish files.

PR #842, opened 2023-09-21
template-ui: Use v-html to handle <strong> in translation

dbnicholson:
In vue-intl, `formatMessage` always returns a string, which means any embedded HTML in the translation can't be displayed. Use `v-html` to insert the translation as raw HTML.

Fixes: https://github.com/endlessm/endless-key-flatpak/issues/77

### DIFF
--- a/packages/template-ui/src/components/ChannelFooter.vue
+++ b/packages/template-ui/src/components/ChannelFooter.vue
@@ -12,9 +12,8 @@
       </template>
       <template #right>
         <div v-if="isEndlessApp" class="endless-seal text-right">
-          <div class="mb-1">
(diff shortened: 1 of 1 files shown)


manuq:
I thought that was fixed in 55e4bbcc95cc18644ff65daf1484318bce8d76e6 but probably wasn't tested. Kolibri docs https://kolibri-dev.readthedocs.io/en/develop/i18n.html#icu-message-syntax point to ICU syntax and I see `<boldThis>` in there https://formatjs.io/docs/core-concepts/icu-syntax/#rich-text-formatting . Maybe we can use that?

manuq:
I gave it a quick test with `boldThis` but it didn't work. So yeah let's use `v-html`.

PR #847, opened 2023-09-22
Merge es translations

manuq:
Helps #765


dbnicholson (approved):
Looks good for the short term. Also why we need the CSVs in the repo so we can actually see when the baseline moves. Essentially, we're missing the gettext equivalent of `msgmerge` which updates the PO files from the POT file. Apparently Crowdin does that for LE.

PR #849, opened 2023-09-22
Spanish translation: add missing message

manuq:


pwithnall (approved):

packages/ek-components/src/constants.js
```diff
@@ -80,6 +80,20 @@ function getEkComponentsConstantsTranslator() {
       message: 'I like to experiment with a bit of everything.',
       context: 'Subtitle of a content pack',
     },
+
+    // For filters:
+    mediaFilterLabel: {
+      message: 'learning activity',
```
dbnicholson, line 86: This was capitalized `Learning activity` before.

dbnicholson (changes requested):

packages/ek-components/src/constants.js
manuq: Yeah, it was inconsistent before and I made them all consistently de-capitalized now. In any case we can ask Joana later.

PR #852, opened 2023-09-26
Handlers: Fix root URL redirect

### LINKED ISSUE #799: The channel descriptions in the discovery page is only shown when the app is launched
Test package: https://github.com/endlessm/endless-key-flatpak/actions/runs/6153210584 & Manatee 6.44-397

Each time you launch the app, you can see the channel description is there under the title in each card. (except for the feature channels)
However, if you reload the app, swtich to different tab and return to the discovery page or get into any channel then exit back to the discovery page, the desciption will be disappeared.

[Screen recording 2023-09-12 5.01.59 PM.webm](https://github.com/endlessm/kolibri-explore-plugin/assets/31236978/62a9a280-8a1e-4014-98d0-1a45f74bb0e0)

manuq:
The redirect should happen after the loading state is set. Each handler sets loading to true at the very beginning and to false at the very end. But because of the redirect, the TOPICS handler (showChannels) was starting in the middle of the ROOT handler (decideWelcome). Thus the state was mixed up and not to be trusted.

Fixes #799

### DIFF
--- a/kolibri_explore_plugin/assets/src/modules/topicsRoot/handlers.js
+++ b/kolibri_explore_plugin/assets/src/modules/topicsRoot/handlers.js
@@ -162,10 +162,10 @@ export function decideWelcome(store) {
             store.commit('CORE_SET_PAGE_LOADING', false);
             store.commit('CORE_SET_ERROR', null);
           } else {
-            console.debug('Welcome: Redirecting to Explore page...');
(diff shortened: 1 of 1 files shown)


dbnicholson (approved):
Makes sense. I was no longer able to reproduce the issue with this change.

Issue #778, opened 2023-08-30
Add collection downloader tests


dbnicholson:
Since #743 there's a very basic backend test suite that correctly sets up Kolibri and Django. What should really be tested is the collection downloader code in `collectionviews.py`. Unfortunately, this is heavily dependent on running tasks to import channels and content. Potentially you could mock parts of it, but you wouldn't get very far without:

* Fake collections data.
* Actual channel databases since the downloader goes from importing channels to importing content from those channels immediately. These are sqlite databases. Without actual databases, you'd have to mock the download and then insert a bunch of objects in the content database.
* Running tasks. It might be possible to mock the tasks being executed, but it would be a lot easier to actually run the tasks. It's complicated to run a `Worker` since the job storage database is manage with sqlalchemy instead of Django. While pytest-django handles the Django databases nicely, you're on your own with sqlalchemy.

I did most of this, but it's still in rough shape. It'll need a bit more polishing to be generally useful.

PR #854, opened 2023-09-26
851 add spanish pack backend

### LINKED ISSUE #851: Spanish starter pack can not be downloaded
With the latest plugin release 7.1 we have the UI translated to Spanish. The initial animation and the Welcome screen is in Spanish. 

When proceeding in the Welcome screen the Spanish starter pack is not downloaded though - the progress stays at zero percentage.

The URL that is tried is http://127.0.0.1:58297/es-es/explore/#/download/spanish/0001

The starter pack is defined in https://github.com/endlessm/endless-key-collections/blob/main/json/spanish-0001.json - other starter packs have the same naming format
https://github.com/endlessm/endless-key-collections/blob/main/json/artist-0001.json

The UI itself is translated fine. This is just about the starter pack not being downloaded correctly.

manuq:
Fixes #851


dbnicholson (approved):
I should really finish off #778 so we can catch these earlier.

PR #859, opened 2023-09-27
Feedback form: Use different URL for Spanish

### LINKED ISSUE #858: Use different feedback link for Spanish
The app should link to a different URL for the feedback form when running in Spanish. Bonus points for preparing this to other languages and use the English URL as fallback.

Spanish link: https://endlessos.typeform.com/to/ohjef1FG 

From Joana: Other languages can be supported by the ai translations that Typeform provides (added to the English form), if the user system's is language that has the ai translation support it will appear translated!

Image

manuq:
As the original feedback URL, it can be overriden with a setting. Useful for changing the ES feeback URL in the online instance.

Unfortunately the plugin option can't be generalized to other languages because the INI file should be kept simple (ie no mapping data type).

Fixes #858

### DIFF
--- a/kolibri_explore_plugin/assets/src/components/DiscoveryNavBar.vue
+++ b/kolibri_explore_plugin/assets/src/components/DiscoveryNavBar.vue
@@ -65,6 +65,7 @@
 
 <script>
 
+  import { currentLanguage, languageIdToCode } from 'kolibri.utils.i18n';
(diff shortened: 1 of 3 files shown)


dbnicholson (approved):
It's unfortunate that this will require a new option for every language. It would be nice if we could use a generic target that redirected based on the user's language.

PR #860, opened 2023-09-28
Routing: Fix one more page mounting early

### LINKED ISSUE #799: The channel descriptions in the discovery page is only shown when the app is launched

manuq:
Like in 4deba013. In this case the Topics (home) page could be mounted early when redirected from the Download page. Because the redirect was happening after setting "loading" and "error" to false.

Fixes #799

### DIFF
--- a/kolibri_explore_plugin/assets/src/modules/topicsRoot/handlers.js
+++ b/kolibri_explore_plugin/assets/src/modules/topicsRoot/handlers.js
@@ -215,9 +215,9 @@ export function decideDownload(store, grade, name) {
             });
           } else {
             console.debug('Conditions not met to download, assuming as completed.');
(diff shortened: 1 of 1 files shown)


dbnicholson (approved):
I don't totally get it, but I'm sure you're right :stuck_out_tongue:

Issue #868, opened 2023-10-04
Include platform and locale data in Typeform hidden fields


cassidyjames:
We often wonder what platform and locale feedback is coming from. It turns out Typeform supports hidden fields with a URL parameter-like hash, so we should be able to include this easily:

https://www.typeform.com/help/a/using-hidden-fields-360052676612/

We may want to include the app version in there while we're at it. I'm not sure if there's other automatic information we'd like to include.

manuq:
Sounds simple enough! I was wondering about our URL redirects, but all seems fine: If I navigate to https://endlessos.org/key-feedback-es#name=manuel&foo=bar I end up in https://endlessos.typeform.com/to/ohjef1FG#name=manuel&foo=bar

manuq:
I have a PR with the language and also the Kolibri version + the Endless Key version as displayed in our About modal (eg. "Manatee v6.46.0"). But still draft because I'm missing the platform. I see Kolibri perseus plugin is using `navigator.platform` but this document says it's outdated? https://developer.mozilla.org/en-US/docs/Web/API/Navigator/platform

manuq:
After a second thought, we can leave platform as "Unknown" for now. The rest is working. When I try in localhost I am redirected to this URL: https://endlessos.typeform.com/to/ohjef1FG?typeform-source=localhost#language=es-419&platform=Unknown&kolibriVersion=0.16.0b1

manuq:
Actually, the way to go with platform is to pass "KOLIBRI_PROJECT" to the hidden fields. This is [what Kolibri uses for telemetry too](https://github.com/learningequality/kolibri/blob/00010153e3c3329acb3540166819a1c68816d203/kolibri/utils/options.py#L620-L627). For example we pass "endless-key-android" in the Android app: https://github.com/endlessm/kolibri-installer-android/blob/a575aaeeff7904eb51ee218c90f83ff585aca5f9/src/kolibri_android/kolibri_utils.py#L69

manuq:
The Windows app is passing 'endless-key-windows' too. https://github.com/endlessm/endless-key-app/blob/951e1555525c71499a8c1b0c044342a5b410c2bc/kolibri-electron/src/index.js#L28

@dylanmccall or @starnight can you confirm if the Flathub app is passing KOLIBRI_PROJECT as the Android and Windows apps? I couldn't find it.

starnight:
[...]

PR #869, opened 2023-10-04
Feedback form: Add hidden fields

manuq:
Helps #868


manuq:
Ready for review. It will still say "Unknown" for platform but the rest is working and it's better than nothing.

kolibri_explore_plugin/assets/src/components/DiscoveryNavBar.vue
```diff
@@ -169,6 +186,17 @@
           this.$store.commit('topicsRoot/SET_SHOW_SIDE_NAV', true);
         }
       },
+      getBuildInfo() {
+        const buildInfo = urls.static(`build-info.json`);
```
dbnicholson, line 190: Where does this come from?

kolibri_explore_plugin/assets/src/components/DiscoveryNavBar.vue
```diff
@@ -99,11 +101,23 @@
         return assets.EndlessLogo;
       },
       feedbackUrl() {
+        const hiddenFormFields = new URLSearchParams({
+          language: currentLanguage,
+          kolibriVersion: window.kolibriCoreAppGlobal.version,
```
dbnicholson, line 106: Should the explore version also be included here?

dbnicholson (commented):
Makes sense to me, although I don't really understand how the key version is gathered.

Issue #863, opened 2023-10-02
Run content pack selection even if channels exist


dbnicholson:
Currently the pack selection is skipped if any channels exist under the assumption that only happens once the pack has been selected. That prevents the content from being preseeded as we're trying to do [in the image builder](https://github.com/endlessm/eos-image-builder/issues/117#issuecomment-1738094529) with the result that the Discovery page is empty.

manuq:
So we should persist the pack selection somewhere. Or use the lack of download state as evidence of "pack hasn't been selected". Maybe the latter, otherwise current users will get a pack selection once again.

manuq:
OK, I have found an issue with the previous PR when a download was attempted in a new session even after the download was completed in a previous session. Not good! This new PR should fix it: https://github.com/endlessm/kolibri-explore-plugin/pull/870

This makes the `get_should_resume` API a bit irrelevant. But honestly I don't want to make further changes without a test suit.

dbnicholson:
I merged that and it got released in 7.7.0. If there are any further issues, I think we should open a new issue.

dbnicholson:
I opened #874 about using the database to persist the download state and pack selection in a more reliable way.

vanessa-chang:
So far, no issues were found on the latest Android and Windows package for resuming the content download.

I will close this issue first.

PR #870, opened 2023-10-05
863 followup

manuq:
Helps #863


dbnicholson (approved):
This seems to work better, although the first time I tried it made me go through the download again.

manuq:
> This seems to work better, although the first time I tried it made me go through the download again.

Ouch.. I will have to test further.

dbnicholson:
It might be OK since I might have been in a funky state while I was doing testing for the caching issue. As in, I cleared all the site data at one point while I was looking at whether the browser was using cached objects or not. That's a big reason why I'd want to store the download state in the database, but it's a pretty niche use case. I'll try again today to see if I can reproduce it.

Issue #725, opened 2023-08-02
Update color scheme / theme 


jofilizola:
**Application on the Channel:** 

![Image](https://github.com/endlessm/endless-key-content-private/assets/51095367/4c4e97bf-1c43-48d6-95ab-a6b3f90f6920)

**Application on Discovery:**

![Image](https://github.com/endlessm/endless-key-content-private/assets/51095367/1037804d-4db7-412a-a279-1d9d35030dd9)

Header Layout proposal...

![Image](https://github.com/endlessm/endless-key-content-private/assets/51095367/00a47dcc-ebeb-4d35-abd0-d7688819c06a)

_Originally posted by @jofilizola in https://github.com/endlessm/endless-key-content-private/issues/5#issuecomment-1585181012_

jofilizola:
Color scheme: https://www.figma.com/file/ItL7d0N9yJALcu12S50dQn/EK-Ladybird?type=design&node-id=401%3A168338&mode=design&t=RFElpqUkshbeu6r1-1

manuq:
The PR above is from a livecoding session with @jofilizola. A quick and significant change that helps this issue: change the topic cards.

jprvita:
Since we are changing the "Get Endless Key" button in #865, we will need to deploy it to key.endlessos.org.

erikos:
Kicked off https://github.com/endlessm/kolibri-explore-plugin/actions/runs/6404240647 to have the latest plugin changes in.

manuq:
Reopening. I've found that the download progress bar has regressed. The PR above makes it orange again.

vanessa-chang:
Tested Chromebook 7.8-415

Below items are checked:
- Progress bar: Orange
- Topic cards: Done
- Topic pills: Done

PR #871, opened 2023-10-05
Download page: Make progress bar orange again

manuq:
This has regressed in the theme changes.

Helps #725


dbnicholson (approved):
Makes sense.

Issue #872, opened 2023-10-05
Channel view broken with Kolibri 0.16.0-beta6


dbnicholson:
This was originally https://github.com/endlessm/kolibri-installer-android/issues/190. Currently, `AppView` is [using `metadata_cache`](https://github.com/endlessm/kolibri-explore-plugin/blob/2f6959e20140cfab7723b4a87215c9ad4a1a8d44/kolibri_explore_plugin/views.py#L64). That causes an exception in 0.16.0-beta6 because `metadata_cache` was changed to [handle the cache value itself](https://github.com/learningequality/kolibri/commit/afa58e8c62bbcd960e0071dc88d1a6540b0d7ad8) instead of letting the Django middleware [do it](https://github.com/django/django/blob/c669cf279ae7b3e02a61db4fb077030a4db80e4f/django/middleware/cache.py#L100). The problem is that `AppView` is a regular Django `View` without a `render` callable.

That's a Kolibri bug that I'll report upstream. Fortunately, it should be easy to fix downstream as that's abusing `metadata_cache`, anyways.

dbnicholson:
This was released in 7.7.0 and is in all the apps.

vanessa-chang:
Verified passed in Chromebook 7.8-415 | Flatpack v0.7 10-16 | Windows 7.8.1.0

PR #873, opened 2023-10-05
Fix channel view with Kolibri 0.16.0-beta6

dbnicholson:
The Kolibri `metadata_cache` decorator is tied to the content lifetime,
but that has nothing to do with the channel app data. Furthermore,
`metadata_cache` in Kolibri 0.16.0-beta6 is broken for this usage since
regular `View` classes don't have `render` functions.

Instead, simply use the `cache_control` decorator to set `max-age=600`.
This caches the responses for 10 minutes. Even though these are likely
very static, I don't want to set these longer for now since we have no
control over when the apps bundle gets updated.

Fixes: https://github.com/endlessm/kolibri-explore-plugin/issues/872

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -21,7 +21,7 @@ pytest-django = "~=3.10"
 nodeenv = "==1.3.3"
 ipdb = "==0.13.2"
 # FIXME: Move to "~=0.16.0" when it's released.
-kolibri = { file = "https://github.com/learningequality/kolibri/releases/download/v0.16.0-beta1/kolibri-0.16.0b1-py2.py3-none-any.whl" }
+kolibri = { file = "https://github.com/learningequality/kolibri/releases/download/v0.16.0-beta6/kolibri-0.16.0b6-py2.py3-none-any.whl" }
 # Needed for old requests bundled with kolibri to work with newer python.
 cryptography = "~=37.0"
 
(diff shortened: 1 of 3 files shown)


manuq (approved):
Looks good! Should we add a comment to link the upstream issue?

dbnicholson:
> Looks good! Should we add a comment to link the upstream issue?

We could, but I didn't actually file it yet 😜. And actually when I thought about it a little more, I think I understand why it's that way. That decorator is for content API views. That means they're rest framework viewsets that always have render functions.

We were completely misusing it for the side effect of a long cache control max age. So, it might be nice to have it work for generic views or throw a nicer exception, but I think it could very easily be wontfixed, too.

dbnicholson:
I opened https://github.com/learningequality/kolibri/issues/11362 now, so I'll add that to the commit message. I also think we can do very long caching after looking closer at the URLs. They always contain the explore plugin version, so it should be safe to use very long caching so long as we always update the apps bundle with the plugin.

dbnicholson:
@manuq I updated the commit message and bumped the max age up to 1 week. Do you want to take a look or should I merge? I'd like to get this released today since I believe both the Flatpak and the Android app are on 0.16.0-beta6 and currently broken.

dbnicholson:
I'm merging so I can make a release.

PR #885, opened 2023-10-18
Trigger kolibri-installer-android with dispatch event

dbnicholson:
Since the move to Gradle, kolibri-installer-android requires explicitly changing the explore plugin version before using the new release. Like endless-key-app, a workflow has been added there to handle the custom `kolibri-explore-plugin-release` repository dispatch event. The Jenkins triggering is handled there and is no longer needed in this repo.

After this the `JENKINS_USERNAME` and `JENKINS_PASSWORD` GitHub Actions secrets should be removed.

Helps: https://github.com/endlessm/kolibri-installer-android/issues/192

This depends on https://github.com/endlessm/kolibri-installer-android/pull/193 and shouldn't be merged until that lands.


dbnicholson:
https://github.com/endlessm/kolibri-installer-android/pull/193 is now ready for review, but this shouldn't be merged until that is.

starnight (approved):
Thanks!!!

PR #886, opened 2023-10-19
build: Copy setup.cfg data to pyproject.toml

pwithnall:
This is needed to use newer versions of kolibri-tools, which will read configuration from `pyproject.toml` in preference to `setup.cfg`, if the former exists.

See https://github.com/learningequality/kolibri/issues/11146

For now, keep the old configuration present in `setup.cfg` as well, in case people are using an older version of kolibri-tools. That can be dropped once we bump our dependency.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -77,3 +77,12 @@ testpaths = 'kolibri_explore_plugin/test'
 # FIXME: Try to narrow the deprecation warnings to only those caused by
 # Kolibri.
 filterwarnings = 'ignore::DeprecationWarning'
+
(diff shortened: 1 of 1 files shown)


pwithnall:
This is actually only needed once kolibri-explore-plugin depends on a new-enough version of kolibri-tools that contains https://github.com/learningequality/kolibri/issues/11146. At that point, our i18n would break without this PR, because kolibri-tools will read `pyproject.toml` in preference to `setup.cfg`, and won’t merge keys between the two.

Submitting this PR now, even though it’s not actually needed yet, to avoid that potential trip hazard in future.

manuq:
Thank you @pwithnall

PR #887, opened 2023-10-19
docs: Various updates to the i18n documentation based on recent work

pwithnall:
Update the documentation with the latest knowledge and workflow based on how we actually ended up doing the Spanish translation.


dbnicholson (approved):
Bravo to you for figuring all of that out. :clap:

Issue #891, opened 2023-10-20
Download different extra metadata for Spanish


manuq:
According to a new requirement: https://docs.google.com/spreadsheets/d/1tTmp_CqjGHezSw_cpDMKjJZUTyTSykkGlzWpStzCs0U/edit#gid=0

There should be extra metadata imported for Spanish. And as I understand, this should be different than the extra metadata for English.

Currently: the metadata of every other manifest (content pack) is imported. The new requirement is to group the manifests by language and don't import English for Spanish or viceversa.

vanessa-chang:
It verified pass in Narwhal v7.11.0 in spanish pack

PR #892, opened 2023-10-20
Collections backend: Import extra metadata per language

manuq:
Group the manifest files in EN and ES. Then when extra metadata is imported, use the manifest for the corresponding group.

Helps #891


dbnicholson (approved):
That seems good. We'll obviously need to get the `spanish-extras` pack straightened out.

Issue #890, opened 2023-10-20
Key on EOS: app does not work offline


vanessa-chang:
Follow up from https://github.com/endlessm/eos-image-builder/issues/117

There are some customized images the would need the app content preloaded, so that the user can access it offline.

Content: https://github.com/endlessm/endless-key-content-private/issues/103

AC:
- User should be able to access the content either online/offline
- User should be able to get the corresponding package based on the locale setting

vanessa-chang:
Tested an es preload image: https://images.endlessos.org/builds/12693/

There are findings:
- It would take ~2 mins to run the download process
- If the app is started offline, the download progress bar will sit at 1%. It will not continue after re-connect to an internet.
- The Library only show the preloaded channels (ie. 8 channels for the current Spanish starter pack)

dbnicholson:
I think this is going to be fairly straightforward, at least for the case where we've properly preloaded the expected data. For content nodes, I believe that Kolibri will find that everything is present and the scheduled tasks will become no-ops.

The issue is channel database imports. Since channel databases are versioned, Kolibri always tries to download them to see if there's a new version available. However, the collections packs already contain the desired version of the databases. We can check that we already have that version of the database available and skip creating a task for it.

Skipping content task creation would be harder since we'd have to duplicate all the Kolibri code that determines what content files are needed. It's all pretty hairy as it has to descend the content tree in the channel and apply some heuristics to decide what nodes to include and which files from those nodes are needed. I'm hoping we don't need that.

Now, if the preload is missing data that the pack expects, then I don't know how to handle that nicely. We'd need to decide what to do in the case that initial download tasks are failing because of missing network access. Currently it just loops forever trying to restart the tasks over and over. I'm going to consider that handling out of scope for this task.

dylanmccall:
It's actually news to me that we're scheduling tasks here in the first place. Looking at [53-ek-content-preload](https://github.com/endlessm/eos-image-builder/blob/master/hooks/image/53-ek-content-preload), my understanding is the image builder leaves us with a fully populated database and all those content nodes marked `available`, right? So, why is the explore plugging trying to download stuff at all? I could have sworn it didn't do that before, but maybe I just did a bad job testing things with https://github.com/endlessm/endless-key-content-private/issues/97 :b Sorry for my confusion around this. I had assumed it was working a certain way.

[...]

PR #899, opened 2023-10-27
Skip downloading when all content available

dbnicholson:
This allows completing the downloader offline when all content is available by skipping unnecessary channel imports. Otherwise, Kolibri will always tries to import channels in case the remote version has been updated. We don't need that since the desired channel version is specified in the starter pack.

This is on top of #897, which allowed me to confirm it would work and catch a couple bugs I would have otherwise missed.

Fixes: #890

### DIFF
--- a/kolibri_explore_plugin/collectionviews.py
+++ b/kolibri_explore_plugin/collectionviews.py
@@ -82,11 +82,25 @@ class EndlessKeyContentManifest(ContentManifest):
         self.available = None
         super().__init__()
 
+    def get_latest_channels(self):
+        """Return set of channel id and latest version tuples"""
+        return {
+            (channel_id, max(self.get_channel_versions(channel_id)))
+            for channel_id in self.get_channel_ids()
+        }
+
     def get_extra_channel_ids(self):
         all_channel_ids = _get_channel_ids_for_all_content_manifests(
             self.language
         )
         return all_channel_ids.difference(self.get_channel_ids())
+
+    def get_latest_extra_channels(self):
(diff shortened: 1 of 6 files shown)


dbnicholson:
#897 was merged, so I rebased this on master. Ready for review now.

kolibri_explore_plugin/collectionviews.py
```diff
@@ -82,12 +82,26 @@ def __init__(self):
         self.available = None
         super().__init__()
 
+    def get_latest_channels(self):
+        """Return set of channel id and latest version tuples"""
+        return {
+            (channel_id, max(self.get_channel_versions(channel_id)))
+            for channel_id in self.get_channel_ids()
```
manuq, line 89: :+1:

kolibri_explore_plugin/collectionviews.py
```diff
@@ -134,20 +148,34 @@ def get_channelimport_tasks(self):
...
-            get_remotechannelimport_task(channel_id)
-            for channel_id in self.get_extra_channel_ids()
-        ]
+        tasks = []
+        for channel_id, channel_version in self.get_latest_extra_channels():
+            metadata = get_channel_metadata(channel_id)
+            if metadata and metadata.version >= channel_version:
+                logger.debug(
+                    f"Skipping extra channel import task for {channel_id} "
+                    "since already present"
```
manuq, line 174: :+1:

manuq (approved):
Amazing, with tests included!

PR #904, opened 2023-12-19
Remove discontinued channel apps from customApps.js

dylanmccall:
The source files for several channel apps were removed in 40b02b32, but these were still referenced in customApps.js.

Custom thumbnails and icons for these channels are still in the repository, so they are not changed here.

https://phabricator.endlessm.com/T35147

### DIFF
--- a/kolibri_explore_plugin/assets/src/customApps.js
+++ b/kolibri_explore_plugin/assets/src/customApps.js
@@ -39,7 +39,6 @@ export const CustomChannelApps = {
   '2f95235c3709511fa12d007f31ed6a7b': 'steam',
   '1520f018610256549c98ca0140cceebe': 'virtual-field-trips',
   eb4373b5da054c07879d0c969dc1976a: 'virtual-science-teachers',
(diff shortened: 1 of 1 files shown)


kolibri_explore_plugin/assets/src/customApps.js
```diff
@@ -141,7 +137,6 @@ export const EkIguanaSections = [
   {
     name: 'Literature and Language Arts',
     channels: [
-      '59bb2e5a3d2e5e3b85b87d9ab4daa2f3', // reading
```
manuq, line 144: Removing channels from EkIguanaSections means nothing to current Endless Key. EkIguanaSections was for the hardcoded order of channels in the landing page of the USB option (which still has the channels in the USB stick). A real cleanup would be to remove EkIguanaSections altogether with the corresponding views/components:
- kolibri_explore_plugin/assets/src/components/EkIguanaList.vue
- kolibri_explore_plugin/assets/src/views/DiscoveryPageEkIguana.vue

manuq (approved):

PR #905, opened 2024-01-04
Support "downloading" a content pack without an internet connection

dylanmccall:
https://phabricator.endlessm.com/T35137


manuq (commented):
Looks great. In the future it would be nice to also check if Kolibri Studios is reachable. I remember that we looked at this a while back, and the Device tab in Kolibri does that.

PR #907, opened 2024-01-24
Persist content pack choice

manuq:
https://phabricator.endlessm.com/T35159


kolibri_explore_plugin/collectionviews.py
```diff
@@ -647,7 +650,15 @@ def _create_manifest(grade, name, language):
...
-_read_content_manifests()
+def ensure_initiated(api_function):
+    """Decorator to initiate only once in the first API call."""
+
+    def wrapper(*args, **kwargs):
+        if _collection_download_manager is None:
+            _read_content_manifests()
+        return api_function(*args, **kwargs)
+
+    return wrapper
```
dylanmccall: Is anything stopping us from memoizing the output of `_read_content_manifests()` so we can move all these globals out of here? Django doesn't have a _built in_ facility for caching generic Python objects … but there is [django-cache-memoize](https://pypi.org/project/django-cache-memoize/), and a little pet memoize decoration in [kolibri.core.notiifcations.utils](https://github.com/learningequality/kolibri/blob/develop/kolibri/core/notifications/utils.py).

kolibri_explore_plugin/assets/src/modules/topicsRoot/handlers.js
```diff
@@ -147,9 +148,9 @@ export function decideWelcome(store) {
     return Promise.resolve();
   }
 
-  return Promise.all([getDownloadStatus(), getShouldResume()]).then(
-    ([status, { shouldResume, name, sequence }]) => {
-      if (_isDownloadComplete(status)) {
+  return Promise.all([currentCollectionExists(), getDownloadStatus(), getShouldResume()]).then(
+    ([currentExists, status, { shouldResume, name, sequence }]) => {
+      if (currentExists || _isDownloadComplete(status)) {
```
dylanmccall, line 153: :tada:

kolibri_explore_plugin/collectionviews.py
```diff
@@ -103,26 +105,28 @@ def get_latest_channels(self):
...
-        self, grade, name, language, validate=False
+        self, name, sequence, language_id, validate=False
     ):
-        self.grade = grade
-        self.name = name
-        self.language = language
+
+        self.state, created = CollectionState.objects.get_or_create(
+            name=name, sequence=int(sequence)
+        )
```
dylanmccall, line 126: It looks like (if we move this below `super().read(manifest_filename, validate)`) we could do…

```suggestion
        data = self._manifest_data.get("metadata", {})
        self.state, created = CollectionState.objects.get_or_create(
            name=name, sequence=int(sequence), defaults={
                "title": data.get("title"),
                "subtitle": data.get("subtitle"),
                "description": data.get("description"),
                "required_gigabytes": int(data.get("required_gigabytes", 0)),
            }
        )
```

And then we could drop that whole `if created` block.

dylanmccall (approved):
Thanks for renaming collection grade / name here. Much clearer now. I have some small suggestions, but feel free to merge as-is if they're too much trouble.

kolibri_explore_plugin/collectionviews.py
manuq: Memoizing could be part of a follow-up for persisting the collection manifest. Anyways I would keep that in a separate model. My intention was to decouple the collection manifest from the collection state. I wanted (but ran out of time) to come up with an EndlessKeyCollection class that didn't inherit from ContentManifest anymore. Instead:
```python
class EndlessKeyCollection():
    def __init__(self):
        self.state = None # CollectionState object (Django model)
        self.manifest = None # ContentManifest object
```

PR #909, opened 2024-02-07
collectionviews: Ignore session state if it uses the wrong format

dylanmccall:
It is possible for COLLECTIONS_STATE to be populated with data from a previous version of kolibri-explore-plugin, which results in unhandled errors in new code which expects it to be in a different format.

https://phabricator.endlessm.com/T35159

### DIFF
--- a/kolibri_explore_plugin/collectionviews.py
+++ b/kolibri_explore_plugin/collectionviews.py
@@ -940,8 +940,8 @@ def resume_download(request):
     # Init the download manager and start downloading
     try:
         _collection_download_manager.from_state(saved_state)
-        name = saved_state["name"]
-        sequence = saved_state["sequence"]
+        name = saved_state["collection_name"]
+        sequence = saved_state["collection_sequence"]
         logger.info(f"Download resumed for name={name} sequence={sequence}")
         logger.info(f"Resumed download state: {saved_state}")
     except DownloadError as err:
--- a/kolibri_explore_plugin/collectionviews.py
+++ b/kolibri_explore_plugin/collectionviews.py
@@ -6,6 +6,7 @@ import os
 import time
 from enum import auto
 from enum import IntEnum
+from threading import RLock
 
 from django.utils.translation import gettext_lazy as _
 from kolibri.core.content.errors import InsufficientStorageSpaceError
@@ -71,6 +72,8 @@ PROGRESS_STEPS = {
     "downloading": 0.9,
     "completed": 1,
 }
+
+ENSURE_INITIATED_RLOCK = RLock()
 
 
 class ChannelNotImported(Exception):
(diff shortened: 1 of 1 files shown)


kolibri_explore_plugin/collectionviews.py
```diff
@@ -910,7 +927,7 @@ def resume_download(request):
     Returns download status.
     """
 
-    saved_state = request.session.get("COLLECTIONS_STATE")
+    saved_state = _get_saved_collections_state(request)
```
wjt, line 930: It's a preëxisting bug, but: `_get_saved_collections_state` validates that the state contains at least the following keys:

```
    collections_state_keys = (
        "collection_name",
        "collection_sequence",
        "stage",
    )
```

but the block below, inside the `try`, uses `saved_state["name"]` and `saved_state["sequence"]`. One of these two is surely wrong! I think the block below was missed in 88360791.

kolibri_explore_plugin/collectionviews.py
```diff
@@ -803,6 +803,23 @@ def _get_latest_channels_for_all_collections(language_id):
...
+        "stage",
+    )
+    saved_state = request.session.get("COLLECTIONS_STATE")
+
+    if not saved_state:
+        return None
+
+    if not all(key in saved_state for key in collections_state_keys):
+        return None
+
```
wjt, line 819: In practice, `CollectionDownloadManager.from_state` and `CollectionDownloadManager.to_state` are what define the format of this dictionary, so shouldn't this validation live in that class?

wjt (changes requested):

kolibri_explore_plugin/collectionviews.py
dylanmccall: :+1: Indeed it should! Fixed in the newest commit.

kolibri_explore_plugin/collectionviews.py
dylanmccall: In sorting that out, I noticed a _different_ problem, resolved in d4db24e5 :)

kolibri_explore_plugin/collectionviews.py
```diff
@@ -763,9 +766,10 @@ def ensure_initiated(api_function):
     """Decorator to initiate only once in the first API call."""
 
     def wrapper(*args, **kwargs):
-        if _collection_download_manager is None:
-            _read_content_manifests()
-        return api_function(*args, **kwargs)
+        with ENSURE_INITIATED_RLOCK:
+            if _collection_download_manager is None:
+                _read_content_manifests()
+            return api_function(*args, **kwargs)
```
wjt, line 772: Is it intentional that the wrapped function is called with the lock held?

IIUC, the lock is meant to ensure that `_read_content_manifests()` is called only once; so I would expect that the lock could be released before calling the wrapped `api_function`. But I'm probably wrong!

wjt (commented):
I have a query about the new locking. If I am misunderstanding and the lock does need to be held while calling the wrapped_function, then consider this an approval.

kolibri_explore_plugin/collectionviews.py
dylanmccall: Ooh yes, that isn't intentional. Thanks for noticing :)

